PR #16, opened 2019-06-08
Timers for Recent Blocks + Recent Transactions

robin-thomas:
* Timers for Recent Blocks + Recent Transactions
* Switch between testnet dropdown options

### DIFF
--- a/components/HeaderDropdownNav.js
+++ b/components/HeaderDropdownNav.js
@@ -15,15 +15,13 @@ class HeaderDropdownNav extends Component {
     this.state = {
       network: "Testnet One"
     };
-
-    this.selectNetwork = this.selectNetwork.bind(this);
   }
 
-  selectNetwork(e) {
+  selectNetwork = e => {
     this.setState({
       network: e.target.innerHTML
     });
-  }
+  };
 
   render() {
     return (
--- a/components/HeaderDropdownNav.js
+++ b/components/HeaderDropdownNav.js
@@ -1,0 +1,51 @@
+import { Component } from "react";
+
+import { Dropdown } from "react-bootstrap";
+
+const HeaderDropdownItem = props => (
+  <Dropdown.Item className="header-nav-item-dropdown" onClick={props.onClick}>
+    {props.title}
+  </Dropdown.Item>
+);
+
+class HeaderDropdownNav extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      network: "Testnet One"
+    };
+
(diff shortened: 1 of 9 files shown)


components/dashboard/DashboardTransactionsRow.js
```diff
@@ -49,86 +53,120 @@ const TransactionImage = {
...
+
+  componentDidMount() {
+    this.timer = setInterval(() => this.tick(), 1000);
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.timer);
+  }
+
+  tick() {
```
frol: I think we can avoid code duplication with a separate component, which will handle the ticks.

package.json
```diff
@@ -20,6 +20,7 @@
   "homepage": "https://github.com/nearprotocol/near-explorer#readme",
   "dependencies": {
     "bootstrap": "^4.3.1",
+    "moment": "^2.24.0",
```
frol: FYI, Moment.js [used to be quite fat](https://github.com/moment/moment/issues/3376) the last time I checked it. [Luxon](https://moment.github.io/luxon/docs/manual/why.html) may be a better alternative, but I haven't checked it myself.

I think we can live with Moment.js for now, but we should keep this point in mind once.

components/HeaderDropdownNav.js
```diff
@@ -0,0 +1,51 @@
...
+
+class HeaderDropdownNav extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      network: "Testnet One"
+    };
+
+    this.selectNetwork = this.selectNetwork.bind(this);
```
frol: It seems easier to define the methods with arrow functions and avoid this manual rebinding. So, you just drop this line and define the method as:

```
selectNetwork = (e) => {
  ...
}
```

Just a random article on this topic: https://javascriptweblog.wordpress.com/2015/11/02/of-classes-and-arrow-functions-a-cautionary-tale/

package.json
robin-thomas: I'm pretty used to `Moment.js`. So used it. But yes, will take a look into that later.

components/dashboard/DashboardTransactionsRow.js
robin-thomas: Yes, I did think of that initially. Anyway, have created a new Timer component to handle the ticks.

components/HeaderDropdownNav.js
robin-thomas: Thanks. Switched to that.

PR #17, opened 2019-06-09
Use react context to share state between components/pages

robin-thomas:


components/dashboard/DashboardBlocks.js
```diff
@@ -43,69 +45,30 @@ const DashboardBlocks = () => (
...
-            </Link>
-          </Col>
-        </Row>
+        <DataConsumer>
+          {context => (
+            <Row className="gutter-4">
+              {context.blocks.map((block, index) => {
+                return (
+                  <DashboardBlocksBlock
+                    key={context.blocks[index].blockNumber}
```
frol: Can we use `block` instead of `context.block[index]`?

components/HeaderDropdownNav.js
```diff
@@ -2,48 +2,37 @@ import { Component } from "react";
...
-            onClick={this.selectNetwork}
-          />
-          <HeaderDropdownItem
-            title="Testnet Two"
-            onClick={this.selectNetwork}
-          />
+          {context.networks.map((network, index) => {
+            return (
+              <HeaderDropdownItem
+                key={index}
```
frol: I think the `network` string is a better key than `index`.

components/dashboard/DashboardBlocks.js
robin-thomas: Yes. We can. My mistake. Will fix it now.

components/HeaderDropdownNav.js
robin-thomas: Updated.

components/dashboard/DashboardBlocks.js
robin-thomas: Updated

components/HeaderDropdownNav.js
frol: @robin-thomas I see that you have changed the implementation of `updateNetwork`, but the `key` is still `index`.

PR #19, opened 2019-06-10
Backend initial version

frol:
I will add the README with instructions on how to run the project shortly.

There are three components now:

1. [WAMP (Web Application Messaging Protocol)](https://wamp-proto.org/) router (crossbar), which is responsible for request-response and publish-subscribe type of messaging between the frontend and the backend
2. Next.js frontend
3. Node.js backend (manages DB and handles WAMP requests, and will also push messages through websocket [WAMP])


backend/src/index.js
```diff
@@ -0,0 +1,58 @@
...
+      return process.env.WAMP_NEAR_EXPLORER_BACKEND_SECRET || "back";
+    }
+    throw "WAMP authentication error: unsupported challenge method";
+  }
+});
+
+async function main() {
+  wamp.onopen = async session => {
+    console.log("WAMP connection is established. Waiting for commands...");
+    await session.register(
```
frol: This is where the backend handles the requests (there is an example of a message broadcast below)

components/utils/DataProvider.js
```diff
@@ -13,34 +15,87 @@ class DataProvider extends Component {
...
+  }
+
+  componentDidMount() {
     this.updateNetwork(0);
   }
 
+  static async getBlocksInfo() {
+    let blocks;
+    try {
+      blocks = await call(".select", [
```
frol: Reaching the backend from the frontend is as simple as `call('remote-function-name', ['params'])`.

frol:
I have bumped into the following bug: https://github.com/zeit/next.js/issues/7551

It only gets reproduced in the release build (`npm run build && npm run start`).

frol:
I worked around the issue by exporting the whole context object `DataContext` from `components/utils/DataProvider.js` instead of only `DataContext.Consumer`. I don't really understand what is going on, but it fixed the issue (`"Cannot read property 'currentNetwork' of undefined"`)

frol:
I have added the installation instructions, but since the bug with React Context is still not fixed, the "release" configuration will show HTTP 500 error:

```
TypeError: Cannot read property 'currentNetwork' of undefined
    at Object.children (/mnt/storage/projects/near.ai/near-explorer/.next/server/static/f9IcGnnQ_5eXJ9nqgeda_/pages/index.js:2267:16)
    at a.render (/mnt/storage/projects/near.ai/near-explorer/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:44:64)
    at a.read (/mnt/storage/projects/near.ai/near-explorer/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:41:58)
    at renderToString (/mnt/storage/projects/near.ai/near-explorer/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:53:83)
    at render (/mnt/storage/projects/near.ai/near-explorer/node_modules/next-server/dist/server/render.js:86:16)
    at renderPage (/mnt/storage/projects/near.ai/near-explorer/node_modules/next-server/dist/server/render.js:211:20)
    at Function.value (/mnt/storage/projects/near.ai/near-explorer/.next/server/static/f9IcGnnQ_5eXJ9nqgeda_/pages/_document.js:812:41)
    at Object.loadGetInitialProps (/mnt/storage/projects/near.ai/near-explorer/node_modules/next-server/dist/lib/utils.js:42:35)
    at Object.renderToHTML (/mnt/storage/projects/near.ai/near-explorer/node_modules/next-server/dist/server/render.js:218:36)
    at processTicksAndRejections (internal/process/task_queues.js:86:5)
```

I will investigate the issue tomorrow.

components/dashboard/DashboardBlocksBlock.js
```diff
@@ -28,13 +29,13 @@ const DashboardBlocksBlock = props => (
               />
               <span className="d-none d-sm-block">@{props.witness}</span>
               <span className="d-sm-none">
-                @{props.witness ? props.witness.substring(0, 7) : null}..
+                {props.witness ? `@${props.witness}` : null}..
```
robin-thomas: @frol I used `props.witness.substring(0, 7)` because that section is rendered only on mobiles, where anything greater than that will break the screen.

components/utils/DataProvider.js
```diff
@@ -13,34 +15,87 @@ class DataProvider extends Component {
...
+      DataProvider.getDetails(),
+      DataProvider.getBlocksInfo()
+    ]);
+    return {
+      details,
+      blocks
+    };
+  }
+
+  componentDidMount() {
```
robin-thomas: @frol Why make the call in `componentDidMount` instead of `componentWillMount`?
Does the backend call take that much time?

components/utils/DataProvider.js
```diff
@@ -13,34 +15,87 @@ class DataProvider extends Component {
...
     this.updateNetwork(0);
   }
 
+  static async getBlocksInfo() {
+    let blocks;
+    try {
+      blocks = await call(".select", [
+        "SELECT hash, height, timestamp, list_of_approvals as listOfApprovals FROM blocks LIMIT 10"
```
robin-thomas: @frol What order are the blocks returned in? Are the blocks created later on top?

components/utils/DataProvider.js
frol: `componentWillMount` is considered legacy and unsafe: https://reactjs.org/docs/react-component.html#the-component-lifecycle

components/dashboard/DashboardBlocksBlock.js
frol: Thank you for pointing it out; I will bring the substring call back.

components/utils/DataProvider.js
frol: Good point. The DB structure is still lacking, so this was barely an example of how to fetch data from the backends.

robin-thomas: Ah, that's fine. Maybe you can add an `ORDER_BY height DESC` to the query, which should fix it right?

components/utils/DataProvider.js
robin-thomas: The reason I used `componentWillMount` is that I wanted the initial values populated before the render happens (rather than seeing the render happening, and then immediately rendering it again with the updated values).

I would have preferred doing the initialization in `constructor` if I could have.

But now I see you have some async calls happening. Then `componentDidMount` makes more sense.

frol: When you need something server-rendered, you should use `static async getInitialProps` (next.js built-in "extension"; just keep in mind that it works out of the box only on the page components and you have to manually call it on child components if you need it down the line). `componentWillMount` is not blocking.

frol:
I have found the relevant bug report for the React Context issue: https://github.com/zeit/next.js/issues/7167, and it has already been fixed in the master (and canary releases): https://github.com/zeit/next.js/pull/7200, so I just updated `package.json` to require the latest 8.1.1 canary build.

I believe we can merge this PR as is and move on with incremental changes in next PRs.

robin-thomas:
@frol I'll merge it for now. Like you said, let's do move on with incremental changes in next PRs.

PR #30, opened 2019-07-16
Refactored (simplified) read-only database setup for com.nearprotocol.explorer.select

frol:
/cc @robin-thomas

### DIFF
--- a/backend/config/config.json
+++ b/backend/config/config.json
@@ -1,44 +1,14 @@
 {
   "development": {
     "dialect": "sqlite",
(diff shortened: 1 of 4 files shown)


robinth91:
@frol, Will it be possible to connect to the real-time db even during development?
It's loaded in read-only mode anyway.

frol:
@robinth91 yes, you already can connect to wss://near-explorer-wamp.nearprotocol.com/ws (pass via environment variable `WAMP_NEAR_EXPLORER_URL`)

PR #41, opened 2019-07-24
Improve dashboard txn display with templating

robin-thomas:


vgrichina:
Your PR Server URL is https://near-explorer-frontend-pr-41.onrender.com.

Follow its progress at https://dashboard.render.com/web/srv-bks3vgefgkcssdfspde0.

Cheers,

frontend/components/dashboard/DashboardTransactionsRow.js
```diff
@@ -1,8 +1,54 @@
+import mustache from "mustache";
 import { Row, Col } from "react-bootstrap";
 
 import Timer from "../utils/Timer";
 import GetTransactionIcon from "../utils/GetTransactionIcon";
 
+const prettifyTransaction = (kind, args) => {
+  if (kind === "AddKey") {
+    try {
+      args = JSON.parse(args);
```
frol: I think, we should resolve args once in the DataProvider instead of doing it every single time.

In fact, I consider moving this parsing to the backend side, but that will require adding an additional option to the `.select` RPC call as sequelize does not know which columns of the raw SELECT are of JSON type.

frontend/components/dashboard/DashboardTransactionsRow.js
```diff
@@ -1,8 +1,54 @@
...
+    } catch (err) {
+      console.log(err);
+      return args;
+    }
+
+    return mustache.render(
```
frol: I have a concern regarding the need using mustache: https://github.com/nearprotocol/near-explorer/issues/32#issuecomment-514615945

frontend/components/dashboard/DashboardTransactionsRow.js
robin-thomas: Yes. It makes sense to do it in DataProvider once. I'll shift it over there.

frontend/components/dashboard/DashboardTransactionsRow.js
robin-thomas: So you recommend to do it React way, by making separate components for each transaction type?

robin-thomas: If you look at the latest pr, you can see I had split it into a component per transaction type. But I had also used mustache as templating the string message inside `DataProvider`.

frontend/components/dashboard/DashboardTransactionsRow.js
```diff
@@ -1,92 +1,19 @@
...
-    `}</style>
-  </Row>
-);
+import AddKey from "./transaction/AddKey";
+import CreateAccount from "./transaction/CreateAccount";
+import FunctionCall from "./transaction/FunctionCall";
+import TransactionRow from "./transaction/transaction";
+
+const DashboardTransactionRow = props => {
+  switch (props.txKind) {
```
frol: Have you considered using a map?

```javascript
const components = {
    AddKey,
    CreateAccount,
};

...
  const Component = components[props.txKind];
  return <Component {...props} />;
```

frontend/components/dashboard/DashboardTransactionsRow.js
frol: I still don't understand the need for `mustache` over [in this particular case] JS templating strings.

I want to make a note, that I am not requiring the changes and avoiding `mustache`, I want to discuss and see if we really need it, like how this new dependency helps us.

frontend/components/dashboard/DashboardTransactionsRow.js
robin-thomas: @frol I had. But I wasn't sure how to handle the `default` condition (it's just for now since I haven't seen all type of transactions in the real data) in the map.

frol: @robin-thomas To handle the missing type you can just catch the `undefined` `Component`:

```js
  const Component = components[props.txKind] || UnknownTransactionRow;
  return <Component {...props} />;
```

robin-thomas: Alright. Will update the PR.

frontend/components/dashboard/DashboardTransactionsRow.js
robin-thomas: @frol So how do you recommend doing the same templating of strings (if we don't use `mustache`)?

frontend/components/dashboard/DashboardTransactionsRow.js
```diff
@@ -1,92 +1,17 @@
...
-    `}</style>
-  </Row>
-);
+import AddKey from "./transaction/AddKey";
+import CreateAccount from "./transaction/CreateAccount";
+import FunctionCall from "./transaction/FunctionCall";
+import TransactionRow from "./transaction/transaction";
+
+const DashboardTransactionRow = props => {
+  const components = {
```
frol: I think we can safely move this const to the top level, so it does not get recreated on every render.

PR #45, opened 2019-07-26
Migrated timestamps type into BIGINT (milliseconds) instead of strings

frol:
@robin-thomas FYI, I have changed the internal storage for timestamps (`blocks.timestamp` and `nodes.last_seen`) to save data as [UNIX epoch] milliseconds, and it is now a number type instead of a string.


vgrichina:
Your PR Server URL is https://near-explorer-frontend-pr-45.onrender.com.

Follow its progress at https://dashboard.render.com/web/srv-bktkcgefgkcntf8reaog.

Cheers,

PR #49, opened 2019-07-27
Block search

robin-thomas:


vgrichina:
Your PR Server URL is https://near-explorer-frontend-pr-49.onrender.com.

Follow its progress at https://dashboard.render.com/web/srv-bku5qkr60hnbjo0q1380.

Cheers,

frontend/components/utils/Pagination.js
```diff
@@ -9,6 +9,9 @@ const Pagination = props => {
   useEffect(() => {
     const ele = document.getElementById(props.elementId);
     const isAtBottom = () => {
+      if (ele === null || ele === undefined) {
```
frol: I believe we should just skip entirely the event setting if the element has not been found, right?

frontend/components/utils/Search.js
```diff
@@ -1,60 +1,89 @@
...
+
+  return (
+    <div className="search">
+      <Row noGutters="true">
+        <Col md="1" xs="2" className="align-self-center">
+          <img src="/static/images/icon-search.svg" className="search-icon" />
+        </Col>
+        <Col md="11" xs="10" className="align-self-center">
+          <div className="d-none d-sm-block">
+            <form onSubmit={onSubmit}>
```
frol: 1. It is not obvious that you should hit enter (submit the form) to start the search.
2. It is not clear if anything is happening at all once you submit (there is no indication of the process)

robin-thomas: @frol 
1. Have modified the search box to look like the one used by Google. Users can click on the search icon at the end or press Enter. Exactly what Google follows.

2. Have added a loading overlay for search. Please check the latest commit.

robin-thomas:
@frol We are using the infinity scrolling (just like followed by Facebook and Twitter). We cannot have pagination when we have real-time data. So if the user refreshes the page, they would be provided with latest blocks.

frontend/components/utils/Pagination.js
robin-thomas: @froi
> Also, we also should stop trying to load more blocks when backend returns an empty list (try searching for an exact block and see that it keeps trying to load more blocks).

Hmm, but we are dealing with real-time data. So what if the first couple of times, the block wasn't found, but was found later. So we cannot return an empty list then.


> Another one, it would be a better UX if we display "no blocks" if there are no blocks found.

Yes. We should do that. Can I make it in another pr?

frol: Ok, that is fine for now. Let’s only address the first comment, and it is good to be merged.

robin-thomas: Which comment are you talking about?

frol: > I believe we should just skip entirely the event setting if the element has not been found, right?

robin-thomas: @frol Should be fixed now. Can you check again?

PR #73, opened 2019-09-06
Added nodes listing

frol:


frontend/components/Blocks.js
```diff
@@ -104,16 +104,6 @@ const BlocksOverlay = ({ loading, setLoading }) => {
 const Blocks = () => {
   const [loading, setLoading] = useState(false);
 
-  const constructBlock = (block, index, length) => {
```
frol: This turned out to be dead code.

Issue #84, opened 2019-09-28
Port to TypeScript


deacix:
The whole code should be ported to TypeScript to improve on boarding new users and have better code handling.

deacix:
After some hours of analyzing is my suggestion is to move to Angular. You will get great toolset and a stable environment for your app instead of putting different libs and configs together..

frol:
What kind of configs do you refer to? Next.js provides TS environment out of the box (just name a file with ts or tsx extension). I just couldn’t find a use for TS in the React (frontend) app. On the backend side there is no framework at all, it is a plain Node.js application, so I just didn’t want spend my time setting up TS there.

deacix:
can you ping me in telegram: @deacix

PR #104, opened 2019-10-10
Introduced unit testing for frontend

### LINKED ISSUE #87: Introduce unit testing for frontend
## Story

As a developer who is working on near-explorer, I want to test my written code so I can verify that my implementations is working and valid.

## Acceptance Criteria

- [x] Configuration for unit testing is setup for frontend
- [x] Unit testing is working locally
- [x] Unit testing is working in the CI

frol:
*Resolves #87*


frol:
I have introduced some tests for `Action*` components and ported those components to TypeScript (#84). The tests are simple snapshot-based tests (`jest` + `react-test-renderer` is a great combo to do regression testing with diffs, if the snapshot does not match, it will fail the test, but it is just a matter of `npm run test --updateSnapshot` to "mark" the new snapshot as a baseline)

frontend/src/components/transactions/types.ts
```diff
@@ -0,0 +1,43 @@
+export interface Transaction {
```
frol: These types are available in `nearlib`, but we don't use `nearlib` on the frontend, so I would extract them into a separate package in the future. Ideally, I would prefer having them defined by nearcore public API (JSON RPC is not great at this, so ProtoBuf / Flatbuffers / ... with automatic code-generation is the direction I want to move in future)

PR #106, opened 2019-10-11
Link Near logo in header to Dashboard/Home Page

kcole16:


frontend/src/components/utils/Header.jsx
```diff
@@ -7,14 +8,18 @@ export default () => (
   <Container fluid="true" className="near-main-container">
     <Row className="header-container">
       <Col xs="2" className="px-0 d-md-none align-self-center">
-        <img
-          className="near-main-logo-mobile"
-          src="/static/images/near_icon_wht.svg"
-        />
+        <Link href="/">
```
frol: We should wrap the `img` tag in empty `a`, so the native link is there (this helps in navigation and also in cases when the page is server-rendered):

```html
<Link href...><a><img .../></a></Link>
```

In this case you also don’t need the CSS rule.

kcole16: Good call. Added

PR #119, opened 2019-10-31
Implemented multinet support

### LINKED ISSUE #110: Switch Network (Stake Wars)
## Story

As a user testing my app on multiple networks, I should be able to easily switch between them (e.g. testnet and stake wars net), so I can view my activity on each network.

## Acceptance Criteria

- [x] User can view their current network as the default/selected item in a dropdown in the navbar
- [x] User can switch networks using the dropdown, and view the newly selected network's activity after a refresh
  - [x] Testnet
  - [x] Stake Wars
- [x] Mobile ok
- [ ] Tested

frol:
*Resolves #110*

I had to refactor the API helpers and introduce more configuration options via environment variables on both the frontend and the backend sides. The routing is done by having different domain names (explorer.nearprotocol.com, explorer.staging.nearprotocol.com, ...) matched to the relevant near networks (specifically, RPC endpoints). Ultimately, a single frontend server instance and a single WAMP router instance handle all networks, but I have chosen to run backend servers independently for each network (yet, they are deployed into a single Docker container on Render.com, since there are enough resources to handle all those, and it is easier to maintain).

@deacix This PR introduces breaking changes in `frontend/src/libraries/explorer-wamp`. You should be good to go with a simple change like this:

```js
import * as BlocksApi from ...;

const blocks = await BlocksApi.searchBlocks(...);
```

into

```js
import BlocksApi from ...;

const blocks = await new BlocksApi().searchBlocks(...);
```

I wrapped the standalone methods into classes to handle multinet on the server side.

Let me know if you need any assistance.

/cc @kcole16


frol:
Unit tests and e2e tests pass locally. Unit tests pass on CI, but e2e fail due to the fact that the multinet support on the backend is not deployed (included in this PR). We should come up with a strategy to run e2e tests against the local backend.

Issue #99, opened 2019-10-04
Exact Search for TX, Account, Block


kcole16:
## Story

As a user, I want to be able to search for a transaction by it's hash, an account by name, and a block by block number, so I can view the details of the TX/account/block to debug without interpreting the URL schema or finding a link.

## Acceptance Criteria

- [x] User can search by TX hash
- [x] User can search by account name
- [x] User can search by block hash
- [x] User can search by block height
- [x] User is redirected if a match is found
- [x] User receives a message "Result not found" if no match
- [ ] Tested

## Assets

- [Designs](https://app.zeplin.io/project/5c6f069601b1361ba39eaee7/screen/5c6f0a977372fb3306dd0211)

kcole16:
@deacix Added the designs for this (in Zeplin)

kcole16:
@icerove Adding you to this issue as well. Just needs an enhancement:

Currently, you can search by block hash, whereas we expect the user to search by block number (since they are more likely to have that)

bowenwang1996:
@kcole16 there is not a one-to-one mapping between block height and block, especially for recent blocks.

kcole16:
@bowenwang1996 Interesting, is there ever a guaranteed one-to-one mapping, or should we just keep search to block hash?

bowenwang1996:
For finalized blocks yes we can say that the mapping from blocks to heights is injective, but still not surjective though. In other words, for some height, there might not exist a corresponding block.

icerove:
So do we still want to add the function to search from block height? @kcole16

kcole16:
@icerove Yes, let's add block height in addition to block hash, so users can search by either

PR #130, opened 2019-11-25
Feature/search box 99

kcole16:
References #99


frontend/src/components/dashboard/DashboardHeader.jsx
```diff
@@ -1,73 +1,174 @@
...
-  transactionsPerSecond,
-  lastDayTxCount,
-  accountsCount
-}) => (
-  <div className="dashboard-info-container">
-    <Row noGutters>
-      <Col xs="12" md="3">
-        <Link href="/nodes">
-          <a>
+    alert("Nothing found!");
```
frol: This is not pretty, but I prefer working on tasks incrementally instead of blocking them forever in PRs, so I would create a separate issue to address the error reporting.

PR #140, opened 2019-12-06
Search by height

### LINKED ISSUE #99: Exact Search for TX, Account, Block

icerove:
fix #99

### DIFF
--- a/frontend/src/components/dashboard/DashboardHeader.jsx
+++ b/frontend/src/components/dashboard/DashboardHeader.jsx
@@ -23,9 +23,7 @@ export default class DashboardHeader extends React.Component {
     const blockPromise = new BlocksApi()
       .getBlockInfo(searchValue)
       .catch(() => {});
-    const blockHeightPromise = new BlocksApi()
-      .searchBlocks(searchValue)
-      .catch(() => {});
+
     const transactionPromise = new TransactionsApi()
       .getTransactionInfo(searchValue)
       .catch(() => {});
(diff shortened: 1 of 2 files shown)


icerove:
should we remove the search on block page? Or should that search also redirect to the block page?

frol:
@icerove that search is filtering the blocks by prefix. I am not sure how useful that is, but it works.

frontend/src/components/dashboard/DashboardHeader.jsx
```diff
@@ -23,6 +23,9 @@ export default class DashboardHeader extends React.Component {
     const blockPromise = new BlocksApi()
       .getBlockInfo(searchValue)
       .catch(() => {});
+    const blockHeightPromise = new BlocksApi()
+      .searchBlocks(searchValue)
+      .catch(() => {});
```
frol: It will search by a height *prefix* (i.e. not an exact search). I suggest we just use `getBlockInfo(blockId)`, where `blockId` is going to be either block hash or block height and then change the query from `WHERE blocks.hash = :hash` to `WHERE blocks.hash = :blockId OR blocks.height = :blockId`. Once we change the implementation of `getBlockInfo`, we don't need this separate call, as `blockPromise` will already search by hash or height.

icerove: got it

Issue #134, opened 2019-12-03
Update NEAR token units to support yoctoNEAR


kcole16:
nearprotocol/nearcore#1723

@frol What do we need to update in Explorer to support this change?

frol:
I believe, it is just a single place: https://github.com/nearprotocol/near-explorer/blob/82b69a20ae7960b26b01b63f7a02acccc7d59599/frontend/src/components/utils/Balance.jsx (in fact, this copy-paste from Wallet should be moved to nearlib as discussed, and in that case we only need to drop this implementations in favor of the nearlib one).

icerove:
I agree with Frol

icerove:
I think for now what we can do is update explorer since writing in nearlib is not happened for now

PR #146, opened 2019-12-12
Change nomination

icerove:
fix the issue #134 but we should still wait ubtil nearcore and nearlib is okay to update

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -35,5 +35,15 @@ module.exports = withCSS({
     wampNearExplorerUrl:
       process.env.WAMP_NEAR_EXPLORER_URL || defaultWampNearExplorerUrl,
     googleAnalytics: process.env.NEAR_EXPLORER_GOOGLE_ANALYTICS
+  },
+  webpack: (config, { isServer }) => {
+    // Fixes npm packages that depend on `fs` module
+    if (!isServer) {
+      config.node = {
+        fs: "empty"
+      };
+    }
+
+    return config;
   }
 });
--- a/frontend/src/components/transactions/__tests__/__snapshots__/ActionMessage.test.tsx.snap
+++ b/frontend/src/components/transactions/__tests__/__snapshots__/ActionMessage.test.tsx.snap
@@ -61,11 +61,11 @@ Array [
 ]
 `;
 
-exports[`<ActionMessage /> renders Stake 1`] = `"Staked: <0.00001 Ⓝ ed25519:BgXFiJS..."`;
+exports[`<ActionMessage /> renders Stake 1`] = `"Staked: 0 Ⓝ ed25519:BgXFiJS..."`;
 
 exports[`<ActionMessage /> renders Transfer 1`] = `
 Array [
-  "Transferred <0.00001 Ⓝ to ",
+  "Transferred 0 Ⓝ to ",
   <a
     className="account-link"
     href="/accounts/receiver.test"
--- a/frontend/src/components/transactions/__tests__/__snapshots__/ActionMessage.test.tsx.snap
+++ b/frontend/src/components/transactions/__tests__/__snapshots__/ActionMessage.test.tsx.snap
@@ -61,11 +61,11 @@ Array [
 ]
 `;
 
-exports[`<ActionMessage /> renders Stake 1`] = `"Staked: 0 Ⓝ ed25519:BgXFiJS..."`;
+exports[`<ActionMessage /> renders Stake 1`] = `"Staked: <0.00001 Ⓝ ed25519:BgXFiJS..."`;
 
 exports[`<ActionMessage /> renders Transfer 1`] = `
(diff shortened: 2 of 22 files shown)


icerove:
this one fix the bug for fs in next and nearlib

icerove:
image
after change, it will like this. I think it is still nearlib gives wrong method to shortcut the number.

frontend/src/components/transactions/__tests__/__snapshots__/ActionMessage.test.tsx.snap
```diff
@@ -61,11 +61,11 @@ Array [
 ]
 `;
 
-exports[`<ActionMessage /> renders Stake 1`] = `"Staked: <0.00001 Ⓝ ed25519:BgXFiJS..."`;
+exports[`<ActionMessage /> renders Stake 1`] = `"Staked: 0 Ⓝ ed25519:BgXFiJS..."`;
 
 exports[`<ActionMessage /> renders Transfer 1`] = `
 Array [
-  "Transferred <0.00001 Ⓝ to ",
+  "Transferred 0 Ⓝ to ",
```
frol: We should fix `formatNEAR` to actually reflect that the amount is not 0, but rather very small.

@janedegtiareva Can nearlib format small yoctoNEAR amounts in a human-friendly format, e.g. `1 yoctoN`? We used to have `<0.00001 N` here. Do you think it is reasonable?

@icerove @kcole16 It seems that to make the information non-confusing and useful, we should present both `<0.00001 N (1 yoctoNEAR)` (we may want to show the exact amount, e.g. "1 yoctoNEAR", as an on-hover popup)

frontend/src/components/utils/Balance.jsx
```diff
@@ -19,20 +16,11 @@ const convertToShow = amount => {
 };
 
 export const formatNEAR = amount => {
-  if (amount === "0") {
-    return "0";
-  } else if (amount.length < NOMINATION - 5) {
+  let ret = utils.format.formatNearAmount(amount, 5);
+  if (ret.startsWith("0.00000")) {
```
frol: This does not seem to work as it is evident from the unit tests.

frontend/next.config.js
```diff
@@ -35,5 +35,15 @@ module.exports = withCSS({
...
       process.env.WAMP_NEAR_EXPLORER_URL || defaultWampNearExplorerUrl,
     googleAnalytics: process.env.NEAR_EXPLORER_GOOGLE_ANALYTICS
+  },
+  webpack: (config, { isServer }) => {
+    // Fixes npm packages that depend on `fs` module
+    if (!isServer) {
+      config.node = {
+        fs: "empty"
+      };
+    }
```
frol: Is this something wrong with nearlib? Do we have an issue there? Can we fix it there? /cc @janedegtiareva

icerove: zeit/next.js#7755 the issue and method is here.

icerove: Vlad and I talked about this and this definitely can fix in the nearlib, nearlib does not fit browser configuration and webpack maybe a good method to do with that

icerove:
find out the test is okay just because testnet is down yesterday. If nearlib is not changing anymore, this version for explorer is ready to go.
@Frol I know maybe this is not okey for you, but the changes should happen on nearlib side

frontend/next.config.js
frol: Let's create a tracking issue on nearlib and put a link to the issue in a comment here.

frontend/src/components/transactions/__tests__/__snapshots__/ActionMessage.test.tsx.snap
frol: @icerove I still don't see this resolved. "Transferred 0 N" is not useful.

kcole16:
> I would like not to regress on the "<0.00001 N" display. Other than everything is fine.

I created a separate issue to encapsulate this #152 (and put it as highest priority), as we're proceeding with yoctoNEAR rollout currently.

frontend/src/components/transactions/__tests__/__snapshots__/ActionMessage.test.tsx.snap
frol: Well, this same problem is going to be in Wallet.

@vgrichina @janedegtiareva How do you display small NEAR amounts in Wallet? It makes sense to have 1 yoctoNEAR rendered as something more human-friendly than 0 N.

vgrichina: @frol wallet still displays `<0.00001 N`. 

https://github.com/nearprotocol/near-wallet/blob/41cb65246134308dd553b532dfb314b45b38b65c/src/components/common/Balance.js#L26

Issue #132, opened 2019-11-28
Combine actions in a single TX into one line on Transactions page and Dashboard


kcole16:
Currently, each action is shown as an individual line in the TX list on the Dashboard and Transactions page, even when multiple actions are part of a single transaction.

**Example**
![](https://yc3.s3.us-east-2.amazonaws.com/Screen+Shot+2019-11-27+at+4.25.45+PM.png)

These three actions are all in the same TX (note the TX hash). The line item should be "Batch Transaction", with an open drawer component containing the three actions.

**[Designs](https://projects.invisionapp.com/share/TCS6KG8G9V7#/screens)**
![](https://yc3.s3.us-east-2.amazonaws.com/Screen+Shot+2019-12-09+at+12.55.59+PM.png)

frol:
It is unclear whether the first action is a good representation of the whole transaction, which is why we went with the flat structure

kcole16:
Hmm, that's a good point. We should find a way to indicate the grouping to the user though. Will see if @jakestutzman has any ideas

icerove:
how about color the transactions in same light color for the same hash to mark these lines are from the same single transaction? Light color for the line not to confuse the content and easy to figure out.

kcole16:
@icerove @frol Updated the issue description with the new designs.

kcole16:
Closed by #147

Issue #143, opened 2019-12-09
Failed TX showing as "Completed"


kcole16:
[This TX](https://explorer.nearprotocol.com/transactions/Hs3cNu9UtuuN5mbnmpDZCcrEZfjsNHyMXdZ1YyaNsaT9) is incorrectly showing as "Completed" in Explorer:

RPC status check response
```
Transaction kendall123:Hs3cNu9UtuuN5mbnmpDZCcrEZfjsNHyMXdZ1YyaNsaT9
{
  receipts: [
    {
      id: 'BAzV7AysJWdjk5zsqRssDxzLXUEsjVL9ggt7dKNih5j9',
      outcome: {
        gas_burnt: 2,
        logs: [],
        receipt_ids: [ '6VPdP4fjsinJJ4P1VyLy1jQ8QHBv6h7oEiytd2KCFhqb' ],
        status: {
          Failure: {
            error_message: `Can't complete the action "Transfer", because account "fakeaccount12312412" doesn't exist`,
            error_type: 'ActionError::AccountDoesNotExist'
          }
        }
      }
    },
    {
      id: '6VPdP4fjsinJJ4P1VyLy1jQ8QHBv6h7oEiytd2KCFhqb',
      outcome: {
        gas_burnt: 0,
        logs: [],
        receipt_ids: [],
        status: { SuccessValue: '' }
      }
    }
  ],
  status: {
    Failure: {
      error_message: `Can't complete the action "Transfer", because account "fakeaccount12312412" doesn't exist`,
      error_type: 'ActionError::AccountDoesNotExist'
    }
  },
  transaction: {
    id: 'Hs3cNu9UtuuN5mbnmpDZCcrEZfjsNHyMXdZ1YyaNsaT9',
    outcome: {
      gas_burnt: 2,
      logs: [],
      receipt_ids: [ 'BAzV7AysJWdjk5zsqRssDxzLXUEsjVL9ggt7dKNih5j9' ],
      status: {
        SuccessReceiptId: 'BAzV7AysJWdjk5zsqRssDxzLXUEsjVL9ggt7dKNih5j9'
      }
    }
  }
}
```

PR #147, opened 2019-12-13
Combine transaction

icerove:
fix issue #132 
comment: it is a little different from the designs, since every action has been written into a component that shows the whole information, if changed to what design shows, there will be many changes and create several new component to reuse and inspect for different situation. Have a look and give feedback about this


kcole16:
> I like the designed version better as it does not have duplicated information which may confuse people.

@icerove I agree with @frol here. We want to communicate to the user that each action was one transaction,  by one user,  confirmed with one timestamp.

icerove:
okay, I will do that.

kcole16:
@icerove Just noticed some missing formatting on the [TX detail](https://near-explorer-frontend-pr-147.onrender.com/transactions/4CLfJE6PV1kuRpDPDgmk98Up5bVuuWs1isVx2vxskxWZ) and [block detail](https://near-explorer-frontend-pr-147.onrender.com/blocks/AJtxLT6cRA5WUnys78qDGKsM7PHUDEUpA8h1g634RQ3S) pages.



On these pages, we can remove the "Batch Transaction" text, timestamp, and sender (since it's listed above).



Forgot to spec this before!

icerove:
> @icerove Just noticed some missing formatting on the [TX detail](https://near-explorer-frontend-pr-147.onrender.com/transactions/4CLfJE6PV1kuRpDPDgmk98Up5bVuuWs1isVx2vxskxWZ) and [block detail](https://near-explorer-frontend-pr-147.onrender.com/blocks/AJtxLT6cRA5WUnys78qDGKsM7PHUDEUpA8h1g634RQ3S) pages.
Make sence to TX detail page but block detail page still need batch I think

kcole16:
> Make sence to TX detail page but block detail page still need batch I think

Good point!

icerove:
single transaction is not correct on detail page

frontend/src/components/transactions/ActionsList.tsx
```diff
@@ -1,31 +1,220 @@
...
+                </Col>
+                <Col md="4" xs="5" className="ml-auto text-right">
+                  <Row>
+                    <Col className="action-row-txid">
+                      <TransactionLink transactionHash={transaction.hash} />
+                    </Col>
+                  </Row>
+                  <Row>
+                    <Col className="action-row-timer">
+                      <span className="action-row-timer-status">{`Completed`}</span>{" "}
```
frol: Since #143 there is a proper `transaction.status` value, which can be formatted as `<ExecutionStatus status={transaction.status} />` (`import ExecutionStatus from "../utils/ExecutionStatus";`)

frontend/src/components/transactions/ActionsList.tsx
```diff
@@ -1,31 +1,220 @@
...
+import AccountLink from "../utils/AccountLink";
+import TransactionLink from "../utils/TransactionLink";
+import Timer from "../utils/Timer";
 
 export interface Props {
   actions: (T.Action | keyof T.Action)[];
   transaction: T.Transaction;
   viewMode?: ViewMode;
   reversed?: boolean;
+  TD?: boolean;
```
frol: What is this?

frontend/src/components/transactions/ActionsList.tsx
icerove: got it and change

frontend/src/components/transactions/ActionsList.tsx
icerove: on TransactionDetail page, kendall think no more need to show detail on each transaction row, because it is already show on the title, so I add this to check whether on the detail page to just show transaction information

kcole16:
@icerove Let's center the action descriptions here:

icerove:
@kcole16 You mean in detail page, every action is in the center? I am not very sure what do you want?

kcole16:
> @kcole16 You mean in detail page, every action is in the center? I am not very sure what do you want?

Ah right, I meant align the action description (e.g. "New key added") with the icon on the left, like this:

frontend/src/components/transactions/ActionsList.tsx
```diff
@@ -1,31 +1,222 @@
...
+                      </span>{" "}
+                      <Timer time={transaction.blockTimestamp} />
+                    </Col>
+                  </Row>
+                </Col>
+              </Row>
+            </Col>
+          </Row>
+        ) : null}
+        {actionRows}
```
frol: I suggest nesting the batched rows in terms of HTML structure. This way you will have better control over indentation (you just set margin for the block and no need to apply `margin-left` to individual lines) and other styling.

frontend/src/components/transactions/ActionsList.tsx
frol: 1. TD is completely broken naming. Please, never use such names. I believe, we can completely avoid this special case here.
2. Merge `ActionRowBatch` into `ActionRow` (extend `viewMode` if necessary). Please, avoid code duplication.

frontend/src/components/transactions/ActionsList.tsx
```diff
@@ -164,9 +164,8 @@ export default class extends React.Component<Props> {
 
           .action-row-title {
             font-family: BentonSans;
-            font-size: 14px;
+            font-size: 1rem;
```
kcole16: Why this change? Makes the font quite a bit larger

icerove: because this makes the logo and text not in the same baseline and could not center

frontend/src/components/transactions/ActionRow.tsx
```diff
@@ -40,152 +51,75 @@ export default class extends React.Component<Props, State> {
...
+                      transaction={transaction}
+                      actionKind={actionKind}
+                      actionArgs={actionArgs}
+                    />
+                  </Col>
+                </Row>
+              </Col>
+            </Row>
+          </Col>
+        </Row>
```
frol: There is quite a duplication with the `else` branch. Let's merge it.

frontend/src/components/transactions/ActionRow.tsx
```diff
@@ -40,152 +51,75 @@ export default class extends React.Component<Props, State> {
...
-          }
-
-          .action-row-timer {
-            font-family: BentonSans;
-            font-size: 12px;
-            color: #999999;
-            font-weight: 100;
-          }
+    let actionRow;
+    if (detail || batch) {
```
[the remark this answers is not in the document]
icerove: we need two flags, batch is for batch the whole transactions and show batch line like this: 
image

but for detail page, we will only display the basic line just no details for every actions on the transaction detail page, like this: 
image

frontend/src/components/transactions/ActionRow.tsx
icerove: I tried, but I found it is not worth to merge, because it is not actually a couple duplication, so many different place to split if we really merge them into one. See the <Col> you wrote for the format of the cell. if we just show message and no other message, what I delete for simple actionRow is the simplest.

kcole16:
@frol Think this is ready to merge?

frol:
I think it is ready to merge.

@icerove Please, review my changes.

frol:
I have been considering better naming for `ActionRowBlock` and realized that we use a similar block layout for AccountRow and BlockRow, so we should consider moving `ActionRowBlock` to utils, rename it to something more generic, and reuse it everywhere.

icerove:
I will merge it and open a new PR to make changes for `ActionRowBlock`

Issue #144, opened 2019-12-09
Bug in live-updates on the dashboard


frol:
## Story

![image](https://user-images.githubusercontent.com/304265/70470192-2a667200-1ad3-11ea-93e5-25f114b54fe7.png)

Keeping the Dashboard in a background for a while (in my case it was over an hour) and then coming back leads to an "infinite" long list of recent blocks, and the browser tag gets completely unresponsive. I don't have a guess why that could have happened.

## Acceptance Criteria

- [ ] Bug is resolved
- [ ] Mobile ok
- [ ] Tested

icerove:
Does transactions also have this bug?

icerove:
I repro the bug and found out that it is highly probably because the react-flip-move have some issues to finish the animation. I inspect that the real block it will display is much larger than the not correct showing blocks and after few seconds it goes well.

frol:
I have not tested nor I observed it (it might be just due to the low traffic of transactions)

frol:
Maybe there is a parameter to flip-move that can help it. Alternatively, we can detect “out of focus tab” and just render without flip-move. In fact, it makes sense to do fewer requests when the tab is inactive.

icerove:
>In fact, it makes sense to do fewer requests when the tab is inactive.
Do you mean to do lazy request?

frol:
I mean that we don't need to send any requests while the tab is not active as users won't see the changes anyway, but the browser will have this constant load every 10 seconds.

frol:
The problem is getting reported by community: https://github.com/nearprotocol/stakewars/issues/96

We should fix this sooner rather than later.

mikedotexe:
Interesting, some folks reporting issues with Sequelize having a memory leak, comments continuing until last month:
https://github.com/sequelize/sequelize/issues/9276#issuecomment-438687763

Also, I did notice that SQLite is being used for production, which I found odd at first glance. 
https://github.com/nearprotocol/near-explorer/blob/master/backend/config/database.js#L15

I ran near-explorer in Docker for about 30 minutes, looked at the filesystem (with `docker exec -it CONTAINER_NAME /bin/sh`) and saw that the file had grown to 12.6 megs. Not a huge deal, but it's obviously going to grow fast. I guess I've never seen SQLite used to power a database that's written to so often. Perhaps a solution might be to change the [`dialect` to MariaDB](https://sequelize.org/master/manual/dialects.html#mariadb)?

[...]

PR #149, opened 2019-12-13
Fix block bug

icerove:
frontend fix the #144


frol:
I confirm that disabling FlipMove for blocks list resolves the issue. I wonder what is going on.

icerove:
@frol should we fix this bug first and then think of other method to improve this since now it is a bug not just an issue?

frol:
I would prefer to see if there is a quick fix without UX degradation (ideally, fix flip-move, otherwise try other solutions). If that is impossible to achieve in a timely manner, let’s merge this (please, fix the layout for big screen as the blocks list renders in a single column instead of two columns for me)

Issue #142, opened 2019-12-06
List of Accounts


kcole16:
## Story

As a user, I want to see a list of accounts on NEAR, so I can explore common usage patterns of NEAR and also understand roughly how many unique users there are.

## Acceptance Criteria

- [ ] User can view a list of accounts by selecting the "Accounts" tab in the navbar
- [ ] Each account line item contains the relevant details
  - [ ] Account ID
  - [ ] NEAR balance
  - [ ] Date created
- [ ] Clicking on a account row redirects to the Account Detail page
- [ ] Mobile ok
- [ ] Tested

## Assets

[Designs](https://projects.invisionapp.com/share/TCS6KG8G9V7#/screens/365131593)

icerove:
done

PR #151, opened 2019-12-18
Account list

icerove:
deploy the list of account #142


backend/models/Transaction.js
```diff
@@ -40,7 +40,7 @@ module.exports = (sequelize, DataTypes) => {
     {
       tableName: "transactions",
       underscored: true,
-      timestamps: false,
+      timestamps: true,
```
frol: This is not useful as synchronization queries nearcore in "arbitrary" order, and thus the timestamps added on the DB level are just useless. Let's discuss what you want to achieve, so I can help you.

frontend/src/pages/accounts/[id].jsx
```diff
@@ -4,7 +4,7 @@ import React from "react";
 
 import TransactionIcon from "../../../public/static/images/icon-t-transactions.svg";
 
-import AccountsApi from "../../libraries/explorer-wamp/accounts";
+import AccountApi from "../../libraries/explorer-wamp/accounts";
```
Comment (frol) [inline frontend/src/pages/accounts/[id].jsx]: Let's rename it to `AccountsApi` in the `libraries/explorer-wamp/accounts` to be consistent with the existing *Api naming.

frontend/src/components/accounts/AccountsList.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import * as A from "../../libraries/explorer-wamp/accounts";
```
frol: For what it worth, I used `* as T` where `T` states for "types" (not "transactions") ^_^

backend/models/Transaction.js
icerove: we need the time when new account created.

frontend/src/pages/accounts/[id].jsx
Comment (icerove) [reply frontend/src/pages/accounts/[id].jsx]: got it

icerove:
image
@frol @kcole16 We now have the create time for account but what is the address? Do we still deploy the gray line in the detail page?

kcole16:
> @frol @kcole16 We now have the create time for account but what is the address? Do we still deploy the gray line in the detail page?

Let's add the "Created Time" to the existing line. 

We don't really have addresses to show here, so shouldn't include that.

frol:
The address can be a transaction hash when the CreateAccount action has been performed.

backend/src/sync.js
```diff
@@ -85,6 +86,18 @@ async function saveBlocks(blocksInfo) {
...
               );
+              models.Account.bulkCreate(
+                blockInfo.transactions.map(tx => {
+                  return {
+                    accountId: tx.receiver_id,
+                    transactionHash: tx.hash,
+                    timestamp
+                  };
+                })
+              ).catch(function(err) {
```
frol: If account creation fails here, we will never get back to it in the future as we just let it continue. We should implement a proper fallback or at least do not mask the issue with `catch`, and let it crash.

backend/src/sync.js
```diff
@@ -85,6 +86,18 @@ async function saveBlocks(blocksInfo) {
...
+                console.error(err);
+              });
+              return;
```
frol: This should return a promise. Thus, combine the two above `bulkCreate` calls with `Promise.all([...])` and return that.

frontend/src/components/accounts/AccountRow.tsx
```diff
@@ -0,0 +1,110 @@
...
+}
+
+export default class extends React.Component<Props, State> {
+  state: State = {
+    amount: ""
+  };
+
+  _getDetail = async () => {
+    const detail = await new AccountsApi().getAccountInfo(this.props.accountId);
+    this.setState({ amount: detail.amount });
```
frol: Given you only need the amount, use `AccountsApi().queryAccount(this.props.accountId)` instead of `getAccountInfo`.

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -1,7 +1,8 @@
 import { ExplorerApi } from ".";
 
-interface AccountId {
+export interface AccountId {
```
frol: It seems to make sense to rename this to just `Account`.

backend/src/sync.js
icerove: changed

backend/src/sync.js
icerove: changed

frontend/src/libraries/explorer-wamp/accounts.ts
icerove: but we already name this `export type Account = AccountId & AccountStats & AccountInfo;` is it good to name interface and type same?

backend/src/sync.js
```diff
@@ -62,29 +62,41 @@ async function saveBlocks(blocksInfo) {
...
+                ),
+                models.Account.bulkCreate(
+                  blockInfo.transactions.map(tx => {
+                    return {
+                      accountId: tx.receiver_id,
+                      transactionHash: tx.hash,
+                      timestamp
+                    };
+                  })
+                )
```
frol: We should only add an account to the table and set the `tx.hash` when there is `CreateAccount` action in the transaction.

frontend/src/libraries/explorer-wamp/accounts.ts
frol: Hmm, my bad. Well, we should come up with a better name here as it does not return only AccountId anymore, and it was a bad name in the first place. The only name I, currently, have in mind is `AccountBasicInfo`.

frol:
LGTM.

Let's fix TypeScript errors before we merge:

```
src/components/accounts/AccountsList.tsx(6,15): error TS2694: Namespace '"/home/runner/work/near-explorer/near-explorer/frontend/src/libraries/explorer-wamp/accounts"' has no exported member 'AccountId'.
```

(see CI errors for more details)

icerove:
@frol I fix the unit test but for jest snapshot, I am in different time zone now, and it shows different time. How do you fix this problem?

icerove:
there is no help when I change my time zone to pacific timezone, really interesting.

Issue #154, opened 2019-12-31
[BUG] Incorrect info on account page


ilblackdragon:
When switching from one account to another, the transaction list doesn't update.

# Steps to reproduce

- Open https://explorer.nearprotocol.com/accounts/near
- Click on `@corey` to switch to that page
- Observe that the top panel changed, but not the list of transactions

icerove:
@ilblackdragon  It is not a bug but you should wait for several seconds for that change. I can add a loading icon for that. But it is still okay for now.

frol:
It is definitely a UX bug, so we should add a loader there

PR #155, opened 2020-01-07
Account page add loading

icerove:
fix issue #154


icerove:
there is some issue with my git, I will fix this

icerove:
componentDidUpdate(prevProps: Props) {
    if (this.props.accountId !== prevProps.accountId) {
      this.setState({ transactions: null });
    }
  }

PR #164, opened 2020-01-28
Fixed CI to be all green

frol:


frol:
@icerove Heads up! I have updated CI configs, so we should be all green and maintain that state from now on.

PR #166, opened 2020-02-06
Pagination tx

### LINKED ISSUE #61: TX List: Search and pagination
We already have this for blocks, so we should also have this for transactions.

icerove:
fix #61 and #173


frol:
Imagine you have 1 million transactions and accounts. The current implementation will query 15 latest, then 30 latest, then 45 latest, ... 1 million latest at a single query. This implementation is fine as a temporary solution. The current limitation is on the backend side, as I mentioned: https://github.com/nearprotocol/near-explorer/issues/61#issuecomment-574236190. It is still better than nothing, so let's have some code quality review and merge it, and refactor it later once the new backend indexing service is built.

frontend/src/components/transactions/Transactions.tsx
```diff
@@ -64,28 +73,70 @@ export default class extends React.Component<Props, State> {
...
+    const bottom = this._isAtBottom();
+    if (bottom) {
+      document.removeEventListener("scroll", this._onScroll);
+      await this._loadTransactions();
+      document.addEventListener("scroll", this._onScroll);
+    }
+  };
+
+  _loadTransactions = async () => {
+    this.setState({ stop: this.state.stop + this.props.limit, loading: true });
```
frol: `stop` variable raised some questioning in my mind. At first, I just didn't like the naming, but on the second thought I realized that with the current design it would be much cleaner just using `this.state.transactions.length` and query for `this.state.transactions.length + this.props.limit`.

Another naming item is `this.props.limit`. It looses its semantic in this case, so we should rename it to `paginationSize` or something.

**The following is a valid concern (I wrote it before I realized the depth of the issue), but irrelevant given the above.**

`setState` function is not a sync function, so you should never assume:
1. that `this.state` is up to date (pass a function into `setState`, which should expect the previous state and produce a new one:

```js
this.setState(({ stop }) => ({ stop: stop + this.props.limit, loading: true }))
```
2. that consequent `setState` will be executed in the right order (in theory, the latter `loading: false` may end up executed before this `loading: true`), so you should use a callback as a second argument to the first `setState`. I suggest [creating a promise there](https://stackoverflow.com/a/47019631/1178806) and awaiting for it and `fetchTransaction` with `await Promise.all([...])` and only then call to `setState({ loading: false })`

backend/src/sync.js
```diff
@@ -83,6 +83,7 @@ async function saveBlocks(blocksInfo) {
                       signerPublicKey: tx.signer_public_key || tx.public_key,
                       signature: tx.signature,
                       receiverId: tx.receiver_id,
+                      txHeight: blockInfo.header.height + tx_index,
```
frol: I am sorry, but this is an invalid assumption. If the current block height is 10, and it has 2 transactions, then we will have transactions with txHeight 10 and 11, but then once the next block comes and has at least one transaction, we have a collision of txHeight 11.

icerove:
fine, I will separate them into three, through and if you have better idea, just make high level component to let this better

PR #168, opened 2020-02-06
update argument display

### LINKED ISSUE #167: Contract arguments are displayed broken
Here is the example: https://explorer.nearprotocol.com/transactions/5dxrDYuPnrNnMUkZpKfRSHiKsvcAodaziQeoDv2oy971

> Called method: send_message with args {"receiver_id","app_id","message",\"subject\",\"content\"} in contract: @sheeet

On the dashboard and the transactions list it should be in a short form:

> Called method: send_message in contract: @sheeet

and on the transaction details page:

> Called method: send_message in contract: @sheeet
> <pre>{base64-decoded args}</pre>

The main motivation is that the args can get really logs.

NOTE: there is no need to do anything but base64 decoding!

icerove:
*Resolves #167*


icerove:
> Only display args on the transaction details page
why?

frontend/src/components/transactions/ActionMessage.tsx
```diff
@@ -7,6 +7,7 @@ export interface Props<A> {
   actionKind: keyof TransactionMessageRenderers;
   actionArgs: A;
   transaction: T.Transaction;
+  showArg?: boolean;
```
frol: This way you leak the abstraction. Let's rename it to a more general thing `showDetails`.

frontend/src/components/transactions/ActionMessage.tsx
```diff
@@ -48,25 +49,19 @@ const transactionMessageRenderers: TransactionMessageRenderers = {
...
     return (
       <>
-        {`Called method: ${
-          actionArgs.method_name
-        } with args {${parameter}} in contract: `}
+        {`Called method: '${actionArgs.method_name}' in contract: `}
         <AccountLink accountId={receiverId} />
+        {showArg ? (
+          <pre>
+            <i>Arguments in method: {args}</i>
```
frol: Let's move the caption "Arguments in method" outside of the `<pre>` block and make sure we have text wrapping (https://stackoverflow.com/questions/248011/how-do-i-wrap-text-in-a-pre-tag) for the code. I don't feel like we need italic here.

frontend/src/components/transactions/ActionMessage.tsx
```diff
@@ -48,25 +49,19 @@ const transactionMessageRenderers: TransactionMessageRenderers = {
...
     return (
       <>
-        {`Called method: ${
-          actionArgs.method_name
-        } with args {${parameter}} in contract: `}
+        {`Called method: '${actionArgs.method_name}' in contract: `}
         <AccountLink accountId={receiverId} />
+        {showDetails ? (
+          <>
+            Arguments in method: <pre> {args} </pre>
```
frol: We don't need those extra spaces around `args`: `<pre>{args}</pre>`

frol:
@icerove deleting a legal test is not a solution!

icerove:
BTW, I never delete your existing test, just delete one new test I write but find out there is some issue with it and not good to hold. Since you asked to fix and add, I change it back

frontend/src/pages/transactions/[hash].jsx
```diff
@@ -50,6 +50,8 @@ export default class extends React.Component {
             <ActionsList
               actions={this.props.actions}
               transaction={this.props}
+              detalizationMode="minimal"
```
Comment (frol) [inline frontend/src/pages/transactions/[hash].jsx]: Good catch!

PR #172, opened 2020-02-08
Fix #169

### LINKED ISSUE #169: Transaction deposit value should be correct when 2+ actions have attached deposit
## Story

https://github.com/nearprotocol/near-explorer/blob/605dc80cf9f17e0640295f1712f0f95936c3a107/frontend/src/components/transactions/TransactionDetails.tsx#L28-L40

This way having two transaction actions with `deposit` will override each other. We should accumulate a total deposit. Keep in mind that it is a big number (this is why it is a string in the first place), so you should use big numbers library (see what Wallet uses for that).

## Acceptance Criteria

- [x] The deposit is accumulated
- [x] Tested with big numbers

icerove:


frontend/src/components/transactions/ActionMessage.tsx
```diff
@@ -51,16 +51,8 @@ const transactionMessageRenderers: TransactionMessageRenderers = {
...
-      parameter = "";
-    } else {
-      const argums = args.slice(1, args.length - 1).split(",");
-      parameter = argums.map(ar => {
-        const index = ar.indexOf(":");
-        return ar.slice(0, index);
-      });
-    }
+    const args_object = JSON.parse(args);
+    let parameter = Object.keys(args_object);
```
frol: This is not related to the issue, right? We address it separately, so revert these changes for this PR.

frontend/src/components/transactions/TransactionDetails.tsx
```diff
@@ -36,9 +38,24 @@ export default class extends React.Component<Props, State> {
...
+      .map(dp => new Big(dp))
+      .reduce((current, dp) => dp.plus(current), 0)
+      .c.join("");
+    if (bigIntDeposit.length < max) {
+      const tail = "0".repeat(max - bigIntDeposit.length);
+      const deposit = bigIntDeposit.concat(tail);
+      this.setState({ deposit });
+    } else {
+      this.setState({ deposit: bigIntDeposit });
+    }
```
frol: I fail to understand the intension of this code. Could you clarify the intent or simplify the code?

icerove:
fail to revert, so open another one

PR #175, opened 2020-02-13
Fix 169

### LINKED ISSUE #169: Transaction deposit value should be correct when 2+ actions have attached deposit

icerove:
fix #169
<fix>
<description> 
fix the bug for not correctly showing the sum of deposit

<test>
test with no deposit
test with one deposit
test with 1+ deposit


frontend/src/components/transactions/TransactionDetails.tsx
```diff
@@ -36,9 +39,30 @@ export default class extends React.Component<Props, State> {
...
+      .c.join("");
+    /* give back string "20005" */
+    if (bigIntDeposit.length < max) {
+      const tail = "0".repeat(max - bigIntDeposit.length);
+      const deposit = bigIntDeposit.concat(tail);
+      /* give back string "20005000" */
+      this.setState({ deposit });
+    } else {
+      this.setState({ deposit: bigIntDeposit });
+    }
```
frol: There is `.toString()` method for Big number. Why don't we use it?

icerove: Actually .toString() will return "1.0000001e+25" instead of "10000000000000000000000000" which is not accepted by nearlib function for Balance component. We should use .toFixed(). I will change this.

frontend/src/components/transactions/TransactionDetails.tsx
```diff
@@ -16,29 +18,34 @@ export interface Props {
...
+      .map(action => {
+        let actionKind: keyof T.Action;
+        let actionArgs: any;
+        if (typeof action === "string") {
+          actionKind = action;
+          actionArgs = {};
+        } else {
+          actionKind = Object.keys(action)[0] as keyof T.Action;
+          actionArgs = action[actionKind];
+        }
```
frol: Let's isolate these changes in the `src/libraries/explorer-wamp/transactions.ts`, so we avoid duplication here and in the ActionRow component.

I suggest we preprocess the actions from:

```
[actionKind[string] | {actionKind[string]: actionArgs[object]}]
```

to

```
[
{
  "kind": actionKind[string],
  "args": actionArgs[object],
}
]
```

Currently, we can have:

```js
[
      "CreateAccount",
      {
        Transfer: {
          deposit: "10000000000000000000"
        }
      },
      {
        AddKey: {
          access_key: {
            nonce: 0,
            permission: "FullAccess"
          },
          public_key: "ed25519:8LXEySyBYewiTTLxjfF1TKDsxxxxxxxxxxxxxxxxxx"
        }
      }
]
```

I want to unify it to:

```js
[
      {
        "kind": "CreateAccount",
        "args": null,
      },
      {
        "kind": "Transfer",
        "args": {
          deposit: "10000000000000000000"
        }
      },
      {
        "kind": "AddKey",
        "args": {
          access_key: {
            nonce: 0,
            permission: "FullAccess"
          },
          public_key: "ed25519:8LXEySyBYewiTTLxjfF1TKDsxxxxxxxxxxxxxxxxxx"
        }
      }
]
```

So let's do this logic right when we receive the data: https://github.com/nearprotocol/near-explorer/blob/5c5d9c3fb3cd34457c9719ae8f7123635e02c3d5/frontend/src/libraries/explorer-wamp/transactions.ts#L166

icerove:
see the test and tell me whether it is the thing in your mind. Or what other test do we still want?

frontend/package.json
```diff
@@ -38,6 +38,8 @@
     "@types/enzyme": "^3.10.3",
     "@zeit/next-css": "^1.0.1",
     "autobahn": "19.7.3",
+    "big.js": "^5.2.2",
```
frol: We don't neet it. We use `bn.js` instead.

frontend/src/components/transactions/ActionRow.tsx
```diff
@@ -33,17 +31,8 @@ export default class extends React.Component<Props, State> {
...
-    if (typeof action === "string") {
-      actionKind = action;
-      actionArgs = {} as T.Action;
-    } else {
-      actionKind = Object.keys(action)[0] as keyof T.Action;
-      actionArgs = action[actionKind] as T.Action;
-    }
-
+    let actionKind = action.kind as keyof T.Action;
+    let actionArgs = action.args;
```
frol: No need to have a local variable anymore, just pass `action.args` directly to the component props.

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -57,12 +57,17 @@ export interface Action {
   DeleteKey: DeleteKey;
 }
 
+export interface ActionWrapper {
+  kind: string;
```
frol: Is there any problem with defining it `keyof Action` type instead of generic string?

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -57,12 +57,17 @@ export interface Action {
   DeleteKey: DeleteKey;
 }
 
+export interface ActionWrapper {
```
frol: Since we are going to use this interface down the line, I think it is fair to rename it to `Action` and the `Action` we have above rename to `RpcAction`.

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -167,7 +172,20 @@ export default class TransactionsApi extends ExplorerApi {
...
+              const kind = Object.keys(action)[0];
+              _uniformActions.push({
+                kind: kind,
+                args: action[kind]
+              });
+            } else {
+              _uniformActions.push({ kind: action, args: {} });
+            }
+          });
+          transaction.actions = _uniformActions;
```
frol: Here is a better way to express it:

```ts
          const actions = transactionExtraInfo.transaction.actions;

          transaction.actions = actions.map((action: RpcAction | string) => {
            if (typeof action === "string") {
              return { kind: action, args: {} };
            } else {
              const kind = Object.keys(action)[0];
              return {
                kind,
                args: action[kind]
              };
            }
          });
```

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -136,7 +141,7 @@ export default class TransactionsApi extends ExplorerApi {
       const transactions = await this.call<
         (TransactionInfo & (StringActions | Actions))[]
```
frol: So, since we don't query actions here now, we should just expect `TransactionInfo[]` type here. And as a result, we can remove `StringActions` interface.

frontend/src/components/transactions/ActionRow.tsx
```diff
@@ -33,17 +31,8 @@ export default class extends React.Component<Props, State> {
...
-    let actionArgs: T.Action;
-    if (typeof action === "string") {
-      actionKind = action;
-      actionArgs = {} as T.Action;
-    } else {
-      actionKind = Object.keys(action)[0] as keyof T.Action;
-      actionArgs = action[actionKind] as T.Action;
-    }
-
+    let actionKind = action.kind as keyof T.Action;
```
frol: Once we address the comments in `explorer-wamp/transactions.ts`, we should not have a need for this type casting.

frontend/package.json
icerove: deleted

frontend/src/components/transactions/ActionRow.tsx
icerove: deleted

frontend/src/libraries/explorer-wamp/transactions.ts
icerove: changed back

frontend/src/libraries/explorer-wamp/transactions.ts
icerove: tried but has error 
image

frol: And what happens if you do `args: action[kind] as RpcAction[keyof RpcAction]`? (is it even valid? Anyway, find the way to force the types working here as the code is much cleaner and easier to understand this way.)

icerove: changed a little, have a look

PR #178, opened 2020-02-15
Pagination tx/account page

### LINKED ISSUE #173: pagination for accounts list page and change tx/block page same format
## Story
Implement pagination to account list page
## Acceptance Criteria

- [ ] Mobile ok
- [ ] Tested

icerove:
fix #173 and #61


frontend/src/components/accounts/Accounts.tsx
```diff
@@ -4,42 +4,138 @@ import AccountsApi, * as A from "../../libraries/explorer-wamp/accounts";
...
+    }
+    const count = await this._accountApi.getAccountLength();
+    return count;
+  };
 
   render() {
-    const { accounts } = this.state;
-    if (accounts === null) {
+    const { accounts, loading } = this.state;
+    if (accounts === []) {
```
frol: Sadly, `[] === []` is `false` is JS world. In fact, since you have added the `loading` state variable, we can use it here.

In fact, we can completely remove this `if` block as it should work just fine without it.

frontend/src/components/transactions/Transactions.tsx
```diff
@@ -64,28 +68,101 @@ export default class extends React.Component<Props, State> {
...
+    } else {
+      transactions = (await this._transactionsApi.getTransactions({
+        signerId: this.props.accountId,
+        receiverId: this.props.accountId,
+        blockHash: this.props.blockHash,
+        tail: this.props.reversed,
+        limit: this.state.transactions.length
+      })) as T.Transaction[];
+    }
+    this.setState({ transactions });
```
frol: This is a code duplication for nothing. Just extract `limit` variable.

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -74,10 +74,22 @@ export default class AccountsApi extends ExplorerApi {
     }
   }
 
-  async getAccounts(): Promise<AccountBasicInfo[]> {
+  async getAccounts(
+    limit: number = 15,
+    lastIndex: number = -1
```
frol: Last index... It is really misleading.

frontend/src/components/accounts/Accounts.tsx
```diff
@@ -4,42 +4,138 @@ import AccountsApi, * as A from "../../libraries/explorer-wamp/accounts";
...
   fetchAccounts = async () => {
     if (this._accountApi === null) {
       this._accountApi = new AccountsApi();
     }
-    const accounts = await this._accountApi.getAccounts();
+    let accounts;
+    if (this.state.accounts.length === 0) {
+      accounts = await this._accountApi.getAccounts(this.props.limit);
+    } else {
+      accounts = await this._accountApi.getAccounts(this.state.accounts.length);
```
frol: What is the exact point of re-querying the same number of the latest accounts?

PR #192, opened 2020-02-20
fix small amount display

### LINKED ISSUE #152: Correctly display small amounts of NEAR (<0.00001)
### Acceptance Criteria

- [x] Amounts of NEAR smaller than 0.00001 are displayed as "<0.00001 N"
- [x] On hover, the amount in yoctoNEAR is displayed "1000000 yoctoN"

**Original Thread**
> We should fix `formatNEAR` to actually reflect that the amount is not 0, but rather very small.

> @janedegtiareva Can nearlib format small yoctoNEAR amounts in a human-friendly format, e.g. `1 yoctoN`? We used to have `<0.00001 N` here. Do you think it is reasonable?

> @icerove @kcole16 It seems that to make the information non-confusing and useful, we should present both `<0.00001 N (1 yoctoNEAR)` (we may want to show the exact amount, e.g. "1 yoctoNEAR", as an on-hover popup)

> _Originally posted by @frol in https://github.com/nearprotocol/near-explorer/pull/146_

icerove:
fix #152


frontend/src/components/utils/Balance.jsx
```diff
@@ -8,7 +8,21 @@ const Balance = ({ amount }) => {
...
+        .precise {
+          display: none;
+        }
+        .rough:hover + .precise {
+          display: block;
+          color: #20c997;
+        }
+      `}</style>
+    </>
+  );
```
frol: Unless we have a specific design, it is better to stick to available solutions. There is a tooltip component in react-bootstrap which we already use. https://react-bootstrap.github.io/components/overlays/#tooltips

frontend/src/components/transactions/ActionMessage.tsx
```diff
@@ -101,12 +101,19 @@ const transactionMessageRenderers: TransactionMessageRenderers = {
     actionArgs: { deposit }
   }: Props<T.Transfer>) => (
     <>
-      {`Transferred ${formatNEAR(deposit)} Ⓝ to `}
+      {`Transferred `}
+      <span>
```
frol: The `span` looks useless to me.

frontend/src/components/utils/Balance.jsx
```diff
@@ -2,13 +2,21 @@
 /// https://github.com/nearprotocol/near-wallet/blob/41cb65246134308dd553b532dfb314b45b38b65c/src/components/common/Balance.js
 
 import { utils } from "nearlib";
+import { OverlayTrigger, Tooltip, Button } from "react-bootstrap";
```
frol: Button is not used.

icerove: deleted

icerove:
> https://near-explorer-frontend-pr-192.onrender.com/accounts/illia.near
> 
> I don't feel like it is user-friendly. The tooltips display positions are odd (both in the top "Balance" box and on the Action line).
> 
> @icerove @kcole16 do you like how it looks now?

it only has four direction, left, top, right and bottom. which one do you prefer?

frol:
Here is how it looks in Wallet:

image

frol:
I like it when the tooltip is near to the item and visually attached to it. Currently, the positioning is off:

image

And it breaks the lines:
image

Also, we should make the formatting at least somewhat useful for the users (at the very least, we need commas there; the further improvements like showing microNEAR instead of ridiculous yoctoNEAR precision, but that is another story, and requires a general decision across the AppLayer).

icerove:
When I build Balance.js in the wallet, it used to have commas to three digits to display but when we use nearlib function it disappears

icerove:
see the latest version

icerove:
if you want microNEAR now, I can fix it

frontend/src/components/utils/Balance.jsx
```diff
@@ -9,10 +9,11 @@ const Balance = ({ amount }) => {
     throw new Error("amount property should not be null");
   }
   let amountShow = convertToShow(amount);
+  let amountPrecise = formatPreciseNEAR(amount);
   return (
     <OverlayTrigger
       placement={"bottom"}
-      overlay={<Tooltip>{amount} yokto Ⓝ</Tooltip>}
+      overlay={<Tooltip>{amountPrecise} yokto Ⓝ</Tooltip>}
```
frol: It is "yocto" with "c" instead of "k", and let's remove the space between yocto and N: "yoctoⓃ"

Issue #191, opened 2020-02-19
Update "Address" name and description on account detail page


kcole16:
On `/account/<accountname>`, the `Address` attribute is confusing, as it is actually the hash of the transaction that created the account. The attribute name should be changed, and the TX hash should link to the detail page.

## Acceptance Criteria

- [x] `Address` is changed to `Creation Hash`
- [x] The hash redirects to the corresponding TX detail page
- [x] Mobile ok
- [x] Tested

PR #203, opened 2020-02-25
Fix account detail page display

### LINKED ISSUE #190: Break "Storage" into two fields on Account detail page
`Storage (Paid / Used` is very misleading currently. Breaking this metric into two descriptive fields will better explain the concept to users.

## Acceptance Criteria

- [x] `Storage (Used/Paid)` is replaced by
  - [x] `Storage Used (Bytes)` : `300 B`
  - [x] `Storage Last Paid` :  `#<block height>`
- [x] Block height links to corresponding block detail page
- [x] Mobile ok
- [x] Tested

icerove:
fix #190 and #191


frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -6,16 +6,19 @@ import { Account } from "../../libraries/explorer-wamp/accounts";
...
+import BlockLink from "../utils/BlockLink";
+import TransactionLink from "../utils/TransactionLink";
 
 export interface Props {
   account: Account;
+  blockHash: string;
 }
 
-export default ({ account }: Props) => {
+export default ({ account, blockHash }: Props) => {
```
frol: Let's stop for a moment and think about the context.

Imagine you are opening this file the very first time in your life and you know that it is an AccountDetails component. Do you expect account infromation being passed via props? Probably, yes. Do you expect some random `blockHash` being passed? I would certainly cast "WTF?!".

I want to ask you to think about this puzzle yourself.

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -46,16 +49,27 @@ export default ({ account }: Props) => {
...
+          />
+        </Col>
+        <Col md="2">
+          <CardCell
+            title="Last Paid"
             imgLink="/static/images/icon-m-block.svg"
-            text={`${account.storageUsage.toLocaleString()} / ${account.storagePaidAt.toLocaleString()}`}
+            text={
+              <BlockLink blockHash={blockHash}>
+                {account.storagePaidAt}
```
frol: As a result, it becomes really disconnected.

frol: When we refer to block height, we use `#` in front of the number.

frontend/src/components/accounts/AccountDetails.tsx
icerove: change the account type include the hash

frontend/src/components/accounts/AccountDetails.tsx
icerove: added

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -6,16 +6,19 @@ import { Account } from "../../libraries/explorer-wamp/accounts";
 
 import Balance from "../utils/Balance";
 import CardCell from "../utils/CardCell";
+import BlockLink from "../utils/BlockLink";
+import TransactionLink from "../utils/TransactionLink";
 
 export interface Props {
   account: Account;
+  blockHash: string;
```
frol: Remove it, it is not used.

frontend/src/components/accounts/__tests__/AccountDetails.test.tsx
```diff
@@ -18,6 +18,7 @@ describe("<AccountDetails />", () => {
             timestamp: Number(new Date(2019, 1, 1)),
             address: "EVvWW1S9BFaEjY1JBNSdstb7ZTtTFjQ6cygkbw1KY4tL"
           }}
+          blockHash="FMwFGEPnXdZxKP9rJ9d7WPrK4pVtMy31szMioC5RNQgk"
```
frol: Remove it ^

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -4,6 +4,7 @@ export interface AccountBasicInfo {
   id: string;
   timestamp: number | string;
   address: string;
+  blockHash?: string;
```
frol: What is a block hash of an account?

Well, I believe, the best solution here is to make the `storagePaidAt` block height conversion to block hash right in the Account Details component just using the component state for that. To avoid blocking on waiting for the block hash, you can render without a `<BlockLink>` first (just render as text `#100500`) and once we resolve the `storagePaidAtBlockHash`, we can render that cell with `<BlockLink>`.

Thus, this change should be reverted.

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -46,16 +49,23 @@ export default ({ account }: Props) => {
...
+            imgLink="/static/images/icon-storage.svg"
+            text={`${account.storageUsage.toLocaleString()} B`}
+          />
+        </Col>
+        <Col md="2">
+          <CardCell
+            title="Last Paid"
             imgLink="/static/images/icon-m-block.svg"
-            text={`${account.storageUsage.toLocaleString()} / ${account.storagePaidAt.toLocaleString()}`}
+            text={`#${account.storagePaidAt.toLocaleString()}`}
```
frol: Let's have `storagePaidAtBlockHash` in the state and fetch the block info (on `componentDidMount` / `componentDidUpdate`) via `BlocksApi` by `account.storagePaidAt` block height to receive the block hash. Here (in the `render` method) we should just show:

* a plain `#${account.storagePaidAt.toLocaleString()}` (when `state.storagePaidAtBlockHash` is not available)
* or
    ```jsx
    <BlockLink blockHash={this.state.storagePaidAtBlockHash}>
      {`#${account.storagePaidAt.toLocaleString()}`}
    </BlockLink>
    ```

icerove: @frol if we do so, the test for AccountDetails will not okay since we have BlocksApi here to load.

frol: @icerove Does the test suite trigger `componentDidMount` or `componentDidUpdate`? Can we use the same workaround as we have in `Transactions.tsx` (define the Api as null and assign it on the first use)?

icerove: will have try

icerove: sorry, but not able to keep test with Api use

icerove: I think the error happen in import not in component 
image

frol: Both import and tests work fine for https://github.com/nearprotocol/near-explorer/blob/321ef470f12c48527cc171f0f4640ae09a366646/frontend/src/components/transactions/Transactions.tsx. Commit the change and let's have a look.

icerove: @frol I do not think we have a test for Transactions.tsx. where can I see that and learn?

frol: Oh well, you are right. Then, we have two options:

1. Make `frontend/src/libraries/config.ts` non-failing (add some checks around `getConfig()` and set some empty values if it is `undefined`)
2. Mock it with jest (`frontend/jest.setup.js` already has some mocks)

icerove:
@frol The difference between your method and my method is that you make `blockHash` to be a state but I pass as props. Let's see what you said before:
>Let's stop for a moment and think about the context.

>Imagine you are opening this file the very first time in your life and you know that it is an >AccountDetails component. Do you expect account infromation being passed via props? >Probably, yes. Do you expect some random blockHash being passed? I would certainly cast >"WTF?!".

is this really solving the problem? Who see this first still confused why blockHash? If not, why cannot pass as props?

frol:
The difference lays in the design.

1. Having a self-contained component, which only receives the data that the parent component has without imposing inadequate requirements, is better. Let's try to use English description of the component:
    a. To display Account Details you need to provide a passport.
    b. To display Account Details you need to provide a passport and a piece of paper where you specify the number of days since your birthday spelled in English.

   I prefer the first variant, which only receives a passport (`account`) via `props`, and the rest can be fetched into the state without bothering the parent component to prepare data for our component.
2. The naming you choose is also super-generic imagine I would define my requirements expressed in 1.b. as: "I need your passport and a date". What "date"? That is exactly my feeling about "blockHash" expected in AccountDetails component props. What "blockHash"? Please, be mindful of naming, otherwise, we will end up in a very bad state fairly quickly.

icerove:
@frol We can have test for Api used component now

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -1,99 +1,146 @@
...
+  fetchBlockHash = async (height: string) => {
+    new BlocksApi()
+      .getBlockInfo(height)
+      .then(block => this.setState({ storagePaidAtBlockHash: block.hash }))
+      .catch(err => console.error(err));
+  };
+
+  componentDidMount() {
+    this.fetchBlockHash(this.props.account.storagePaidAt.toString());
+  }
```
frol: I think we also need to trigger it on `componentDidUpdate` as otherwise clicking on another account link while on Account Details page will lead to a problem of re-rendering with the `storagePaidAtBlockHash` from the previous account.

PR #210, opened 2020-02-27
Auto refresh handler

### LINKED ISSUE #198: make auto refresh a helper function to clean code
## Story

Explorer has so many auto refresh, make it a helper function

## Acceptance Criteria

- [ ] auto refresh working
- [ ] clean duplicate code
- [ ] Mobile ok
- [ ] Tested

## Assets

- Mockup: 

## NOTES

-

icerove:
fix #198


frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -137,7 +139,10 @@ export default class TransactionsApi extends ExplorerApi {
           ${whereClause.length > 0 ? `WHERE ${whereClause.join(" OR ")}` : ""}
           ORDER BY blocks.height ${filters.tail ? "DESC" : ""}
           LIMIT :limit`,
-        filters
+        {
+          filters,
+          limit
+        }
```
frol: This change looks completely broken to me. It is better to introduce `filters` object argument in the other Api methods, than moving limit out this way.

frontend/src/components/accounts/Accounts.tsx
```diff
@@ -2,44 +2,32 @@ import React from "react";
...
+import autoRefreshHandler from "../utils/autoRefreshHandler";
 import PaginationSpinner from "../utils/PaginationSpinner";
+import AccountsList from "./AccountsList";
 
-export interface Props {}
-
-export interface State {
-  accounts: A.AccountBasicInfo[] | null;
+export interface Props {
+  Lists: A.AccountBasicInfo[];
```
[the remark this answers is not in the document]
icerove: The library really make more generic but what HOC I create is good enough for our implementation. I know it has more function to use but we do not require them, right? I will definitely change the name.

frontend/src/libraries/explorer-wamp/transactions.ts
icerove: @frol  filters are adding the `WHERE` clause to find the data.
limit is for `LIMIT :limit` to fetch the limit data.

image
in above, limit is not the attribute for transaction table, but others are. if mixing them together, actually it is worse to make all api consistent and make sense to SQL query language.

frol: `filters` is an object which has/had various options passed to SQL query, e.g. `signerId`. With this change `:signerId` no longer a valid reference, as it is now nested into `filters`, so `:filters.signerId` (and I am not sure if that is even supported by sequelize). Thus, you should do `{ ...filters, limit }` to at least fix the functionality.

Given you still need to blend `limit` into the `filters`, I guess it is pointless to split them in the first place. The naming `filters` is not perfect; we can rename it to `query` or anything better if you have suggestions.

frontend/src/components/accounts/Accounts.tsx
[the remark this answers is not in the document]
frol: 1. use inspiration means that we can take a look at their API design and learn what features they have:
    a. they pass an object, where the key is the props name (this will resolve `Lists` or `items`)
    b. they also pass information about the current state (pending / rejected / fulfilled), which helps to display more useful data
2. There is nothing bad about writing our own HOC if that makes sense. Ours is super tiny, but once we take (1.a) and (1.b) into account, it seems that their component is also of a reasonable size.

Patrick1904:
This is imo good enough for current use case, but I'd consider using [https://www.npmjs.com/package/react-refetch](react-refetch) as it's a pretty sleek, well tested implementation and comes with useful features (future) at a relatively low package size.

icerove:
@frol  I tried to use react-refetch, but its document for typescript seems strange when I use connect, see this:
image

icerove:
what my outer and inner props are 
image

icerove:
also, this said like `it lets you map props to URLs to props.` and I do not see any method to write Api function to it. Do you still want to use it? @frol

frol:
@icerove Yeah, it is a pity that it is so limited to HTTP URLs... I can see that we can hack it together by replacing the fetch function in defaults, but that is going to be a mess. I have not found any better package out there, so it seems we should got with our own implementation.

frontend/src/components/accounts/Accounts.tsx
```diff
@@ -2,44 +2,45 @@ import React from "react";
...
+    let accountsShow = <PaginationSpinner hidden={false} />;
+    if (items.length > 0) {
+      accountsShow = (
+        <FlipMove duration={1000} staggerDurationBy={0}>
+          <AccountsList accounts={items} />
+        </FlipMove>
+      );
     }
-    return <AccountsList accounts={accounts} />;
+    return <>{accountsShow}</>;
```
frol: I like the style we handle spinner in the Nodes.tsx better:

```tsx
const { items } = this.props;
if (items.length === 0) {
  return <PaginationSpinner hidden={false} />;
}
return (
  <FlipMove duration={1000} staggerDurationBy={0}>
    <AccountsList accounts={items} />
  </FlipMove>
);
```

Let's update all the components to use this style.

icerove:
@frol please help me with componentDidUpdate in the Transactions.tsx, now it is really bad but work, I tried many ways to deal with it, and no work

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -0,0 +1,46 @@
...
+      items: []
+    };
+
+    componentDidMount() {
+      this.timer = setTimeout(this.regularFetchInfo, 0);
+    }
+
+    componentWillUnmount() {
+      clearTimeout(this.timer!);
+      this.timer = null;
```
frol: To avoid a race condition when we clear the timeout during `regularFetchInfo` execution we should write it like this:

```ts
const timer = this.timer;
this.timer = null;
if (timer !== null) {
  clearTimeout(timer);
}
```

Imagine the following situation:

1. `clearTimeout` while `regularFetchInfo` is executing
2. `regularFetchInfo` check for `this.timer !== null` and it is not `null` yet
3. `this.timer = null`
4. back to `regularFetchInfo` setting `this.timer = setTimeout(...)`

It is either super rare or even not possible in JS runtime, but I still feel uneasy about the race possibility (timers are the number one reason web pages become resources hogs).

icerove:
> @icerove what is exactly wrong with the Transactions.tsx? Removing the `window.location.reload(true)` does not change the page behavior for me.

open an account detail page or block detail page, when click the link to another account or block, the transactions part will remain the previous account/block, and after fetching tx after 10 seconds, it will change to the correct transactions. ( recommend try test account since it is easy to click to another account page).
we used to have this issue and deal with componentDidUpdate but when I tried to add this to autorefreshHandler, failed.

PR #213, opened 2020-03-03
Tx fee

### LINKED ISSUE #124: Show transaction fees on TX Detail page
## Story

As a user, I want to view the transaction fees for a given transaction, so I can better understand my costs and NEAR's economics.

## Acceptance Criteria

- [x] Total Gas Cost (gas used * gas price)
- [x] Gas Units Used (units of gas)
- [x] Gas Unit Price (price per unit in NEAR) Ⓝ

icerove:
fix #124 

gas price is fetch from database block table gas price column 

gas used is accumulated from transaction outcome and receipt outcome gas_burnt result

tatol gas fee is gas price times gas used

### DIFF
--- a/backend/models/Block.js
+++ b/backend/models/Block.js
@@ -34,7 +34,7 @@ module.exports = (sequelize, DataTypes) => {
         allowNull: false
       },
       gasPrice: {
-        type: DataTypes.BIGINT,
+        type: DataTypes.STRING,
         allowNull: false
       }
     },
--- a/backend/models/Block.js
+++ b/backend/models/Block.js
@@ -34,7 +34,7 @@ module.exports = (sequelize, DataTypes) => {
         allowNull: false
       },
       gasPrice: {
-        type: DataTypes.STRING,
+        type: DataTypes.BIGINT,
         allowNull: false
       }
     },
--- a/backend/src/sync.js
+++ b/backend/src/sync.js
@@ -22,7 +22,7 @@ async function saveBlocks(blocksInfo) {
               totalSupply: blockInfo.header.total_supply || "",
               gasLimit: blockInfo.header.gas_limit || 0,
               gasUsed: blockInfo.header.gas_used || 0,
-              gasPrice: blockInfo.header.gas_price || 0
+              gasPrice: blockInfo.header.gas_price || "0"
             };
           })
         );
@@ -30,19 +30,6 @@ async function saveBlocks(blocksInfo) {
         await models.Chunk.bulkCreate(
           blocksInfo.flatMap(blockInfo => {
             let { chunks } = blockInfo;
-            if (chunks === undefined) {
-              chunks = [
-                {
-                  blockHash: "",
-                  shard_id: 0,
-                  signature: "",
-                  gas_limit: 0,
-                  gas_used: 0,
-                  height_created: 0,
-                  height_included: 0
-                }
-              ];
-            }
             return chunks.map(chunkInfo => {
               return {
                 blockHash: blockInfo.header.hash,
--- a/backend/src/sync.js
+++ b/backend/src/sync.js
@@ -22,7 +22,7 @@ async function saveBlocks(blocksInfo) {
               totalSupply: blockInfo.header.total_supply || "",
               gasLimit: blockInfo.header.gas_limit || 0,
               gasUsed: blockInfo.header.gas_used || 0,
-              gasPrice: blockInfo.header.gas_price || "0"
+              gasPrice: blockInfo.header.gas_price || 0
             };
           })
         );
@@ -33,11 +33,11 @@ async function saveBlocks(blocksInfo) {
             if (chunks === undefined) {
               chunks = [
                 {
+                  blockHash: "",
                   shard_id: 0,
                   signature: "",
                   gas_limit: 0,
                   gas_used: 0,
-                  gas_price: "0",
                   height_created: 0,
                   height_included: 0
                 }
--- a/frontend/src/components/blocks/BlockDetails.tsx
+++ b/frontend/src/components/blocks/BlockDetails.tsx
@@ -27,7 +27,7 @@ export default ({ block }: Props) => {
             </Col>
             <Col md="4">
               <CardCell
-                title="Gas Used (bytes)"
+                title="Gas Used"
                 imgLink="/static/images/icon-m-size.svg"
                 text={block.gasUsed.toLocaleString()}
               />
--- a/frontend/src/components/blocks/BlockDetails.tsx
+++ b/frontend/src/components/blocks/BlockDetails.tsx
@@ -12,7 +12,6 @@ export interface Props {
 }
 
 export default ({ block }: Props) => {
-  console.log(block);
   return (
     <>
       <Row noGutters>
--- a/frontend/src/components/blocks/BlockDetails.tsx
+++ b/frontend/src/components/blocks/BlockDetails.tsx
@@ -12,12 +12,13 @@ export interface Props {
 }
 
 export default ({ block }: Props) => {
+  console.log(block);
   return (
     <>
       <Row noGutters>
         <Col className="block-info-container">
           <Row noGutters className="block-info-header">
-            <Col md="3">
+            <Col md="4">
               <CardCell
                 title="Transactions"
                 imgLink="/static/images/icon-m-transaction.svg"
@@ -25,11 +26,18 @@ export default ({ block }: Props) => {
                 className="border-0"
               />
             </Col>
-            <Col md="2">
+            <Col md="4">
               <CardCell
-                title="Size"
+                title="Gas Used (bytes)"
                 imgLink="/static/images/icon-m-size.svg"
-                text="TODO"
+                text={block.gasUsed.toLocaleString()}
+              />
+            </Col>
+            <Col md="4">
+              <CardCell
+                title="Gas Price"
+                imgLink="/static/images/icon-m-filter.svg"
+                text={block.gasPrice.toLocaleString()}
               />
             </Col>
           </Row>
--- a/frontend/src/components/transactions/TransactionDetails.tsx
+++ b/frontend/src/components/transactions/TransactionDetails.tsx
@@ -21,13 +21,15 @@ export interface State {
   deposit: BN;
   transactionFee: BN;
   gasUsed: BN;
+  gasAttached: BN;
 }
 
 export default class extends React.Component<Props, State> {
   state: State = {
     deposit: new BN(0),
     transactionFee: new BN(0),
-    gasUsed: new BN(0)
+    gasUsed: new BN(0),
+    gasAttached: new BN(0)
   };
 
   setDeposit = () => {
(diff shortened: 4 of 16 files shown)


ilblackdragon:
@icerove "Gas Used" is not measured in N

ilblackdragon:
@icerove it's not bytes, it's just Gas

icerove:
@ilblackdragon what is the unit of gas? really confused if it is not bytes or near N

ilblackdragon:
Gas is the unit. It's an equalizing unit of bandwidth and compute.

See:
- Section 4.1 in https://near.ai/economics 
- https://blockgeeks.com/guides/ethereum-gas/

Also:
- "Total gas cost" probably should say "Transaction fee".
- And there is also "Attached gas" for FunctionCalls, e.g. what user specified as allowed amount of gas for this function.

For example you can see fields here - https://etherscan.io/tx/0xacdc5837d037c3f2c34b870653338b0bc3673db3e62fa24985cfd261f2ea9fe6 (click see More) , "attached gas" they call "gas limit".
They also have ? around each term to familiarize people with concepts, which a good practice.

icerove:
@ilblackdragon  
> And there is also "Attached gas" for FunctionCalls, e.g. what user specified as allowed amount of gas for this function.

what do you want to do with this? make an extra column to put it or add to total gas?

ilblackdragon:
Add an extra column.

backend/models/Block.js
```diff
@@ -34,7 +34,7 @@ module.exports = (sequelize, DataTypes) => {
         allowNull: false
       },
       gasPrice: {
-        type: DataTypes.STRING,
+        type: DataTypes.BIGINT,
```
frol: We need to revert this change.

BIGINT in Sqlite is i64, but NEAR amounts are in u128, so we cannot store them in BIGINT (it would be more efficient to encode these values into varchar[8], but that requires quite some work, so it is easier to have STRING here). It is unfortunate as I would not expect gasPrice to go over i64 limit, but it is safer to match the API and consider fixing this on the nearcore side first.

backend/src/sync.js
```diff
@@ -22,7 +22,7 @@ async function saveBlocks(blocksInfo) {
               totalSupply: blockInfo.header.total_supply || "",
               gasLimit: blockInfo.header.gas_limit || 0,
               gasUsed: blockInfo.header.gas_used || 0,
-              gasPrice: blockInfo.header.gas_price || "0"
+              gasPrice: blockInfo.header.gas_price || 0
```
frol: revert ^

backend/src/sync.js
```diff
@@ -33,11 +33,11 @@ async function saveBlocks(blocksInfo) {
             if (chunks === undefined) {
```
frol: This `if` block can be completely removed as nearcore has the chunks implemented long time ago and the testnet has them as well.

frontend/src/components/blocks/BlockDetails.tsx
```diff
@@ -17,19 +17,26 @@ export default ({ block }: Props) => {
...
               />
             </Col>
-            <Col md="2">
+            <Col md="4">
               <CardCell
-                title="Size"
+                title="Gas Used"
                 imgLink="/static/images/icon-m-size.svg"
-                text="TODO"
+                text={block.gasUsed.toLocaleString()}
```
frol: Given it should be a string, use the same util function that we use for NEAR amounts (I hope it is generic enough).

frontend/src/components/transactions/TransactionDetails.tsx
```diff
@@ -36,11 +40,31 @@ export default class extends React.Component<Props, State> {
...
+      ? new BN(this.props.transaction.transactionOutcome.outcome.gas_burnt)
+      : new BN(0);
+    const receiptFee = this.props.transaction.receiptsOutcome
+      ? this.props.transaction.receiptsOutcome
+          .map(receipt => new BN(receipt.outcome.gas_burnt))
+          .reduce((gasBurnt, currentFee) => gasBurnt.add(currentFee), new BN(0))
+      : new BN(0);
+    const gasUsed = txFee.add(receiptFee);
+    const totalFee = gasUsed.mul(price);
+    this.setState({ totalFee, gasUsed });
```
frol: @bowenwang1996 @ilblackdragon @nearmax I have seen a conversation somewhere mentioning that it is quite hard to compute the fee as you seem to need to collect the gas price from the relevant block where the receipt was issued (which is not the transaction where it gets included). Am I recalling this right? It seems that we cannot match the receipt with the transaction as RPC does not expose any link between them (e.g. receipt does not have a transaction hash or anything like that). :disappointed:

backend/models/Block.js
icerove: I am also really confused about the type of gas price from block. It is string type in the docs but when I write tests in nearlib, it shows it is number. So I do not know which is right

frontend/src/components/transactions/TransactionDetails.tsx
icerove: what I understand is you have said we should "sum" them (transaction outcome and receipt outcome) let's best talk with runtime people to make sure what it is

backend/models/Block.js
frol: @icerove Can you link to the code where it is defined as a number?

frontend/src/components/transactions/TransactionDetails.tsx
frol: My understanding is that we should sum (pseudo-code) `receipt.gas_burnt * receipt.transaction.block.gas_price` (where `receipt.block` is not the same as `receipt.transaction.block`, so we may end up querying `gas_price` from several blocks to compute the transaction fee in NEAR)

bowenwang1996: @frol it should be the sum of (transaction gas burnt + all receipt gas burnt) * gas price

nearmax: To collect the fee in gas you need to sum all gas usage across all receipts issued by the transaction, including the transaction itself. To convert this fee from gas to tokens you only need to know the gas price of the block that includes the transaction, this price is included in every receipt produced by the transaction: https://github.com/nearprotocol/nearcore/blob/4713f5f88a7a57d6b5a13ed1057289858d63f134/core/primitives/src/receipt.rs#L76

frol: It seems that this computation is implemented correctly, but I would fix the variable names:

* `price` -> `gasPrice`
* `txFee` -> `gasBurntByTx`
* `receiptFee` -> `gasBurntByReceipts`
* `totalFee` -> `fee` (or `transactionFee`, but then we also should align `gasUsed` by renaming it to `transactionGasBurnt`, which sounds ridiculous)

Also, I think these computations (deposit as well) are generic enough to move them right into the TransactionApi layer. Thus, you will receive all the computed values "from API".

frontend/src/components/blocks/BlockDetails.tsx
ilblackdragon: gasUsed is in "gas" units, which are not the same as NEAR units.
E.g. we should should just 1 Mgas for 1e6 of gas or 1 Tgas for 1e12 of gas. That is what we use currently in nearcore.

frontend/src/components/transactions/TransactionDetails.tsx
ilblackdragon: `gasAttached` - that either `==` to `gasUsed` for non `FunctionCall` actions or equal to attached gas in the `FunctionCall` Action.

frontend/src/components/blocks/BlockDetails.tsx
icerove: @frol I will create a gas component since we reuse it many times

icerove:
left attach gas not added

frontend/src/components/utils/Gas.tsx
```diff
@@ -0,0 +1,22 @@
...
+export default ({ gas }: Gas) => {
+  let gasShow;
+  if (gas > TGAS) {
+    const tgas = gas.div(TGAS).toString();
+    gasShow = `${tgas} Tgas`;
+  } else if (gas > MGAS) {
+    const mgas = gas.div(MGAS).toString();
+    gasShow = `${mgas} Mgas`;
+  } else {
+    gasShow = gas.toString;
```
frol: Should it be a function call? `gas.toString()`?

frontend/src/components/utils/Gas.tsx
```diff
@@ -0,0 +1,22 @@
+import BN from "bn.js";
+
+interface Gas {
+  gas: BN;
+}
+
+const MGAS = new BN(10 ** 6);
+const TGAS = new BN(10 ** 12);
+
+export default ({ gas }: Gas) => {
```
frol: Since it is a React component, we, conventionally, use `Props` as the type name.

frontend/src/components/transactions/TransactionDetails.tsx
```diff
@@ -103,24 +105,22 @@ export default class extends React.Component<Props, State> {
             <CardCell
               title="Total Gas Cost"
               imgLink="/static/images/icon-m-size.svg"
-              text={<Balance amount={totalFee.toString()} />}
+              text={<Balance amount={transactionFee.toString()} />}
```
frol: One day we should consider using BN type instead of strings throughout our code base.

icerove: nearlib uses string to tackle with amount, so maybe  give issue in nearlib

frol: Yes, that would be also helpful. Especially once we change RPC to encode almost (?) all numbers as strings (we mostly use u64 values and those do not fit into JS number type, so to avoid loss in precision, we should encode as strings and deserialize those into BN for nicer API on near-api-js [fka nearlib] side).

frontend/src/components/transactions/TransactionDetails.tsx
```diff
@@ -36,11 +44,55 @@ export default class extends React.Component<Props, State> {
       })
       .reduce((accumulator, deposit) => accumulator.add(deposit), new BN(0));
     this.setState({ deposit });
+  };
+
+  setAttachedGas = () => {
```
frol: Why is this name?

icerove: you think it is a broken name or you are not clear about the function?
see this https://github.com/nearprotocol/near-explorer/pull/213#discussion_r387437574

frol: I don't see why it is called "set" when it clearly only computes a value and returns it.

frol: I would probably rename all three `setDeposit`, `setAttachedGas`, `setTotalFee` (btw, not only it sets the total fee...) into `collect*` counterparts, which would accept `actions` as their function parameters, so make them pure functions, and create a separate method, which will call all of the functions and `setState` only once.

Also, we should trigger the calculation process (the proposed method ^) not only from `componentDidMount`, but also from `componentDidUpdate`.

icerove:
will fix the test error and then merge

frol:
@icerove I wonder if we can turn off `componentDidMount` triggers when we render with jest as it seems that it does not take into account anything beyond initial render anyway.

icerove:
will make a new PR to figure out that

PR #215, opened 2020-03-03
Fix storage display

icerove:
fix the bug for 'from genesis'


frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -119,9 +120,13 @@ export default class extends React.Component<Props, State> {
             <CardCell
               title="Creation Hash"
               text={
-                <TransactionLink transactionHash={account.address}>
-                  {account.address}
-                </TransactionLink>
+                account.address === message ? (
```
frol: We should just make that string empty wherever we create a "dummy" account object, and check if it is an empty string here and then display a human-readable message here.

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -44,6 +44,7 @@ export default class extends React.Component<Props, State> {
   render() {
     const { account } = this.props;
     const { storagePaidAtBlockHash } = this.state;
+    const message = "from genesis" || "loading";
```
frol: This does not make any sense to me... Well, given ^ we should just remove this. I still wonder what this expression was meant to be.

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -119,9 +119,13 @@ export default class extends React.Component<Props, State> {
             <CardCell
               title="Creation Hash"
               text={
-                <TransactionLink transactionHash={account.address}>
-                  {account.address}
-                </TransactionLink>
+                account.address === "" ? (
+                  ""
```
frol: Should we display "from genesis" here?

icerove: maybe not, let's wait to show real information about genesis when we finish fetching from genesis?

frol: Ok. Fair enough.

PR #221, opened 2020-03-12
Add node info

icerove:
add more information to node


frol:
Since we already have devnet and betanet, we should finish this PR (just make sure it still shows something useful for the testnet)

There is still an issue of misreporting (https://github.com/nearprotocol/nearcore/issues/2306) the telemetry data to a wrong network, so we should also add a condition checking the chain ID to match the one that is relevant for the alias (domain name).

icerove:
I see the link you put, but still not know what to do

frol:
I thought this PR was adding more info to the Nodes List page, that is why I thought it is a draft. We should just land it then.

PR #222, opened 2020-03-13
standardize license to be dual license MIT and Apache

mikedotexe:
Discussions have landed that while `nearcore` will have the GNU license, many other repositories will have a dual license of MIT and Apache. This is one of those repositories.


frol:
README file should also be updated

mikedotexe:
> README file should also be updated

Thanks, Frol. On a roll making similar PRs across many repos, I will get to this tomorrow though. Good eye

PR #224, opened 2020-03-18
Actions table

### LINKED ISSUE #217: Move actions into a separate SQL table
Move Actions into their own table from the Transactions table

icerove:
fix #217

### DIFF
--- a/backend/src/sync.js
+++ b/backend/src/sync.js
@@ -52,17 +52,6 @@ async function saveBlocks(blocksInfo) {
               return Promise.all([
                 models.Transaction.bulkCreate(
                   blockInfo.transactions.map(tx => {
-                    const actions = tx.actions.map(action => {
-                      if (typeof action === "string") {
-                        return { [action]: {} };
-                      }
-                      if (action.DeployContract !== undefined) {
-                        delete action.DeployContract.code;
-                      } else if (action.FunctionCall !== undefined) {
-                        delete action.FunctionCall.args;
-                      }
-                      return action;
-                    });
                     return {
                       hash: tx.hash,
                       nonce: tx.nonce,
@@ -70,8 +59,7 @@ async function saveBlocks(blocksInfo) {
                       signerId: tx.signer_id,
                       signerPublicKey: tx.signer_public_key || tx.public_key,
                       signature: tx.signature,
-                      receiverId: tx.receiver_id,
-                      actions
+                      receiverId: tx.receiver_id
                     };
                   })
                 ),
--- a/backend/src/sync.js
+++ b/backend/src/sync.js
@@ -50,25 +50,26 @@ async function saveBlocks(blocksInfo) {
             .map(blockInfo => {
               const timestamp = parseInt(blockInfo.header.timestamp / 1000000);
               return Promise.all([
-                blockInfo.transactions.map(tx => {
-                  models.Transaction.bulkCreate(
-                    tx.map(() => {
-                      return {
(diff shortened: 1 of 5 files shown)


backend/src/sync.js
```diff
@@ -52,29 +52,44 @@ async function saveBlocks(blocksInfo) {
...
                       signerId: tx.signer_id,
                       signerPublicKey: tx.signer_public_key || tx.public_key,
                       signature: tx.signature,
-                      receiverId: tx.receiver_id,
-                      actions
+                      receiverId: tx.receiver_id
                     };
                   })
                 ),
+                blockInfo.transactions.map(tx => {
```
frol: JS fails you here. `Promise.all()` accepts a list of promises, and so we have a list of:

* one promise created with `models.Transaction.bulkCreate`
* a list of promises created for each `transaction` in `blockInfo.transactions` with `models.Action.bulkCreate`
* one promise created with `models.Account.bulkCreate`

`Promise.all` will consider the list of `models.Action.bulkCreate` promises as an already resolved value, and will not await for it.

Concatenate the list of `models.Transaction.bulkCreate` and `models.Account.bulkCreate` promises with the list of promises created with the `blockInfo.transactions.map`.

Example:

```js
async function qq1() {
  const x = Promise.all(
    [
      [new Promise((resolve) => setTimeout(() => { console.log('QQ1: resolved'); resolve() }, 1000))]
    ]
  );
  await x;
  console.log('QQ1: all the promises are resolved');
};

async function qq2() {
  const x = Promise.all(
    [
      new Promise((resolve) => setTimeout(() => { console.log('QQ2: resolved'); resolve() }, 1000))
    ]
  );
  await x;
  console.log('QQ2: all the promises are resolved');
};
```

When you run `qq1`, you will get:

```
QQ1: all the promises are resolved
QQ1: resolved
```

That is not what you want.

When you run `qq2`, you will get:

```
QQ2: resolved
QQ2: all the promises are resolved
```

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -172,16 +172,22 @@ export default class TransactionsApi extends ExplorerApi {
           //
           // Once the above TODO is resolved, we should just move this to TransactionInfo method
           // (i.e. query the information there only for the specific transaction).
-          const actions = transactionExtraInfo.transaction.actions;
-
-          transaction.actions = actions.map((action: RpcAction | string) => {
+          const actions = await this.call<any>("select", [
```
frol: Well, given the above comments, we cannot do this yet :disappointed: 

You may notice that we store Actions in the DB, but never use them, and you are absolutely right. this is the case now.

Let's keep this code as a comment here, so in future we can simply uncomment it.

icerove: okaaaaay

backend/src/sync.js
frol: @icerove Please, familiarize yourself with Promises and async/await in JS. Here is what you essentialy have here now:

```
Promise.all(
  [
     [undefined, undefined, ...],
     Promise-to-bulk-create-Accounts
  ]
)
```

Well, we should also refactor this code (break it apart).

frontend/src/libraries/explorer-wamp/transactions.ts
frol: Well, Wallet does use actions available from the Explorer Backend, so we should also have a PR for Wallet.

PR #225, opened 2020-03-20
first version of infinite scroll

### LINKED ISSUE #61: TX List: Search and pagination

icerove:
fix #61 and #173


frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -1,8 +1,14 @@
+import InfiniteScroll from "react-infinite-scroll-component";
+
 import React from "react";
 
+import PaginationSpinner from "./PaginationSpinner";
+
 export default (
   WrappedComponent: React.ComponentType<any>,
   fetchDataFn: Function,
+  count: number,
```
frol: `count` is a too generic name

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -1,8 +1,14 @@
...
+  scroll: boolean,
```
frol: `scroll` is not a descriptive name. It is unclear what it refers to

frol: The API of this function leads to poor readability on the call site:

```
autoRefreshHandler(MyComponent, () => {}, 10, false, props);
```

Let's either expect a single argument (an object with all the parameters, like `props` to React components) or reduce the number of arguments to 3: component, config (an object which will contain fetchDataFn, "count" and "scroll"), props.

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -74,15 +74,20 @@ export default class AccountsApi extends ExplorerApi {
...
         `SELECT account_id as id, timestamp, transaction_hash as address 
-        FROM accounts 
+        FROM accounts
+        WHERE timestamp < :endTimestamp
         ORDER BY timestamp DESC
         Limit :limit`,
         {
-          limit
+          limit,
+          endTimestamp: endTimestamp === -1 ? "MAX(timestamp)" : endTimestamp
```
frol: Well, you, essentially, create a more complex query by requesting MAX value and filtering for it later for nothing. Also, you exclude the last/MAX value. Thus, if there is no `endTimestamp`, there is no need to have the WHERE clause.

frontend/src/components/utils/Timer.jsx
```diff
@@ -19,10 +19,12 @@ export default class extends React.PureComponent {
     clearInterval(this.timer);
   }
 
+  newDate = new Date();
+
   componentDidUpdate() {
-    const time = this.props.time === undefined ? new Date() : this.props.time;
+    const time = this.props.time === undefined ? this.newDate : this.props.time;
```
frol: Hmm, I wonder where I get this original implementation from... it does not make any sense to me now to have the time since the first rendering... Let's refactor this in a separate PR forbidding such a use (in terms of TypeScript that would mean that we only allow `props.time` to be of `Date` type, and never undefined).

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -74,15 +74,20 @@ export default class AccountsApi extends ExplorerApi {
...
   }
 
-  async getAccounts(limit: number = 15): Promise<AccountBasicInfo[]> {
+  async getAccounts(
+    limit: number = 15,
+    endTimestamp?: number
+  ): Promise<AccountBasicInfo[]> {
     try {
       return await this.call("select", [
         `SELECT account_id as id, timestamp, transaction_hash as address
```
frol: Rebase from master

frontend/src/libraries/explorer-wamp/nodes.ts
```diff
@@ -17,18 +17,20 @@ export interface NodeInfo {
 }
 
 export default class NodesApi extends ExplorerApi {
-  async getNodes(limit: number = 15) {
+  async getNodes(limit: number = 15, endTimestamp?: number) {
     try {
       return await this.call<NodeInfo[]>("select", [
         `SELECT ip_address as ipAddress, moniker, account_id as accountId, node_id as nodeId, signature, 
-                last_seen as lastSeen, last_height as lastHeight, last_hash as lastHash,
+                last_seen as timestamp, last_height as lastHeight, last_hash as lastHash,
```
frol: I suggest we don't break our APIs throughout the project just for the sake of `autoRefreshHandler`. Let's add a way to specify a "getter" function for `autoRefreshHandler`, which will extract the `timestamp` / `blockTimestamp` / `lastSeen` / or whatever it is from a given element. The same goes for `endTimestamp`, as from the semantics point of view, it does not make sense for `NodesApi.getNodes` to have `endTimestamp` argument.

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -31,29 +45,92 @@ export default (
...
+          break;
+        case "Node":
+          endTimestamp = this.state.items[this.state.items.length - 1].lastSeen;
+          break;
+        case "Transaction":
+          endTimestamp = this.state.items[this.state.items.length - 1]
+            .blockTimestamp;
+          break;
+        default:
+          endTimestamp = undefined;
```
frol: There are at least three options to put the responsibility for this logic to:

1. (this way) put it into the downline component and teach it to handle every single possible type of data. The problem is that you have to hard-code every single usage scenario (it is not scalable and fragile).
2. (previous implementation) enforce the external contract and require the data to be of a specific shape (having "timestamp" field in every item). This works, but you have to repack the data and in the previous attempt you had to go throughout the codebase and rename the field that makes sense for the specific type into a generic "timestamp". As you can see, the change is intrusive to the whole codebase, and it is also fragile and creates a tight coupling from the API through the relevant components down to the autoRefreshHandler.
3. (what I suggested) design authRefreshHandler to receive a function (similarly to `fetchDataFn`) which will receive an item and return a range indicator ("timestamp"), this way autoRefreshHandler won't need to know anything about the structure of the `item`, and there is no need to change the APIs fetching, so all you need is to pass this "timestamp extractor" function to autoRefreshHandler.

frontend/src/libraries/explorer-wamp/nodes.ts
```diff
@@ -17,18 +17,20 @@ export interface NodeInfo {
 }
 
 export default class NodesApi extends ExplorerApi {
-  async getNodes(limit: number = 15) {
+  async getNodes(limit: number = 15, endTimestamp?: number) {
```
frol: I still encourage you to think about the argument naming from the usage perspective. `endTimestamp` does not have a clear meaning when you query a list of nodes.

frol:
It seems that this PR introduced horizontal scrollings on the dashboard:

image

Can we resolve those?

icerove:
> It seems that this PR introduced horizontal scrollings on the dashboard:
> 
> image
> 
> Can we resolve those?

I am not very sure why this happens and I will check first

PR #226, opened 2020-03-20
update url

icerove:
update NEAR_RPC_URL with devnet and betanet


icerove:
update NEAR_RPC_URL with devnet and betanet

# Test Plan
no test add
all test passed

PR #227, opened 2020-03-24
Add contract detail

### LINKED ISSUE #92: Add contract info to Account Detail page
## Story

As a user, I want to view the details of a contract deployed on an account, so I learn about the contract's behavior and determine whether or not to trust the contract.

## Acceptance Criteria

- [ ] Indicator if a contract is deployed to an account
- [ ] If contract, the following is displayed
  - [ ] Last updated timestamp (when the contract was last redeployed), links to TX detail page of the most recent deployContract TX
  - [ ] Locked status (False, if the contract has any FullAccessKeys)
- [ ] Clicking on the creator's name/address redirects to the account detail page
- [ ] Mobile ok
- [ ] Tested

## Assets

- [Designs](https://www.figma.com/file/jS1bjIiVxoEPNOYeXwddjH/NEAR-Explorer?node-id=42%3A1694)

## Notes
**Locked status**
- If account has any FullAccessKey, status is `False`
- If account has no FullAccessKey, status is `True`

icerove:
fix #92


package.json
```diff
@@ -3,7 +3,7 @@
   "devDependencies": {
     "husky": "^2.3.0",
     "lint-staged": "^8.1.7",
-    "prettier": "1.17.1"
+    "prettier": "^2.0.1"
```
frol: Please, revert this change and do as we discussed in chat:

1. We merge all (most) the existing PRs
2. Create a separate PR where only prettier changes will occur

This PR is impossible to review at this state.

backend/models/Account.js
```diff
@@ -10,7 +10,7 @@ module.exports = (sequelize, DataTypes) => {
       },
       transactionHash: {
         type: DataTypes.STRING, // base58
-        allowNull: true
+        allowNull: false
```
frol: This is fixed in https://github.com/nearprotocol/near-explorer/pull/230

backend/models/Contract.js
```diff
@@ -0,0 +1,30 @@
+"use strict";
+module.exports = (sequelize, DataTypes) => {
+  const Contract = sequelize.define(
```
frol: Well, having a copy of a table (same as Account) is suboptimal.

Let's discuss if we have leave without it.

To know if there is any contract deployed to the account_id you can query RPC:

```
$ http post https://rpc.nearprotocol.com jsonrpc=2.0 id=dontcare method=query 'params:=["account/metanear-chat",""]'
```

and if the `"code_hash"` is not `"11111111111111111111111111111111"`, then there is a contract, so we can query that information from the frontend on demand. This code hash is not the transaction hash where it was deployed, but the hash of the binary file.

To know the time and the transaction hash of the latest update, you query Actions for `ContractDeployed` action_type joining with the transactions table to filter by `receiver_id`.

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,155 @@
...
+interface State {
+  display: boolean;
+  locked: boolean;
+  contractInfo: C.ContractInfo;
+}
+
+export default class extends React.Component<Props, State> {
+  state: State = {
+    display: false,
+    locked: true,
```
frol: It is not a safe default. There is no good default, so it is better to start it as undefined (use question mark on the field definition in the State interface).

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,155 @@
...
+      .getContractInfo(this.props.accountId)
+      .then(contractInfo => {
+        if (contractInfo.transactionHash !== null) {
+          contractInfo.accessKeys.map((key: any) => {
+            const permission = key["access_key"]["permission"];
+            if (permission === "FullAccess") {
+              this.setState({ locked: false });
+            }
+            return;
+          });
```
frol: 1. Given it is safer to fail and say that the contract is not locked than the other way around, we should check that there are no other access keys but `FunctionCall`. This way if nearcore will introduce another permission type in the future, that could mean "unlocked" status (e.g. `RedeployAccess`), we won't serve people false trust that the contract is locked.
2. This can be nicely expressed with a built-in JS method `.every`:

    ```js
    locked: contractInfo.accessKeys.every((key) => key["access_key"]["permissions"]["FunctionCall"] !== undefined)
    ```

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,155 @@
...
+  display: boolean;
+  locked: boolean;
+  contractInfo: C.ContractInfo;
+}
+
+export default class extends React.Component<Props, State> {
+  state: State = {
+    display: false,
+    locked: true,
+    contractInfo: {
```
frol: I guess that if you allow `contractInfo` to be `undefined` (which makes a lot more sense then this blank object), you won't need `display` flag as well since you will be able to check if there is anything in `contactInfo` in the `render`.

backend/models/Contract.js
bowenwang1996: why do we need to distinguish between accounts and contracts? The distinction is rather superficial and we can choose whether to display anything about contract based on code hash.

frol: @kcole16 Do we have someone's opinion/vision on the separation of Accounts and Contracts into separate pages? Would we consider having this as one entity on the Explorer?

Well, just from the user perception, one may have a mental model that there are accounts (users ~= humans) and contracts (programs), and they are not the same, but we, probably, should not encourage this mental model...

kcole16: @bowenwang1996 We aren't distinguishing between accounts and contracts currently. We're just indicating whether or not there is a contract deployed to an account.

@frol Agree we shouldn't encourage a mental model of separation. For now, I don't think the list of contracts is that useful, though when we have contracts matching certain standards (fungible/non-fungible tokens, etc), that might make more sense

frol: @icerove So, to sum it up, we don't need separate pages for contracts list and contract details.

@kcole16 I think we should replace the top bar "Contracts" with "Accounts". Agree?

kcole16: Yes, the contract details table should be an additional section on the Account Detail page.

icerove: @frol  We just refactor most of the SQL query without joining, should we do that here?
>To know the time and the transaction hash of the latest update, you query Actions for ContractDeployed action_type joining with the transactions table to filter by receiver_id.

frontend/src/components/contracts/ContractDetails.tsx
icerove: I think you misunderstanding the keys cintent. it is like this:
image
other than "FullAccess", it is FunctionCall object, not every permission is functioncall. and if there is one permission with  "FullAccess", it is unlocked.

kcole16: @icerove What @frol is saying is correct.

Though @frol, isn't it easier to check if any of the keys are `FullAccess` (which would make the contract unlocked), rather than if the only keys present are `FunctionCall`?

frol: @icerove The problem I am describing here is the future-proofing. Currently, there is no difference if you check if there is any FullAccess or there is no other permission but FunctionCall, these two are equivalent right now, and we use it to inform our users about the locked state of the contract. Adding a new permission type, however, can lead to the following consequences:

1. We display an unlocked contract as locked -- that is really bad, people will be assuming false security.
2. We display a locked contract as not locked -- that is not great, but it is just a bug.

It is better to false alarm people than to lead them into false feelings of security.

Example, nearcore adds a new permission "CanRedeployCode" (having this permission means that the contract is not locked even if there is no FullAccess permission). If the account has only this new permission, then with the current check we will falsely make users believe that the contract is locked, but my proposed change will just work and indicate the contract as "not locked".

Thus, the check for "no other permissions but FunctionCall" is better than "there are no FullAccess".

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+interface State {
+  locked?: boolean;
+  transactionHash?: string;
+  timestamp?: number;
+}
+
+export default class extends React.Component<Props, State> {
+  state: State = {
+    locked: true
```
frol: This is not a safe default. There is no good default here, so there should be none.

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,127 @@
...
+  getContractInfo = async () => {
+    new ContractsApi()
+      .getContractInfo(this.props.accountId)
+      .then(contractInfo => {
+        if (contractInfo) {
+          this.setState({
+            transactionHash: contractInfo.transactionHash,
+            timestamp: contractInfo.timestamp,
+            locked: contractInfo.accessKeys.some(
+              (key: any) => key["access_key"]["permissions"] === "FullAccess"
```
frol: 1. This is an inverted logic here. The "locked" contact is one that nobody can ever change, so there is no other permission but FunctionCall.
2. Read my comments about the difference "no FullAccess" vs "no permissions except FunctionCalls" above.

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,127 @@
...
+              <CardCell
+                title="Last Updated"
+                text={moment(timestamp).format("MMMM DD, YYYY [at] h:mm:ssa")}
+                className="block-card-created-text account-card-back border-0"
+              />
+            </Col>
+            <Col md="2">
+              <CardCell
+                title="Status"
+                text={locked ? "Locked" : "Unlocked"}
```
frol: Given there is a third state, when we don't know yet, it is better to display nothing at all unless we know the state.

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,127 @@
...
+  transactionHash?: string;
+  timestamp?: number;
+}
+
+export default class extends React.Component<Props, State> {
+  state: State = {
+    locked: true
+  };
+
+  getContractInfo = async () => {
```
frol: Non-descriptive function name: this function does not get a contract info to the caller.

frontend/src/components/utils/Header.jsx
```diff
@@ -32,9 +32,9 @@ export default () => (
         text="Dashboard"
       />
       <HeaderNavItem
-        link="/contracts"
+        link="/accounts"
```
frol: :+1:

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,130 @@
...
+  componentDidUpdate(preProps: Props) {
+    if (this.props.accountId !== preProps.accountId) {
+      this.collectContractInfo();
+    }
+  }
+  render() {
+    const { locked, transactionHash, timestamp } = this.state;
+    let lockedShow;
+    if (locked !== undefined) {
+      lockedShow = locked === true ? "Locked" : "UnLocked";
```
frol, line 57: Perfect defensive programming!

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,130 @@
...
+          CONTRACT
+        </div>
+        <div className="contract-info-container">
+          <Row noGutters className="border-0">
+            <Col md="6">
+              <CardCell
+                title="Code Hash"
+                text={
+                  <TransactionLink transactionHash={transactionHash}>
+                    {transactionHash}
```
frol, line 75: Code hash is the checksum of the Wasm binary blob, that is returned with the `query(account/<account_id>)` API.

The transaction hash when the code was last updated should be a separate thing if needed /cc @kcole16

icerove: if that, wait for new design

kcole16:
Not a fan of the "Last Updated" TX hash below the timestamp, but I'm thinking we should make this a separate issue of handling timestamps and creation hashes in general (as they should be consistent).

Thoughts @icerove @frol ?

Screen Shot 2020-04-02 at 12 22 08 PM

icerove:
@kcole16 We can fix this in this PR and make another one to make all in uniform. could you give new design here since it will change? or describe more to give details

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,138 @@
...
+    let lockedShow;
+    if (locked !== undefined) {
+      lockedShow = locked === true ? "Yes" : "No";
+    }
+    let updated;
+    if (timestamp && transactionHash) {
+      const time = moment(timestamp).format("MMMM DD, YYYY [at] h:mm:ssa");
+      updated = (
+        <>
+          {time} [<TransactionLink transactionHash={transactionHash} />]{" "}
```
frol, line 66: It seems that `{" "}` is not needed here.

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,138 @@
...
+interface Props {
+  accountId: string;
+}
+
+interface State {
+  locked?: boolean;
+  transactionHash?: string;
+  timestamp?: number;
+  codeHash?: string;
+}
```
frol, line 21: BTW, it is a good component design :+1:

frontend/src/pages/accounts/[id].jsx
```diff
@@ -30,10 +31,13 @@ export default class extends React.Component {
...
-          createdAtBlockTimestamp: basic.createdAtBlockTimestamp,
-          createdByTransactionHash: basic.createdByTransactionHash
-        });
+        if (basic) {
+          this.setState({
+            createdAtBlockTimestamp: basic.createdAtBlockTimestamp,
+            createdByTransactionHash: basic.createdByTransactionHash
+          });
+        }
+        return;
```
Comment (frol) [inline frontend/src/pages/accounts/[id].jsx:40]: This `return` does not seem to be useful.

kcole16:
I think we should drop the TX hash, and make the timestamp a link to it. This is inconsistent with the timestamp + hash of the account itself, but I think we should consider consolidating that as well. The actual TX hash doesn't add much value in this situation, especially not a snippet of it.

As for the code hash, maybe truncate it (with ellipses) and display the full hash on hover? 

Screen Shot 2020-04-02 at 4 45 45 PM

frol:
I am ok with the timestamp-link. As to the code hash, it just occurred to me that most of the tools computing checksums use hex (base16) encoding to display the value, so our base58 is a bit meh... Hover is not great on mobile and not great in this particular case in UX perspective (users, when they want to have a full hash, don't want to read it, they want to copy it, right?)

frol:
A few variants:

image

Aligned with the "creation hash":

image

frol:
@corwinharrell Can you help us with this? :smiley:

kcole16:
Fair point re: wanting to copy the hash. 

We could also play with the grid a bit (doesn't help much on mobile).

Screen Shot 2020-04-02 at 5 20 36 PM

I think generally we need to resize the fonts, as they are too often too large for the amount of information we're displaying (this is true on wallet as well, in many places). Though this is definitely a larger item (part of UX papercuts). cc @corwinharrell @jakestutzman

icerove:
@frol @kcole16 so we are waiting for design? cc @corwinharrell

corwinharrell:
Hey guys, just dropping in here to remind that I'm OOO and won't be able to assist on this right now but don't hesitate to ping @jakestutzman if any small changes to design need to be made.

jakestutzman:
Here's what I'm thinking. It puts all of the hashs in a row so their easy to grab if needed. It also highlights the new information in the white spaces, and static info in the grey areas. I also cleaned up the 'contract' label because it was too distracting and took up too much vertical space that it was crowding the data.

Let me know if I missed anything

image

bowenwang1996:
@jakestutzman why do we annotate code hash with `sha-256 base58` specifically? It is the same type as transaction hash.

jakestutzman:
@bowenwang1996 was actually following a suggested annotation by @frol

frol:
@bowenwang1996 I would argue that we should switch to hex (base16) for the "code hash" since this is what the de facto standard for checksums, until there is any discussion around this, I prefer to hint the user hat they are looking at.

frol:
@jakestutzman it is so much better than all the crap I generated!

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  render() {
+    const { locked, transactionHash, timestamp, codeHash } = this.state;
+    let lockedShow;
+    if (locked !== undefined) {
+      lockedShow = locked === true ? "Yes" : "No";
+    }
+    console.log(codeHash, lockedShow);
```
frol, line 62: Let's remove debug outputs

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -0,0 +1,147 @@
...
+    return transactionHash ? (
```
frol, line 63: It is better to just return earlier as it is really hard to parse these long conditions.

frontend/src/libraries/explorer-wamp/contracts.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export default class ContractsApi extends ExplorerApi {
+  async getContractInfo(id: string) {
+    try {
+      const codeHash = await this.queryCodeHash(id);
+      if (codeHash !== "11111111111111111111111111111111") {
+        const [contractInfo, accessKeys] = await Promise.all([
+          this.call<any>("select", [
+            `SELECT transactions.block_timestamp, transactions.hash 
+            from transactions
```
frol, line 18: I noticed that you use `from` in the lowercase, let's use `FROM` to match the style of the rest of the SQL query in the future.

frontend/src/components/contracts/ContractDetails.tsx
icerove: what do you mean? sorry I do not get it

PR #229, opened 2020-03-24
fix(frontend): Fixed React warning about textarea usage

frol:
> Warning: Use the `defaultValue` or `value` props instead of setting children on <textarea>.


frol:
I had Prettier 2.0 locally, so this is what formatted the code on git hook. We will need to update our prettier once we have all the major PRs merged (I push these small PRs today to clear my way pushing an update to the DB models)

icerove:
make sense. I also have prettier 2.0 locally and in PR contract details

frol:
The test pass now since I formatted the code with the Prettier 1.17.1 which matches the version in the package.json.

@icerove Please, review

Issue #220, opened 2020-03-12
Improve performance


kcole16:
### Overview
Explorer performance is suffering at somewhere over >20k transactions, and is unusable at over >100k. 

We've identified the following steps to improve query performance, especially for transaction queries.

* [x] Extract actions into a separate table (from transactions)
* [x] Test loads, profile query performance
    * [x] Goal: 1M TXs
* [x] Set expectations based on load test results

frol:
Currently, we have synced 1.9M blocks, 15M chunks, and 85k transactions for the testnet (the DB size is only 4GB), and Explorer feels snappy! It seems that we resolved the major bottlenecks with #224 and #231. /cc @icerove 

We still want to know our limits (#219), though.

frol:
There is a sum-up: https://github.com/near/near-explorer/issues/219#issuecomment-613641637

We don't have any outstanding performance issues and so far it seems that our read performance is great and we can handle 10M transactions without breaking a sweat.

PR #231, opened 2020-03-25
feature: Improve performance by denormalizing and adding indecies to transactions DB table

frol:
*Related to #220*

## Test Plan

* [x] Existing unit-tests pass
* [x] Manually confirmed that the Dashboard, Transactions list and details, Blocks list and details, and Account list and details pages work as expected

### DIFF
--- a/frontend/src/components/transactions/TransactionDetails.tsx
+++ b/frontend/src/components/transactions/TransactionDetails.tsx
@@ -82,7 +82,7 @@ export default class extends React.Component<Props, State> {
     if (gasAttached === null) {
       gasAttached = gasUsed;
     }
-    const stateUpdate: any = { deposit, gasUsed, gasAttached };
+    const stateUpdate: State = { deposit, gasUsed, gasAttached };
     if (this.state.block) {
       stateUpdate.transactionFee = gasUsed.mul(
         new BN(this.state.block.gasPrice)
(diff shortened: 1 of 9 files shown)


frontend/src/components/transactions/TransactionDetails.tsx
```diff
@@ -71,31 +68,62 @@ export default class extends React.Component<Props, State> {
...
+      stateUpdate.transactionFee = gasUsed.mul(
+        new BN(this.state.block.gasPrice)
+      );
+    }
+    this.setState(stateUpdate);
   };
 
   componentDidMount() {
-    this.updateComputedValues();
+    this.updateBlock();
```
icerove: is this should also have this.updateComputedValues()?

frontend/src/components/transactions/TransactionDetails.tsx
```diff
@@ -71,31 +68,62 @@ export default class extends React.Component<Props, State> {
...
-    const gasPrice = new BN(this.props.transaction.gasPrice);
-    const transactionFee = gasUsed.mul(gasPrice);
     let gasAttached = this.collectGasAttached(this.props.transaction.actions);
     if (gasAttached === null) {
       gasAttached = gasUsed;
     }
-    this.setState({ deposit, transactionFee, gasUsed, gasAttached });
+    const stateUpdate: any = { deposit, gasUsed, gasAttached };
+    if (this.state.block) {
+      stateUpdate.transactionFee = gasUsed.mul(
```
icerove: I am not very sure about this: stateUpdate not have transactionFee included, is this legal ?

frontend/src/libraries/explorer-wamp/details.js
```diff
@@ -24,7 +24,7 @@ export default class DetailsApi extends ExplorerApi {
               WHERE blocks.timestamp > (strftime('%s','now') - 10) * 1000
           ) as transactions_per_second,
           (SELECT height as lastBlockHeight FROM blocks ORDER BY height DESC LIMIT 1) as last_block,
-          (SELECT COUNT(DISTINCT signer_id) as accountsCount FROM transactions) as total_accounts`
+          (SELECT COUNT(*) as accountsCount FROM accounts) as total_accounts`
```
icerove: this may not inclueded genesis aaccount, we should sum all accounts. Maybe we can left this here, but after getting genesis information, we should fix it

frol: Well, counting DISTINCT for all the transactions is computationally heavy, and still does not provide a good estimate on the total count of accounts, and on the accounts page we only display the accounts we have in the table anyway, so it is just more consistent. We should, however, sync genesis info... https://github.com/nearprotocol/near-explorer/issues/185

PR #234, opened 2020-03-28
feature: Avoid using JOINs to get stats for the Dashboard

frol:
## Test Plan

* [x] Manually confirm the queries match the previous numbers


frol:
@icerove I agree that we should keep migrating to TS; unfortunately, I don't have capacity to do that myself these days.

PR #248, opened 2020-04-08
change to click button to load more

### LINKED ISSUE #240: The lists does not load the next page if all the page fit on the screen
## Story

A user cannot get the next "page" of results if the initial data fit on the screen and thus the condition of getting to the "end" of the page does not get triggered.

It seems to be related to https://github.com/ankeetmaini/react-infinite-scroll-component/issues/106

## How to reproduce?

1. Reduce the number of items prefetched by the infinite scroll component (e.g. list of blocks) to 2
2. Open the page
3. You will see the first 2 blocks, but never get the next blocks even though the "Loading" text is there.

## Acceptance Criteria

- [ ] All the data can be viewed
- [ ] **Mobile ok**
- [ ] Tested

icerove:
fix #240


frol:
This button can just solve the problem with infinite scroll! Let's combine this button with the original infinite scrolling!

icerove:
if we use button to load more, why do you still want infinite scroll? @frol

frol:
I would say that we want infinite scroll by default and that button to be just a quick fix to the problem we have with the infinite scroll

icerove:
Let's assume the situation:
when user see page, they see the button "load more" and by default as other website, it is click it and fetch more data that want. If we also have infinite scroll here, when they scroll to the bottom and want to click the button, weird thing happens here, it fetch more data without click the button. It feels more like a bug instead of combine auto fetch and click button fetch. what do you think? @frol

frol:
If we replace the button with a loading animation on the infinite scrolling event, the user may not even see the button is there or see it gets transformed into the loading animation, which is fine.

icerove:
if we replace the button with a loading animation, the user would not click it for sure. But it also does not fix the bug #240. it will not scroll, either.

frol:
I mean the following:

1. Use infinite scrolling
2. Display a button for a manual load just in case the infinite scrolling fails
3. When infinite scrolling event is fired, hide the button and display a loading animation -- this will way the user might not even see the button when the infinite scrolling works fast or might see the button immediately replated with the loading animation -- both cases are intuitive

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -119,18 +121,43 @@ export default (
...
+        <>
+          <InfiniteScroll
+            dataLength={this.state.items.length}
+            next={this.fetchMoreData}
+            hasMore={this.state.hasMore}
+            loader={<PaginationSpinner hidden={false} />}
+            style={{ overflowX: "hidden" }}
+          >
+            <WrappedComponent items={this.state.items} {...props} />
+            {this.state.hasMore && (
```
frol, line 137: 1. Hide the button when the data is loading
2. Let's re-use the "VIEW ALL" button design from the dashboard here and put it in the middle

frol:
image

1. This looks quite different from 
image

2. There should be either the loading indicator or the button, not both at the same time. @icerove Please, please, please, test your changes for UX and completeness.

icerove:
@frol I used to try to use react-infinite-scroll one, but it does not support typescript one.

frol:
@icerove There are two ways dealing with missing TS support:

1. try finding `@types/<package-name>` on npm (`@types/react-infinite-scroller`) and drop it into devDependencies
2. whitelist it in `frontend/next-env.d.ts`

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -49,48 +50,64 @@ export default (
...
     fetchInfo = (count: number) => {
       config
         .fetchDataFn(count)
         .then((items: any) => {
           if (items.length > 0) {
-            this.setState({ items, display: true });
+            this.setState({ items, itemsLength: items.length, display: true });
+            if (items.length < config.count) {
+              this.setState({ hasMore: false });
+            }
```
frol, line 79: It is still possible that we will trigger 2 renders instead of 1 here. Just as a pattern, collect the state changes into an object and the apply that once:

```js
let newState;
if ... {
   newState = ...
   if ... {
       newState.hasMore = false;
   }
} else { ... }
this.setState(newState)
```

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -49,48 +50,64 @@ export default (
...
-      }
-    };
-
     fetchMoreData = async () => {
-      if (config.dashboard) {
-        this.setState({ hasMore: false });
-        return;
-      }
-      if (this.state.items.length > 0) {
+      this.setState({ loading: true });
```
frol, line 90: This may set the loading = true, and then never go into the `if` statement, and thus never call the fetching data method, and ever resolve, so loading will never get to false.

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -49,48 +50,64 @@ export default (
...
+      if (this.state.itemsLength > 0) {
```
frol, line 91: I would use early-return instead of nesting here.

frol:
@icerove I don't see a loading indicator at all. Did you kill it completely? We need the indicator, and need the button, but the button should **not** be displayed when the data is loading, and the indicator should only be visible when the data is loading.

icerove:
for this library we use, when hasMore indicator is true, the loading indicator(whether it is "Loading" or animation), will always show there and thus you will always see the loading animation unless hasMore become false(when till the end). if I do not kill it, it will always like this way.

frol:
Given you already added the `loading` flag into the state, you can pass different components based on the flag into the `loader` props.

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -69,17 +69,29 @@ export default (
     };
 
     fetchInfo = (count: number) => {
+      let newState = this.state;
```
frol, line 72: You are fortunate enough here that you re-assigned this newState later... You should never modify `self.state` as this will hit you back in a very unexpected way. You don't need the old state for `setState`. I will update the code.

PR #250, opened 2020-04-09
Calculate tps

### LINKED ISSUE #246: TPS metric is not accurate
## Story

As a regular user who is interested in the network performance, I want to know the current transactions per second (TPS) load so I can have a clear picture.

Currently, the throughput is computed based on the real-time "now" moment, but it is not accurate since the blocks are not synced immediately. We should use the last block timestamp instead (use a nested query to get the latest block timestamp inline).

I think we should display "N/A" when the lastest synced block is older than 30 seconds.

## Acceptance Criteria

- [ ] TPS is connected to the synced blocks, not the world clock time
- [ ] TPS is "N/A" when the latest synced block is older than 30 seconds
- [ ] Tested

icerove:
fix #246


frontend/src/libraries/explorer-wamp/details.ts
```diff
@@ -0,0 +1,67 @@
...
+        transactionsPerSecond: Tps
+      };
+    } catch (error) {
+      console.error("Details.getDetails failed to fetch data due to:");
+      console.error(error);
+      throw error;
+    }
+  }
+
+  async getTPS() {
```
frol, line 43: I believe, these queries can be merged into the main query just fine.

If you still decide to keep TPS separately (though, i cannot think of a reason), you should at least merge these two queries, so they don’t get out of sync (waiting on a slow network can create a noticeable delay between the two calls and that will become transaction per 10+ seconds)

icerove: np! I separate it first try to see whether SQL query is separated correctly.

frontend/src/libraries/explorer-wamp/details.ts
```diff
@@ -0,0 +1,67 @@
...
+  lastBlockHeight: number;
+  transactionsPer10Second: number;
+  transactionsPerSecond: number | null;
+}
+
+export default class DetailsApi extends ExplorerApi {
+  async getDetails(): Promise<Details> {
+    try {
+      const blockTimestamp = await this.call("select", [
+        `SELECT timestamp FROM blocks WHERE timestamp > ((strftime('%s','now') - 30) * 1000 ) ORDER BY timestamp DESC LIMIT 1`
```
frol, line 17: Translate your code into simple English and find that you end up taking into account more transactions than you want.

1. Let's define NOW to be the current timestamp.
2. You query the block that is not older 30 seconds from now, and is the latest (great! but...)
3. ... you wait here for the response, let's say a minute, just to see how ridiculous the result will be
4. You make a new query with:

    ```
    SELECT COUNT(*) as transactionsPer10Second FROM transactions
              WHERE block_timestamp > (:blockTimestamp - 10 * 1000)  ) as transactions
    ```

    You count the number of transaction happened from the latest block a while ago (since that time we might have already synced new blocks) to the NEW NOW moment.

Add `AND block_timestamp <= :blockTimestamp`.

Please, don't make our database doing extra work when it can avoid doing it -- don't make it count with `strftime('%s','now') * 1000` if you are not going to (and should not) use it anyway.

frontend/src/libraries/explorer-wamp/details.ts
frol: Remove the dead function.

frontend/src/libraries/explorer-wamp/details.ts
```diff
@@ -0,0 +1,67 @@
+import { ExplorerApi } from ".";
+
+export interface Details {
+  accountsCount: number;
+  totalNodesCount: number;
+  onlineNodesCount: number;
+  lastDayTxCount: number;
+  lastBlockHeight: number;
+  transactionsPer10Second: number;
+  transactionsPerSecond: number | null;
```
frol, line 10: BTW, TS has `?` marker meaning that the value can be undefined:

```
transactionsPerSecond?: number
```

frontend/src/libraries/explorer-wamp/details.ts
```diff
@@ -0,0 +1,73 @@
...
+  transactionsPer10Second?: number;
+  transactionsPerSecond?: number;
+}
+
+export default class DetailsApi extends ExplorerApi {
+  async getDetails(): Promise<Details> {
+    try {
+      const blockTimestamp = await this.call("select", [
+        `SELECT timestamp FROM blocks WHERE timestamp > ((strftime('%s','now') - 30) * 1000 ) ORDER BY timestamp DESC LIMIT 1`
+      ]).then((it: any) => (it[0].timestamp !== null ? it[0].timestamp : null));
```
frol, line 18: Are you sure it is going to be `null`? I would expect not even a single element in the `it` array, and thus `undefined` value  accessing the 0th element.

frontend/src/libraries/explorer-wamp/details.ts
```diff
@@ -0,0 +1,73 @@
...
+          total_accounts.accountsCount,
+          last_block.lastBlockHeight,
+          Daytransactions.lastDayTxCount
+        FROM
+          (SELECT COUNT(*) as totalNodesCount FROM nodes) as nodes,
+          (SELECT COUNT(*) as onlineNodesCount FROM nodes WHERE last_seen > (strftime('%s','now') - 60) * 1000) as online_nodes,
+          (SELECT height as lastBlockHeight FROM blocks ORDER BY height DESC LIMIT 1) as last_block,
+          (SELECT COUNT(*) as accountsCount FROM accounts) as total_accounts,
+          (SELECT COUNT(*) as lastDayTxCount FROM transactions
+              WHERE block_timestamp > (strftime('%s','now') - 60 * 60 * 24) * 1000  ) as Daytransactions
```
frol, line 62: Now we have code duplication which may lead to inconsistencies if we forget to update both queries in the future. Redesign this to be sort of a query builder.

frontend/src/libraries/explorer-wamp/details.ts
```diff
@@ -0,0 +1,61 @@
...
+          `
+        ]).then((it: any) => it[0]),
+        blockTimestamp !== null
+          ? this.call("select", [
+              `
+          SELECT
+            transactions.transactionsPer10Second
+          FROM
+            (SELECT COUNT(*) as transactionsPer10Second FROM transactions
```
frol, line 42: There is no need to nest the query since it is a single one...

Issue #252, opened 2020-04-09
Incorrect TX is shown as the Creation TX


kcole16:
**Problem**
The incorrect TX is linked from the `test` account's [Detail page](https://explorer.nearprotocol.com/accounts/test) as the creation TX. It appears someone attempted to create an account named `test` after the original account, [this TX failed](https://explorer.nearprotocol.com/transactions/G9W8xFEvrzoFbtiLiVdHG2gZknHs7yP9oy5GA3MdhLkR) (since the name is taken), and this failing TX is now linked. 

**Expected Behavior**
The original, successful account creation TX should be linked from the Account Detail page. 

**Steps to reproduce**

1. Go to https://explorer.nearprotocol.com/accounts/test
2. Click on the Creation Hash

frol:
This bug nature is the same as #223, as we don't have transaction status at the moment when we sync chunks/transactions. We will get unblocked when we actually start saving transaction status with #218.

frol:
Resolved in Explorer baked by the Indexer for Explorer (see #236).

Issue #223, opened 2020-03-17
Account that is created in a failing transaction may show up on the list in Explorer


frol:
## Story

As a user who uses Explorer Accounts page, I want to know the correct information about the accounts.

It *seems* that currently, we may include an account from a CreateAccount action of a transaction that actually failed.

To resolve this we need to query individual transactions (since https://github.com/nearprotocol/nearcore/issues/1880 is not implemented yet) to know the status. While we are at it, we can also store this status in Explorer DB, so we don't need to query it from the node again and again.

## Acceptance Criteria

- [ ] There should be no "ghost" accounts

frol:
@bowenwang1996 

> In the case where the account is created through a transaction, you can check the final status of the transaction to see whether it succeeded or not. I understand that we still cannot track the accounts created through function calls, but I think we can at least make sure we don't display accounts that do not exist.

We cannot reason about the success of the whole transaction (#212 (comment), see (3)), and we don't track receipts here to match the execution outcome to it. /cc @khorolets FYI

bowenwang1996:
> We cannot reason about the success of the whole transaction

why? transaction status query allows you to check statuses of all receipts.

frol:
@bowenwang1996 While it does, the receipts are applied independently, so the receipt that creates an account may succeed while the transaction may still contain some receipts that fail. To know whether an account was created or not, we need to track the receipt and its execution outcome (which will only be the case once we land the Indexer for Explorer)

frol:
Resolved in Explorer baked by the Indexer for Explorer (see #236).

PR #253, opened 2020-04-10
build access key table

### LINKED ISSUE #162: API to query accounts by public key
For people who import keys into wallet need a way to find set of accounts this keys accessing to.
Usually it will be only one account, and this makes it easy to retrieve that account by public key.

Technical details:
  - need to add additional index in the storage to be able to retrieve this information
  - RPC query, account/<public key> -> returns list of <account_id>

icerove:
fix #162

### DIFF
--- a/backend/src/sync.js
+++ b/backend/src/sync.js
@@ -94,28 +94,35 @@ async function saveBlocks(blocksInfo) {
                 models.AccessKey.bulkCreate(
                   blockInfo.transactions
                     .filter(tx =>
-                      tx.actions.some(
-                        action =>
-                          action === "AddKey" || action.AddKey !== undefined
-                      )
+                      tx.actions.some(action => action.AddKey !== undefined)
                     )
                     .flatMap(tx => {
                       const accountId = tx.receiver_id;
-                      return tx.actions.map(action => {
-                        let accessKeyType;
-                        if (
-                          action.AddKey.access_key.permission === "FullAccess"
-                        ) {
-                          accessKeyType = "FullAccess";
-                        } else {
-                          accessKeyType = "FunctionCall";
-                        }
-                        return {
-                          accountId,
-                          publicKey: action.AddKey.public_key,
-                          accessKeyType
-                        };
-                      });
+                      return tx.actions
+                        .filter(action => action.AddKey !== undefined)
+                        .map(action => {
+                          let accessKeyType;
+                          if (
+                            action.AddKey.access_key.permission === "FullAccess"
+                          ) {
+                            accessKeyType = "FullAccess";
+                          } else if (
+                            action.AddKey.access_key.permission.hasOwnProperty(
+                              "FunctionCall"
+                            )
+                          ) {
+                            accessKeyType = "FunctionCall";
+                          } else {
+                            throw new Error(
+                              "permission type is either FullAccess nor FunctionCall"
+                            );
+                          }
+                          return {
+                            accountId,
+                            publicKey: action.AddKey.public_key,
+                            accessKeyType
+                          };
+                        });
                     })
                 ),
                 models.Account.bulkCreate(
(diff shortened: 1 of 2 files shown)


icerove:
@frol Theoretically, this should work fine. But it is not. Please help me to figure out where I did wrong since I could not find out.

backend/src/sync.js
```diff
@@ -91,6 +91,33 @@ async function saveBlocks(blocksInfo) {
                     });
                   })
                 ),
+                models.AccessKey.bulkCreate(
+                  blockInfo.transactions
+                    .filter(tx =>
```
frol, line 96: You are filtering the transactions which do not have any AddKey, but still include the wrole transaction with all the irrelevant actions. Drop the filter, flatMap on the transaction (that is already the case), filter the actions (this part is missing), and map the filtered values to the proper shape.

Notes:

1. you don't need to check for a stringly-typed action since `AddKey` is always represented as an object (where the key is "AddKey")
2. match the permission exhaustively, and throw an error if there is unknown permission type, so we catch it earlier.

backend/src/sync.js
```diff
@@ -91,6 +91,40 @@ async function saveBlocks(blocksInfo) {
...
```
frol, line 96: This filter looks redundant to me.

backend/src/sync.js
```diff
@@ -91,6 +91,40 @@ async function saveBlocks(blocksInfo) {
...
+                            accessKeyType = "FullAccess";
+                          } else if (
+                            action.AddKey.access_key.permission.hasOwnProperty(
+                              "FunctionCall"
+                            )
+                          ) {
+                            accessKeyType = "FunctionCall";
+                          } else {
+                            throw new Error(
+                              "permission type is either FullAccess nor FunctionCall"
```
frol, line 117: I would really have a hard time understanding this message if it would appear in the logs, wouldn't you?

frol:
We should keep in mind the possibly failing transactions, which already lead us to #223 and #252. Let's add a TODO into the code, please!

backend/src/sync.js
icerove: we have to have this since we need tx.receiverId as accountId and map the transactions directly instead of transactions.actions mapping. That is why filter all the transactions with "AddKey" is necessary. @frol

backend/src/sync.js
icerove: what is your suggestion? @frol

backend/src/sync.js
frol: Sorry, I don't follow what you are saying.

The function passed into `filterMap` will just return an empty list if there are no AddKey actions, and that will be just fine.

backend/src/sync.js
frol: Well, I would just drop the manual parsing altogether and just check that if that is a string, that is the accessKeyType, otherwise, use the `Object.keys()[0]` as the accessKeyType (ensuring that it is not undefined).

Error messages should be helpful, and to be so, we should include the relevant information into them, e.g.:

> Unexpected error during access key permission parsing in transaction ${tx.hash}: the permission type is expected to be a string or an object with a single key, but '${JSON.stringify(permission)}' found.

backend/src/sync.js
```diff
@@ -91,6 +91,43 @@ async function saveBlocks(blocksInfo) {
...
+                    const accountId = tx.receiver_id;
+                    return tx.actions
+                      .filter(action => action.AddKey !== undefined)
+                      .map(action => {
+                        let accessKeyType;
+                        if (
+                          typeof action.AddKey.access_key.permission ===
+                          "string"
+                        ) {
+                          accessKeyType = "FullAccess";
```
frol, line 105:
```
accessKeyType = action.AddKey.access_key.permission;
```

backend/src/sync.js
```diff
@@ -91,6 +91,43 @@ async function saveBlocks(blocksInfo) {
...
+                        } else if (
+                          action.AddKey.access_key.permission !== undefined
+                        ) {
+                          accessKeyType = Object.keys(
+                            action.a.access_key.permission
```
frol, line 110: 1. It is a typo here in `.a.`
2. Let's extract the `permission` into a variable before the `if`s to simplify the code (it will most certainly collapse multiline wraps)

PR #256, opened 2020-04-14
fix details api and ui

icerove:
fix 500 error when details api gives undefined and causes error

### DIFF
--- a/frontend/src/libraries/explorer-wamp/details.ts
+++ b/frontend/src/libraries/explorer-wamp/details.ts
@@ -32,7 +32,7 @@ export default class DetailsApi extends ExplorerApi {
           (SELECT COUNT(*) as lastDayTxCount FROM transactions
               WHERE block_timestamp > (strftime('%s','now') - 60 * 60 * 24) * 1000  ) as Daytransactions
           `
-        ]).then((it: any) => (it.length > 0 ? it[0] : undefined)),
+        ]).then((it: any) => it[0]),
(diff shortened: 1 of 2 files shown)


frontend/src/libraries/explorer-wamp/details.ts
```diff
@@ -32,7 +32,7 @@ export default class DetailsApi extends ExplorerApi {
           (SELECT COUNT(*) as lastDayTxCount FROM transactions
               WHERE block_timestamp > (strftime('%s','now') - 60 * 60 * 24) * 1000  ) as Daytransactions
           `
-        ]).then((it: any) => it[0]),
+        ]).then((it: any) => (it.length > 0 ? it[0] : undefined)),
```
frol, line 35: This is, essentially, a no-op since getting a non-existing index in JS returns undefined anyway.

frontend/src/libraries/explorer-wamp/details.ts
```diff
@@ -10,7 +10,7 @@ export interface Details {
 }
 
 export default class DetailsApi extends ExplorerApi {
-  async getDetails(): Promise<Details> {
+  async getDetails(): Promise<Details | undefined> {
```
frol, line 13: Can you try a question mark operator here?

`Promise<Details?>`

icerove: image

PR #257, opened 2020-04-15
feat(backend): Ignore the old telemetry requests since all the nodes are now updated

frol:


icerove:
do all three net all have the same format and information?

frol:
@icerove Yes! Finally, all the networks have the same format!

PR #258, opened 2020-04-16
fix frontend and add indexes to account

### LINKED ISSUE #185: Include information from genesis records
## Story

As a user who uses Explorer to get valid information, I want to see the full picture (specifically about accounts) so I can make sure that the network holds the data.

The API is currently only available as an experimental feature (on devnet and betanet): https://github.com/nearprotocol/nearcore/pull/2109, and there is [no near-api-js method yet](https://github.com/near/near-api-js/issues/271) (so we can work around it with [`sendJsonRpc`](https://github.com/near/near-api-js/pull/272)).

## Implementation Notes

Use `sendJsonRpc` to query [`EXPERIMENTAL_genesis_records` API](https://github.com/nearprotocol/nearcore/pull/2109), and save accounts to the DB (accounts and access keys tables).

## Acceptance Criteria

- [ ] The accounts created in genesis should be counted and included on the Accounts list page
- [ ] The Account details page work fine for the account created in genesis
- [ ] Tested

icerove:
fix #185 

backend:
1. generate nearcore-EXPERIMENTAL_genesis_records and nearcore-EXPERIMENTAL_genesis_config methods in the src/wamp.js since RPC has this and may use in the future.
2. generate saveGenesis method to insert records into data base. Generate syncGenesisState method in the src/sync.js and expose it in the src/index.js. Do not regular fetch genesis for an hour since it will occur unique error and not insert again at all. other are the same as regularSyncNewNearcoreState.
3. add accountIndex field to make it possible to scroll accounts from genesis. add indexes as transaction table to query 15 accounts.

frontend:
1. delete getAccountBasic function since hash and timestamp from genesis will not be null. Fix getAccountInfo to fetch all necessary account information same time. 
2. delete useless state in [id].js since time and hash is already in the props.

### DIFF
--- a/backend/src/sync.js
+++ b/backend/src/sync.js
@@ -1,5 +1,5 @@
 const models = require("../models");
-const moment = require("moment")
+const moment = require("moment");
 
 const {
   syncFetchQueueSize,
@@ -133,7 +133,7 @@ async function saveBlocks(blocksInfo) {
                     .map((tx, index) => {
                       return {
                         accountId: tx.receiver_id,
-                        accountIndex: timestamp * 1000000 + index, 
+                        accountIndex: timestamp * 1000000 + index,
                         createdByTransactionHash: tx.hash,
                         createdAtBlockTimestamp: timestamp
                       };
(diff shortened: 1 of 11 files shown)


backend/src/sync.js
```diff
@@ -149,6 +151,59 @@ async function saveBlocks(blocksInfo) {
...
+                    the permission type is expected to be a string or an object with a single key, 
+                    but '${JSON.stringify(permission)}' found.`
+                  );
+                }
+                return {
+                  accountId: record.AccessKey.account_id,
+                  publicKey: record.AccessKey.public_key,
+                  accessKeyType
+                };
+              })
```
frol, line 183: Is it possible to deduplicate this?

backend/src/sync.js
```diff
@@ -149,6 +151,59 @@ async function saveBlocks(blocksInfo) {
...
+          ),
+          models.Account.bulkCreate(
+            records
+              .filter(record => record.Account !== undefined)
+              .map((record, index) => {
+                return {
+                  accountId: record.Account.account_id,
+                  accountIndex: time + index,
```
frol, line 191: You use `timestamp * 1000000 + index` in another place. Are they synced?

backend/src/sync.js
icerove: almost cannot. its structure actually the same when insert into db, but data format is different

backend/src/sync.js
icerove: fixed

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -126,7 +126,7 @@ export default (
       switch (category) {
         case "Account":
           endTimestamp = this.state.items[this.state.items.length - 1]
-            .createdAtBlockTimestamp;
+            .accountIndex;
```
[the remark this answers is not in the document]
icerove: I do not have better solution here. there must be two indicator together to decide the scroll indicator now. If you have better solution, tell me

backend/src/sync.js
```diff
@@ -344,5 +399,31 @@ async function syncMissingNearcoreState() {
...
+  const limit = 100;
+  let offset = 0,
+    batchCount;
+  do {
+    let pagination = { offset, limit };
+    console.log(`Sync Genesis records from ${offset} to ${offset + limit}`);
+    const genesisRecords = await nearRpc.sendJsonRpc(
+      "EXPERIMENTAL_genesis_records",
+      [pagination]
+    );
```
frol, line 416: Considering that it has a good chance to fail at least once, there is very little chance that this sync will ever succeed since it will always start from the very beginning and will most probably fail again.

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -142,6 +145,7 @@ export default (
         default:
           endTimestamp = undefined;
       }
+      console.log(endTimestamp);
```
frol, line 148: Drop it.

Issue #201, opened 2020-02-23
Identify UI/UX papercuts


frol:
## Story

As a regular user who operates on the web daily, I want to have smooth and consistent UI and UX so I can feel that the product is reliable.

While we are not feature-complete yet, we have accumulated some hacks here and there. Solving those hacks with a consistent implementation now can boost our productivity in the future.

## Acceptance Criteria

- [ ] Identify main UI/UX papercuts (e.g. unhelpful titles, weird error messages (404 pages are odd), design inconsistencies)
- [ ] Create individual issues for every single area of work

## Notes

We could use the original design mocks to identify the missing design implementations.

frol:
This issue is superseded by #351

PR #266, opened 2020-04-24
Update/term description

### LINKED ISSUE #237: UI should help people to learn core concepts
## Story

As a regular user who just started learning about NEAR, I want to know what I am looking at so I can build my understanding of the system.

Currently, most of the numbers in our UI are quite unintuitive (including NEAR, Gas, number of nodes, block height).

There already have been a few questions about what Tgas is (that is Teragas, which is 10^12 gas).

I believe we should have hints with short explanations and links to the documentation.

## Acceptance Criteria

- [x] Each term in the Detail page tables has a tooltip/modal with explanation
  - [x] Account Detail
  - [x] Transaction Detail
  - [x] Block Detail
  - [x] Dashboard
- [x] Mobile ok
- [ ] Tested

## Notes

List of terms and explanations: https://docs.google.com/spreadsheets/d/1-4-zKSvf_v68AGFFmTgoKoGxtRXLW5dL7L7wMEdAdZg/edit#gid=0

icerove:
fix #237


icerove:
I redesign as Frol suggested.

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -25,12 +25,20 @@ export default class extends React.Component<Props> {
               title="Ⓝ Balance"
               text={<Balance amount={account.amount} />}
               className="border-0"
+              href={"https://docs.nearprotocol.com/docs/concepts/account"}
```
frol, line 28: 1. Looking at this interface I would expect that we define the whole CardCell to become a link when I pass `href`, but it is just a link to the docs in the popup... I think it does not make a lot of sense to make this separation and it makes things easier to customize if you just explicitly include `<a href...>docs</a>` in the `termDescription`.

2. `href` and `termDesription` do not belong to `CardCell`. Let's not abuse it, unless it is absolutely necessary. To me, it makes a lot of sense to just have a standalone component `TermHelpButton`, and you just use it in the title:

    ```jsx
            <CardCell
              title={
                <>
                  {"Ⓝ Balance "}
                  <TermHelpButton>
                    {"Total NEAR token balance. This includes the 'Locked' balance. "}
                    <a href="https://docs.nearprotocol.com/docs/concepts/account">docs</a>
                  </TermHelpButton>
                </>
              }
              text={<Balance amount={account.amount} />}
              className="border-0"
            />
    ```

frol:
@icerove The on-hover implementation looks great to me. Though, it still breaks some of the cells on Dashboard... I will just add a separate issue about that as part of #201 since we need some redesign.

icerove:
@frol 
>Looking at this interface I would expect that we define the whole CardCell to become a link when I pass href, but it is just a link to the docs in the popup
the wallet design is like that. We have the agreement to follow the wallet design. FurtherMore, if the whole cell is the link, how should we show the Modal and description? when click the cell, it will trigger link and cover the button to show the description.

icerove:
@frol 
>href and termDesription do not belong to CardCell. Let's not abuse it, unless it is absolutely necessary. 

Why? Every term description is in the `CardCell` and We do not have other to do so.

frol:
> the wallet design is like that. We have the agreement to follow the wallet design. FurtherMore, if the whole cell is the link, how should we show the Modal and description? when click the cell, it will trigger link and cover the button to show the description.

There is a misunderstanding.

I implied that it does not make any sense to me to have `href` for the CardCell as it is confusing (and I specifically described why I find that confusing).

I don't want to have either `href` or `termDescription` on CardCell component.

I want to have a separate `HelpButton` component doing exactly the thing it is designed to do:

1. display the small (i) button
2. handle the onClick event presenting the help message

frol:
> Why? Every term description is in the `CardCell` and We do not have other to do so.

Please, separate concerns as otherwise, we will end up with a monster component with millions of props for all sorts of cases.

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -22,20 +23,64 @@ export default class extends React.Component<Props> {
         <Row noGutters>
           <Col md="3">
             <CardCell
-              title="Ⓝ Balance"
+              title={
+                <>
+                  {"Ⓝ Balance"}
+                  <TermHelperButton title={"Balance"}>
```
frol, line 29: I clearly did not see the duplication. Since it is my fault, I will fix this.

PR #274, opened 2020-05-01
fix(backend): Block #0 should be synced

### LINKED ISSUE #262: Block #0 should be synced
Block #0 does not seem to get synced.

## Acceptance Criteria

- [x] The block #0 is synced (if available)
- [x] The Block Details page should display the information correctly for the first block from genesis (the parent block hash should say "Genesis" when `prev_hash == "11111111111111111111111111111111"`)
- [x] Manually tested

icerove:
fix #262

### DIFF
--- a/backend/src/sync.js
+++ b/backend/src/sync.js
@@ -15,10 +15,16 @@ async function saveBlocks(blocksInfo) {
       try {
         await models.Block.bulkCreate(
           blocksInfo.map((blockInfo) => {
+            let prevHash = blockInfo.header.prev_hash;
+            if (
+              (blockInfo.header.prev_hash = "11111111111111111111111111111111")
+            ) {
+              prevHash = "Genesis";
+            }
             return {
               hash: blockInfo.header.hash,
               height: blockInfo.header.height,
-              prevHash: blockInfo.header.prev_hash,
+              prevHash,
               timestamp: parseInt(blockInfo.header.timestamp / 1000000),
               totalSupply: blockInfo.header.total_supply || "",
               gasLimit: blockInfo.header.gas_limit || 0,
@@ -334,7 +340,7 @@ async function syncNewNearcoreState() {
   let latestSyncedBlockHeight;
   if (latestSyncedBlock !== null) {
     latestSyncedBlockHeight = latestSyncedBlock.height;
-    console.debug(`The latest synced block is #${latestSyncedBlockHeight}`);
+    console.debug(`### The latest synced block is #${latestSyncedBlockHeight}`);
   } else {
     latestSyncedBlockHeight = latestBlockHeight - 10;
     console.debug("There are no synced blocks, yet.");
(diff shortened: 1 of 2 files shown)


backend/src/sync.js
```diff
@@ -15,10 +15,16 @@ async function saveBlocks(blocksInfo) {
       try {
         await models.Block.bulkCreate(
           blocksInfo.map((blockInfo) => {
+            let prevHash = blockInfo.header.prev_hash;
+            if (
+              (blockInfo.header.prev_hash = "11111111111111111111111111111111")
+            ) {
+              prevHash = "Genesis";
+            }
```
frol, line 23: We don't need to do this during sync. It is better to keep the data of the same type. If we store the hash in base58 form, let's keep it that way.

backend/src/sync.js
```diff
@@ -334,7 +340,7 @@ async function syncNewNearcoreState() {
   let latestSyncedBlockHeight;
   if (latestSyncedBlock !== null) {
     latestSyncedBlockHeight = latestSyncedBlock.height;
-    console.debug(`The latest synced block is #${latestSyncedBlockHeight}`);
+    console.debug(`### The latest synced block is #${latestSyncedBlockHeight}`);
```
frol, line 343: Revert

backend/src/sync.js
```diff
@@ -348,10 +354,19 @@ async function syncOldNearcoreState() {
   let oldestSyncedBlockHeight = 0;
   if (oldestSyncedBlock !== null) {
     oldestSyncedBlockHeight = oldestSyncedBlock.height;
+    console.debug("-------------------------------------------------------");
```
frol, line 357: revert

backend/src/sync.js
```diff
@@ -348,10 +354,19 @@ async function syncOldNearcoreState() {
...
+  let syncOldBlockHeight = oldestSyncedBlockHeight;
+  const batch = 10000;
+  do {
+    console.debug(
+      `-------The oldest synced block is #${syncOldBlockHeight}-------`
+    );
+    let lowBlockHeight = Math.max(syncOldBlockHeight - batch, 0);
+    await syncNearcoreBlocks(syncOldBlockHeight - 1, lowBlockHeight);
+    syncOldBlockHeight -= batch;
+  } while (syncOldBlockHeight > 0);
```
frol, line 369: 1. `syncNearcoreBlocks` already does that in batches, so there is no need to implement this batching here.
2. It seems that the fix is just replace:

```diff
-  await syncNearcoreBlocks(oldestSyncedBlockHeight - 1, 1);
+  await syncNearcoreBlocks(oldestSyncedBlockHeight - 1, 0);
```

backend/src/sync.js
```diff
@@ -373,6 +388,9 @@ async function syncMissingNearcoreState() {
     if (lowHeight > highHeight) {
       return;
     }
+    console.debug(
```
frol, line 391: revert

frontend/src/components/blocks/BlockDetails.tsx
```diff
@@ -101,9 +101,13 @@ export default ({ block }: Props) => {
                   </Term>
                 }
                 text={
-                  <BlockLink blockHash={block.prevHash}>
-                    {block.prevHash}
-                  </BlockLink>
+                  block.prevHash === "Genesis" ? (
```
frol, line 104: After addressing the first comment, use `11111111111111111111111111111111` here.

icerove:
I do not see the ci, are the ci tests still running here?

frol:
It might be an issue of the GitHub org migration. Let me check.

PR #275, opened 2020-05-02
fix(frontend): many fixes to Node List page

### LINKED ISSUE #268: Nodes List page should present all the data
## Story

As a regular user who views Nodes List page, I want to be able to follow the UI and not be distracted or confused so I can better understand the data.

Currently, we have constant reordering on Nodes List page, and the decision to use `last_seen` as an anchor did not play well.

We should come up with a good anchor we can use on the Nodes List page to do pagination and it would be great to preserve the order between update even if the `last_seen` get mixed.

## Acceptance Criteria

- [ ] All the nodes are correctly listed on the Nodes List page
- [ ] There is no jumps / reorderings on the Nodes List page
- [ ] Mobile ok
- [ ] Manually tested

### LINKED ISSUE #269: Better sync status explanations for nodes
Right now internal constants are displayed like "AwaitingPeers" and "NoSync".

We should show more human understandable statuses:
 - AwaitingPeers -> "Waiting for peers"
 - NoSync -> we can just hide this status, as it means node is synced.
 - HeaderSync -> "Syncing headers"
 - BlockSync -> "Syncing blocks"
 - StateSync -> "Syncing state"
 - StateSyncDone -> "State sync is done"

### LINKED ISSUE #270: fix(frontend): Fix UI issues on Nodes List page
As per https://github.com/nearprotocol/near-explorer/pull/259#issuecomment-620549151

> Need to fix this spacing. These tabs need bottom margin. Also, I wouldn't put green circles on both tabs. It doesn't make sense.

Screen Shot 2020-04-28 at 6 27 01 AM

/cc @jakestutzman

### LINKED ISSUE #273: fix(frontend): Nodes list loads slowly on switching tabs
Switching between validating and non-validating tabs on the [Nodes list page](https://explorer.near.org/nodes/validators) takes 10 seconds, so I assume we don't call re-fetch the data immediately and rely on autorefresh handler. We should force fetch the data when switching the tabs

## Acceptance Criteria

- [ ] Nodes list loads quickly on switching the tabs
- [ ] Manually tested

icerove:
fix #270, fix #269, fix #273, fix #268


frontend/src/components/nodes/NodeRow.tsx
```diff
@@ -9,7 +9,46 @@ interface Props {
...
+        break;
+      case "BodySync":
+        explanation = "Syncing body";
+        break;
+      case "NoSync":
+      default:
+        explanation = undefined;
+    }
+    return explanation;
+  };
```
frol, line 45: Hashmap would be much shorter, and no need in saving it to state, just fetch the description by the status identifier right on render.

frontend/src/components/nodes/Nodes.tsx
```diff
@@ -21,6 +21,12 @@ export default class extends React.Component<Props> {
     return await new NodesApi().getNodes(count, this.props.role, endTimestamp);
   };
 
+  componentDidUpdate(prevProps: any) {
+    if (this.props.role !== prevProps.role) {
+      this.autoRefreshNodes = autoRefreshHandler(Nodes, this.config);
+    }
+  }
```
frol, line 28: It is a band-aid. autoRefreshHandler should do this automatically on props changes.

icerove: it supposed to do that. But I think the function name and parameter is not changed when it goes into the autoRefreshHandler, it won't trigger the update.

frol:
@icerove FYI, CI is there! It seems it was just a glitch since I could not find anything suspicious and only switched GitHub Actions off and on. :man_shrugging:

frontend/src/libraries/explorer-wamp/nodes.ts
```diff
@@ -30,7 +30,7 @@ export default class NodesApi extends ExplorerApi {
   ) {
     let whereClause = `WHERE last_seen > (strftime('%s','now') - 60) * 1000 `;
     if (endTimestamp) {
-      whereClause += ` AND last_seen < :endTimestamp`;
+      whereClause += ` AND node_id < :endTimestamp`;
```
frol, line 33: `node_id < :endTimestamp` is crazy :cry:

frontend/src/components/nodes/NodeRow.tsx
```diff
@@ -12,6 +12,15 @@ interface Props {
 export default class extends React.PureComponent<Props> {
   render() {
     const { node } = this.props;
+    let statusIdentifier = new Map([
```
frol, line 15: Let's move this to be a static field on the component class, we don't need to recreate it on every render.

backend/models/Node.js
```diff
@@ -65,7 +65,7 @@ module.exports = (sequelize, DataTypes) => {
       tableName: "nodes",
       underscored: true,
       timestamps: false,
-      indexes: [{ fields: ["last_seen"] }, { fields: ["is_validator"] }],
+      indexes: [{ fields: ["node_id"] }, { fields: ["is_validator"] }],
```
frol, line 68: 1. `node_id` is a primary field, so you don't need to create another index on it.
2. `last_seen` is used to determine the online status, so we can keep it indexed.

backend/models/Node.js
```diff
@@ -9,11 +9,11 @@ module.exports = (sequelize, DataTypes) => {
       },
       moniker: {
         type: DataTypes.STRING,
-        allowNull: false,
+        allowNull: true,
       },
       accountId: {
         type: DataTypes.STRING,
-        allowNull: false,
+        allowNull: true,
```
frol, line 16: It makes sense to just put an empty string if we don't get a string value on telemetry ping.

frontend/src/libraries/explorer-wamp/nodes.ts
icerove: I will change `endTimestamp` to `paginationIndexer`

backend/models/Node.js
icerove: I am not very sure how it works to fix #276 

@frol

frol: https://github.com/near/near-explorer/pull/283

frontend/src/libraries/explorer-wamp/nodes.ts
icerove: change it in another PR since this one is for node list issue.

PR #288, opened 2020-05-06
refactor(frontend): reliable sequence for account list

icerove:
find the sequence for account list page is not reliable and fix it to a more reliable way.


frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -63,23 +68,23 @@ export default class AccountsApi extends ExplorerApi {
...
     try {
       return await this.call("select", [
         `SELECT account_id as id, created_at_block_timestamp as createdAtBlockTimestamp, 
           created_by_transaction_hash as createdByTransactionHash, account_index as accountIndex
           FROM accounts
           ${
-            endTimestamp
-              ? `WHERE (created_at_block_timestamp * 1000000 + account_index) < :endTimestamp`
+            paginationIndexer
+              ? `WHERE created_at_block_timestamp <= :endTimestamp AND account_index < :accountIndex`
```
frol, line 80: This is not going to work.

account_index is a number that is sequentially assigned starting from 0 for every block, so many accounts will have `account_index` === 0, and this filter will filter out all the accounts with `account_index` >= 0.

```sql
WHERE created_at_block_timestamp < :endTimestamp OR (created_at_block_timestamp = :endTimestamp AND account_index <= :accountIndex)
```

Looking into it closer I think we are messing with things here. We are filtering the old accounts (where the timestamp is smaller than the given) and then we order them from newer to older (higher timestamp to lower timestamp)

icerove: I use the first one since the second is not correct. It will always fetch first 15 accounts and not fetching the  following

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -63,23 +68,23 @@ export default class AccountsApi extends ExplorerApi {
...
           FROM accounts
           ${
-            endTimestamp
-              ? `WHERE (created_at_block_timestamp * 1000000 + account_index) < :endTimestamp`
+            paginationIndexer
+              ? `WHERE created_at_block_timestamp < :endTimestamp OR (created_at_block_timestamp = :endTimestamp AND account_index <= :accountIndex)`
               : ""
           }
-          ORDER BY (created_at_block_timestamp * 1000000 + account_index) DESC
+          ORDER BY created_at_block_timestamp DESC, account_index DESC
```
frol, line 83: This query does not make sense to me.

Let's consider the following example:

```
account_id                     | created_at_block_timestamp
lol.test                       | 1
anon.test                      | 1
marcink60.test                 | 1
ff.test                        | 3
damons.test                    | 3
vitalik.test                   | 5
dev-1587569727975              | 6
marketm.test                   | 7
studio5215580270               | 7
bitcoinwallah.test             | 7
chironcoin_platform.test       | 10
studio-wn9p1yo8t               | 11
studio-a2qvmp7xk               | 12
studio-m4gn1ukx5               | 13
chironcoin_celine.test         | 16
chironcoin_matthieu.test       | 16
chironcoin_abhi.test           | 16
pengan.test                    | 16
dev-1587540578053              | 18
test-pr-517-ledger.test        | 19
```

```sql
SELECT
ORDER BY created_at_block_timestamp DESC
LIMIT 4
```

Output:

```
test-pr-517-ledger.test        | 19
dev-1587540578053              | 18
pengan.test                    | 16
chironcoin_abhi.test           | 16
```

after that auto-refresh does:

```sql
SELECT WHERE created_at_block_timestamp < 16
ORDER BY created_at_block_timestamp DESC
LIMIT 4
```

```
studio-m4gn1ukx5               | 13
studio-a2qvmp7xk               | 12
studio-wn9p1yo8t               | 11
chironcoin_platform.test       | 10
```

and after another refresh, it should go even further...

icerove: I tested it manually with real database and some dumb added account, the query is not like what you show, it will query all 16 rows since account_index is different, and then query 13, 12,11,10

frol: Please, set the page size to 4 and list all the queries and the output we get from Explorer frontend.

icerove:
the list of accounts:
first 4:
image

first 8:
image

first 12:
image

when query for total 12 accounts:
image

icerove:
also attach the local database accounts table to row 12:
image

PR #289, opened 2020-05-08
fix overflowing H1s

Patrick1904:
I noticed that H1s tend to overflow the page/screen on mobile devices. I’ve added a dynamic font size (for all H1s) based on screen width (min 28px, max 48px), and word-wrap for longer account IDs.
Screen Shot 2020-05-08 at 4 24 08 PM

### DIFF
--- a/frontend/src/pages/_app.jsx
+++ b/frontend/src/pages/_app.jsx
@@ -123,7 +123,22 @@ export default class extends App {
           }
 
           h1 {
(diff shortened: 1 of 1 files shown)


icerove:
Thank you for fixing this. However, I could not open the render link for this PR and I think it has some issues after you add this css. Please touch them and fix

icerove:
nice!

Issue #265, opened 2020-04-23
Update TPS to display different periods of time


frol:
## Story

As a regular user who estimates the capacity of NEAR Protocol, I want to see the stats over some lengthy period of time so I can better understand the picture.

I would like to see the latest TPS (we have it now) and also the peak TPS over 24-hours. There is a metric "load average" on unix systems, which accumulates an average CPU concurrency over 1-minute, 5-minute, and 15-minute periods, so we can just display those TPS as "101 / 17", and put the description in the pop-up help (#237).

## Acceptance Criteria

- [ ] Query backend for Peak 24h TPS
- [ ] Display the value on the dashboard
- [ ] Design fits the numbers up to 9999 TPS
- [ ] Manually tested

icerove:
@kcole16 @frol Can you give more details which peak TPS you actually want to show on the dashboard since Frol gives very different time zones. Should we show 1 minute, 5 minutes or 15 minutes peak TPS over 24 hrs?

kcole16:
The ideal "peak TPS" is identifying the 1 second period in the previous 24 hours with the most transactions and displaying that number. 

I believe @frol is saying we can choose from the unix accumulations (1, 5, 15 min) to calculate this?

frol:
Well, I actually suggested to display several peaks: last 24h / last hour / last minute. I think servers different needs people may have:
* last 24h answers the question: how busy is the network these days?
* last minute: is the network under a heavy load right now? Should I expect delays?
* last hour: was there a spike of load recently?

icerove:
at this point and according to many issues we have that needs design, we should collect them and redesign the board.

kcole16:
> Well, I actually suggested to display several peaks: last 24h / last hour / last minute.

Ah, got it. I do think that is a good idea, but also relatively low priority right now (especially last minute and hour) as our TX volume is very low.

frol:
We decided to remove TPS for now.

PR #293, opened 2020-05-13
feat(frontend): Replaced TPS with TXs (total number of transactions)

frol:
As per our discussion today, we decided to remove TPS from the dashboard as it is not useful at the moment and confuses people. #265


kcole16:
Looks good! Thinking we should expand on "TPD", perhaps "Transactions (24 hr)" or "Total TXs (24 hr)"?

ilblackdragon:
Let's instead show total number of transactions since the genesis.

Issue #278, opened 2020-05-02
Mark final blocks and non-final tx


ilblackdragon:
Currently it's unclear which blocks are final looking at block details or in the home page to latest blocks. We can add "final" tag for them to identify which ones are final.
Alternatively we can show non-final marker for blocks that are non-final yet.

Tx displayed from non-final blocks should also have a tag that they are confirmed but non final.

frol:
Implementation notes:

1. Let's expose `get-finality-stats` via WAMP (backend/src/wamp.js)
  a. Use the new-ish Block API to query block with finality=final (https://github.com/nearprotocol/docs/pull/330). near-api-js does not expose the new API, so it is better to contribute this new API support to near-api-js first.
  b. return the `header.height` as "final_block_height"
2. Query the new WAMP method from the relevant components (transactions list, transaction etails, blocks list, block details) and pass the "isFinal" flag (`block.height <= finalBlockHeight`) down to the dumb component displaying the marker on non-final transactions and blocks.

bowenwang1996:
@frol I don't think it needs to be that complicated. Every block has `last_final_height` in their header and since you already index blocks, you will know which blocks are supposedly final. However, there is a caveat: the blocks whose height are smaller than `last_final_height` of the head isn't necessarily final because they maybe on a different chain. But for practical purposes I think this is sufficient.

frol:
@bowenwang1996 We don't store that information anywhere. To implement this, we will need to come up with some storage for such stats (either global in-memory variable, or key-value SQLite table). Also, the sync process is not sequential, so we should also keep that in mind, so we cannot blindly update the value with the last fetched block.

Thus, querying RPC is just the most reliable way of doing this.

bowenwang1996:
@frol makes sense, but if you are willing to keep the head of the chain then you don't need to do additional queries. Ultimately this info is only good for the data that you have, so if you keep the head of the chain you don't need additional queries (even if you are out of sync at the moment).

icerove:
```
{
  "jsonrpc": "2.0",
  "id": "dontcare",
  "result": {
    "author": "node3",
    "header": {
      "height": 2003340,
      "epoch_id": "3mHzgVWca7GgBc9kQTUY1neKDqzUhkRbpUrMFs1pbWXL",
      "next_epoch_id": "ghpXGxdeLYR1BHvEDJwXF6LYhBsZjytvMkutFyHCtR5",
[...]

PR #297, opened 2020-05-19
Feature/finalizing block

icerove:
feature #278 with block part 

For tx, we do not store block height anymore and do we want it back and give isFinal tag with comparing block height and final height? Also, I remember we want to keep both execute status and isFinal status, so we should not add `Finalizing` to executive status, instead we have to create a new tag for isFinal.

### DIFF
--- a/frontend/src/libraries/explorer-wamp/blocks.ts
+++ b/frontend/src/libraries/explorer-wamp/blocks.ts
@@ -40,7 +40,7 @@ export default class BlocksApi extends ExplorerApi {
     paginationIndexer?: number
   ): Promise<BlockInfo[]> {
     try {
-      return await this.call("select", [
+      const blocks: BlockInfo[] = await this.call("select", [
         `SELECT blocks.*, COUNT(transactions.hash) as transactionsCount
           FROM (
             SELECT blocks.hash, blocks.height, blocks.timestamp, blocks.prev_hash as prevHash 
@@ -61,6 +61,13 @@ export default class BlocksApi extends ExplorerApi {
           paginationIndexer,
         },
       ]);
+      const finalHeight = await this.getFinalStats();
+      if (blocks.length > 0) {
+        blocks.map(
+          (block: any) => (block.isFinal = block.height <= finalHeight)
+        );
+      }
+      return blocks as BlockInfo[];
     } catch (error) {
       console.error("Blocks.getBlocks failed to fetch data due to:");
       console.error(error);
--- a/frontend/src/libraries/explorer-wamp/blocks.ts
+++ b/frontend/src/libraries/explorer-wamp/blocks.ts
@@ -8,8 +8,8 @@ export interface BlockInfo {
   transactionsCount: number;
   gasPrice: string;
   gasUsed: number;
+  isFinal: boolean;
 }
-
 export default class BlocksApi extends ExplorerApi {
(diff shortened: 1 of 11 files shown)


frontend/src/libraries/explorer-wamp/blocks.ts
```diff
@@ -87,15 +94,21 @@ export default class BlocksApi extends ExplorerApi {
           blockId,
         },
       ]).then((it) => (it[0].hash !== null ? it[0] : null));
-
+      const finalHeight = await this.getFinalStats();
+      block.isFinal = block.height <= finalHeight;
```
bowenwang1996, line 98: if `block` is `null`, wouldn't it crash here?

frontend/src/libraries/explorer-wamp/blocks.ts
```diff
@@ -61,6 +61,13 @@ export default class BlocksApi extends ExplorerApi {
           paginationIndexer,
         },
       ]);
+      const finalHeight = await this.getFinalStats();
+      if (blocks.length > 0) {
+        blocks.map(
+          (block: any) => (block.isFinal = block.height <= finalHeight)
```
bowenwang1996, line 67: Please enlighten me given my nonexistent JS skills. This code looks quite strange to me:
* what is the condition `block.length > 0` for?
* It seems to me that an imperative style would be better here, but please let me if this is some JS pattern.

cc @frol

frontend/src/libraries/explorer-wamp/blocks.ts
icerove: good catch

frontend/src/libraries/explorer-wamp/blocks.ts
icerove: 1. if wamp could not fetch, blocks will be null and length is 0
2. we use this pattern in transactions.ts, too.

frol: 1. `block.length > 0` is not required since an empty array is fine to be mapped.
2. `map` creates a new array of undefined in this case, and we don't need that. Instead, a more readable version would be a simple `for` loop, I agree with @bowenwang1996. Alternatively we can Use `forEach` instead of `map`, but I rarely find it more readable than a simple loop for mutable data.

frontend/src/libraries/explorer-wamp/blocks.ts
```diff
@@ -61,6 +61,13 @@ export default class BlocksApi extends ExplorerApi {
           paginationIndexer,
         },
       ]);
+      const finalHeight = await this.getFinalStats();
```
frol, line 64: This naming is really misleading, and you can notice it right at this line, you save "final *stats*" to "final *height*". Is height the only final stat? Either function, or extract the finalHeight from the final block header.

frontend/src/libraries/explorer-wamp/blocks.ts
frol: It will.

@icerove 

1. Do the final stats fetching in parallel with the select (use `await Promise.all`)
2. Move the assignment after the null check

frol:
Just for the record,

> For tx, we do not store block height anymore

Let's use blockTimestamp for that instead of height since both height and timestamp are monotonically increasing.

PR #301, opened 2020-05-22
fix(frontend): Contracts from genesis should be shown

### LINKED ISSUE #300: [Bug] Contracts from genesis are not shown
## Description
If you open a contract that was deployed in genesis, the account page won't show any contract information.

## Steps to reproduce
1. Open https://explorer.betanet.near.org/accounts/unknown.test
2. Observe that it doesn't show contract information.
3. Check `NEAR_ENV=betanet near state unknown.test` and observe `code_hash: 'DpywzzTNLvpUExfMg8x6yUPuNzEXjQ22D5MpnpKxQunK'`.

icerove:
fix #300

### DIFF
--- a/frontend/src/components/contracts/ContractDetails.tsx
+++ b/frontend/src/components/contracts/ContractDetails.tsx
@@ -29,15 +29,17 @@ export default class extends React.Component<Props, State> {
       .getContractInfo(this.props.accountId)
       .then((contractInfo) => {
         if (contractInfo) {
-          this.setState({
-            codeHash: contractInfo.codeHash,
-            transactionHash: contractInfo.transactionHash,
-            timestamp: contractInfo.timestamp,
-            locked: contractInfo.accessKeys.every(
-              (key: any) =>
-                key["access_key"]["permission"]["FunctionCall"] !== undefined
-            ),
-          });
+          if (contractInfo !== undefined) {
+            this.setState({
+              codeHash: contractInfo.codeHash,
+              transactionHash: contractInfo.transactionHash,
+              timestamp: contractInfo.timestamp,
+              locked: contractInfo.accessKeys.every(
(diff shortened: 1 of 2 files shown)


frontend/src/libraries/explorer-wamp/contracts.ts
```diff
@@ -26,12 +26,19 @@ export default class ContractsApi extends ExplorerApi {
...
-        };
+        if (contractInfo !== undefined) {
+          return {
+            codeHash,
+            transactionHash: contractInfo.hash,
+            timestamp: contractInfo.block_timestamp,
+            accessKeys: accessKeys.keys,
+          };
+        } else {
+          return {
```
frol, line 37: Let's mark the function to actually return the `Promise<ContractInfo>` and fix the interface definition as definitely don't match this return value.

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -29,15 +29,17 @@ export default class extends React.Component<Props, State> {
...
-          this.setState({
-            codeHash: contractInfo.codeHash,
-            transactionHash: contractInfo.transactionHash,
-            timestamp: contractInfo.timestamp,
-            locked: contractInfo.accessKeys.every(
-              (key: any) =>
-                key["access_key"]["permission"]["FunctionCall"] !== undefined
-            ),
-          });
+          if (contractInfo !== undefined) {
```
frol, line 32: I think we can live with a single check here :smile:

icerove: Sorry Frol, I do not understand what do you mean...... give me some hint about this

frol: image

PR #307, opened 2020-05-29
feat(backend): Save block author

### LINKED ISSUE #303: Show block author on Block Details page
We should have this in database already, so it's just to show this and link into account page.

icerove:
Add backend implementation for #303


frol:
Backend implementation is fine :+1: 

Do you want to merge it first and deal with the frontend in a separate PR?

PR #312, opened 2020-06-01
give genesis height as final fetching block

### LINKED ISSUE #311: Do not sync beyond the genesis height
## Story

Currently, Explorer backend tries to sync all the blocks down to block height 0, though there is no point in trying to do that given the fact that there are not blocks expected to be before the genesis height (available via EXPERIMENTAL_genesis_config API [`genesis_height`]).

This currently creates a huge load for no reason, and it seems that Explorer Backend uses excessive resources on this for betanet and gets killed.

## Acceptance Criteria

- [x] Sync only blocks with height greater than `genesis_height`
- [x] Manually Tested

## Implementation Details

Store the genesis height during the genesis sync and use it instead of the hardcoded `0` block height.

icerove:
fix #311

### DIFF
--- a/backend/src/sync.js
+++ b/backend/src/sync.js
@@ -9,6 +9,8 @@ const {
 const { nearRpc } = require("./near");
 const { Result, delayFor } = require("./utils");
 
+let genesisHeight;
+
 async function saveBlocks(blocksInfo) {
   try {
     await models.sequelize.transaction(async (transaction) => {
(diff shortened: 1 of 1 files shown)


backend/src/sync.js
```diff
@@ -350,8 +352,8 @@ async function syncOldNearcoreState() {
     oldestSyncedBlockHeight = oldestSyncedBlock.height;
     console.debug(`The oldest synced block is #${oldestSyncedBlockHeight}`);
   }
-
-  await syncNearcoreBlocks(oldestSyncedBlockHeight - 1, 0);
+  console.log("-------------------------------------------------");
```
frol, line 355: Let’s remove debug logs

PR #313, opened 2020-06-02
feat(backend): Improved the synchronization process

frol:
* Ignore duplicate records, so it does not throw useless exceptions
* Auto-shutdown when genesis height is changed
* Auto-recover the dead backend instances (essentially patches the recent
betanet downtimes and combining with the auto-shutdown we get an
auto-restart feature for free)
* The error message for the missing chunk is now useful

Test Plan
=========

* Manually tested that the duplicate records during blocks sync and
genesis sync do not throw exceptions
* Manually tested that the backend shuts down automatically when the
genesis is different from the last one we observed during the genesis
sync
* Manually tested that runit supervisor auto-recovers the backend
instances when they are killed
* Manually confirmed that the error messages are not spammy

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -13,11 +13,11 @@ RUN apk add --no-cache --virtual=.build-dependencies git && \
     chown nobody: . && \
     apk del .build-dependencies
 COPY ./ ./
-RUN mkdir -p ./db ./services/near-explorer-backend && \
+RUN mkdir ./db && \
     touch /etc/envvars && \
     echo 'export > /etc/envvars; for service in service_*; do if [ -s "$service" ]; then service_path=/etc/$(echo "$service" | sed 's:_:/:g'); mkdir -p "$(dirname "$service_path")"; cp "$service" "$service_path"; chmod +x "$service_path"; fi; done; exec /sbin/runsvdir /etc/service' > /opt/runit-bootstrap && \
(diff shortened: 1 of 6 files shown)


backend/Dockerfile
```diff
@@ -1,17 +1,24 @@
 FROM mhart/alpine-node:12
 
+USER root
+
+RUN apk add --no-cache runit
```
frol, line 5: I use runit as a proper supervisor instead of a hacky sh script I used before in order to be able to run several backends in a single Docker container.

backend/models/index.js
```diff
@@ -56,4 +56,17 @@ db.sequelize = sequelize;
...
 db.Sequelize = Sequelize;
 
+db.resetDatabase = function resetDatabase() {
+  if (config.dialect !== "sqlite") {
+    console.error(
+      `resetDatabase only supports sqlite dialect, but '${config.dialect}' found. No action is taken.`
+    );
+    return;
+  }
+  fs.renameSync(
```
frol, line 66: To reset a DB we do the following:

1. Rename the current sqlite DB file for the backup purposes
2. Exit the backend
3. Rely on the supervisor (`runit`) to bring the service back

frol:
> After ignoring the duplicate insert, will we lose some information from blocks or transactions?

The blocks and transactions should be immutable and should never change over time, so I don't expect to lose anything. However, the forks need to be addressed better, as currently, Explorer will include information from all the oberved forks.

PR #327, opened 2020-06-09
refactor(frontend): fetch transaction status lazily

### LINKED ISSUE #322: Explorer does not show all transactions
Take this account https://explorer.betanet.near.org/accounts/villiamsivertsen for example. On the account page it shows that there are 90 transactions but it only displays 20 transactions and seems to get stuck on loading the rest with error "Failed to load resource: the server responded with a status of 404 ()".

icerove:
fix #322

### DIFF
--- a/frontend/src/components/transactions/ActionRowBlock.tsx
+++ b/frontend/src/components/transactions/ActionRowBlock.tsx
@@ -69,7 +69,11 @@ export default class extends React.Component<Props, State> {
                   <Row>
                     <Col className="action-row-timer">
                       <span className="action-row-timer-status">
-                        <ExecutionStatus status={transaction.status} />
+                        {transaction.status ? (
+                          <ExecutionStatus status={transaction.status} />
+                        ) : (
+                          ""
+                        )}
                       </span>{" "}
                       <Timer time={transaction.blockTimestamp} />
                     </Col>
--- a/frontend/src/components/transactions/ActionRowBlock.tsx
+++ b/frontend/src/components/transactions/ActionRowBlock.tsx
@@ -18,6 +18,7 @@ export interface Props {
   title: React.ReactElement | string;
   children?: React.ReactNode;
   status?: T.ExecutionStatus;
+  isFinal?: boolean;
 }
 
 export interface State {}
@@ -38,6 +39,7 @@ export default class extends React.Component<Props, State> {
       icon,
       title,
       status,
+      isFinal,
       children,
     } = this.props;
 
@@ -72,6 +74,7 @@ export default class extends React.Component<Props, State> {
                     <Col className="action-row-timer">
                       <span className="action-row-timer-status">
                         {status ? <ExecutionStatus status={status} /> : ""}
+                        {isFinal ? "/Finalized" : "/Finalizing"}
                       </span>{" "}
                       <Timer time={transaction.blockTimestamp} />
                     </Col>
--- a/frontend/src/components/transactions/ActionRowBlock.tsx
+++ b/frontend/src/components/transactions/ActionRowBlock.tsx
@@ -17,6 +17,7 @@ export interface Props {
   icon: React.ReactElement;
   title: React.ReactElement | string;
   children?: React.ReactNode;
+  status?: T.ExecutionStatus;
 }
 
 export interface State {}
@@ -36,6 +37,7 @@ export default class extends React.Component<Props, State> {
       transaction,
       icon,
       title,
+      status,
       children,
     } = this.props;
 
@@ -69,11 +71,7 @@ export default class extends React.Component<Props, State> {
                   <Row>
                     <Col className="action-row-timer">
                       <span className="action-row-timer-status">
-                        {transaction.status ? (
-                          <ExecutionStatus status={transaction.status} />
-                        ) : (
-                          ""
-                        )}
+                        {status ? <ExecutionStatus status={status} /> : ""}
                       </span>{" "}
                       <Timer time={transaction.blockTimestamp} />
                     </Col>
--- a/frontend/src/components/transactions/ActionRowBlock.tsx
+++ b/frontend/src/components/transactions/ActionRowBlock.tsx
@@ -73,8 +73,16 @@ export default class extends React.Component<Props, State> {
                   <Row>
                     <Col className="action-row-timer">
                       <span className="action-row-timer-status">
-                        {status ? <ExecutionStatus status={status} /> : ""}
-                        {isFinal ? "/Finalized" : "/Finalizing"}
+                        {status ? (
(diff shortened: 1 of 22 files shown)


icerove:
Need some help and comment for this PR. Separate query nearcore-tx method from getTransactions and give it a new method getTransactionStatus. Now fetching transactions from database is the first step and fetching API for status and actions and outcomes are the second step.

frontend/src/components/transactions/ActionsList.tsx
```diff
@@ -5,7 +5,7 @@ import ActionRow from "./ActionRow";
 import { ViewMode, DetalizationMode } from "./ActionRowBlock";
 
 export interface Props {
-  actions: T.Action[];
+  actions?: T.Action[];
```
frol, line 8: There is no point in using this component if you don't have Actions, isn't it? I suggest we revert this change and conditionally use this component from the parent component based on the `actions` existence.

frontend/src/components/transactions/TransactionAction.tsx
```diff
@@ -1,5 +1,6 @@
 import React from "react";
-import * as T from "../../libraries/explorer-wamp/transactions";
+import { Row } from "react-bootstrap";
+import TransactionsApi, * as T from "../../libraries/explorer-wamp/transactions";
```
frol, line 3: https://github.com/near/near-explorer/blob/master/CONTRIBUTING.md#imports-ordering-and-grouping

frontend/src/components/transactions/ActionRowBlock.tsx
```diff
@@ -69,7 +73,8 @@ export default class extends React.Component<Props, State> {
                   <Row>
                     <Col className="action-row-timer">
                       <span className="action-row-timer-status">
-                        <ExecutionStatus status={transaction.status} />
+                        {status ? <ExecutionStatus status={status} /> : ""}
+                        {isFinal ? "/Finalized" : "/Finalizing"}
```
frol, line 77: Having `isFinal` as undefined (given the types I assume it is a valid case) should not be displayed as "Finalizing".

I suggest we display "Fetching Status... / Checking Finality..." when we don't have the information yet.

frontend/src/components/transactions/TransactionAction.tsx
```diff
@@ -13,39 +14,80 @@ export interface Props {
...
+    );
+    this.setState({ status });
+  };
+
+  fetchFinalStamp = async () => {
+    const finalStamp = await new TransactionsApi().queryFinalTimestamp();
+    this.setState({
+      isFinal: this.props.transaction.blockTimestamp <= finalStamp,
+    });
+  };
```
frol, line 55: Let's use React Context on the App component (pages/_app.jsx) in order to avoid fetching the final timestamp on every single TransactonAction. Final timestamp and block height is useful throughout the Explorer, so it seems that the App component is a reasonable place to query and provide them via Context to all the components below.

frontend/src/components/transactions/TransactionAction.tsx
```diff
@@ -13,39 +14,80 @@ export interface Props {
...
+
   render() {
     const { transaction, viewMode } = this.props;
-
-    if (this.props.actions.length !== 1) {
+    const { status, isFinal } = this.state;
+    if (transaction.actions) {
+      if (transaction.actions.length !== 1) {
```
frol, line 61: I suggest we do short-ciruit error handling, which will reduce the nesting:

```js
if (!actions) {
  return <></>
}
if (actions.length !== 1) {
  return ...
}
return ...
```

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -173,36 +172,22 @@ export default class TransactionsApi extends ExplorerApi {
...
-            // Given we already queried the information from the node, we can use the actions,
-            // since DeployContract.code and FunctionCall.args are stripped away due to their size.
-            //
-            // Once the above TODO is resolved, we should just move this to TransactionInfo method
-            // (i.e. query the information there only for the specific transaction).
-            const actions = transactionExtraInfo.transaction.actions;
-            transaction.actions = actions.map((action: RpcAction | string) => {
+            transaction.actions = actions.map((action: any) => {
               if (typeof action === "string") {
                 return { kind: action, args: {} };
```
frol, line 186: This branch is not possible when we fetch the information from the DB since we already preprocessed the data.

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -219,6 +204,27 @@ export default class TransactionsApi extends ExplorerApi {
...
+      transaction.signerId,
+    ]);
+    const status = Object.keys(
+      transactionExtraInfo.status
+    )[0] as ExecutionStatus;
+    // Given we already queried the information from the node, we can use the actions,
+    // since DeployContract.code and FunctionCall.args are stripped away due to their size.
+    //
+    // Once the above TODO is resolved, we should just move this to TransactionInfo method
+    // (i.e. query the information there only for the specific transaction).
```
frol, line 224: The comment is not relevant here.

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -243,9 +249,13 @@ export default class TransactionsApi extends ExplorerApi {
...
       } else {
         const transactionExtraInfo = await this.call<any>("nearcore-tx", [
-          transactionHash,
+          transactionInfo.hash,
           transactionInfo.signerId,
         ]);
+        transactionInfo.status = Object.keys(
+          transactionExtraInfo.status
+        )[0] as ExecutionStatus;
+
```
frol, line 258: We also need the `actions` extracted from the `transactionExtraInfo`, as Explorer Backend strips DeployContract contract binary and FunctionCall args from the data and don't save it to the DB.

We used to do this on getTransactions:
https://github.com/near/near-explorer/blob/74b1cb84f4821b54a2b8e5fe20c06e1a16f15c10/frontend/src/libraries/explorer-wamp/transactions.ts#L192-L209

We don't need to do it on getTransactions since actions saved on the backend to the DB includes enough information for the pages where we display lists of transactions, but on the Transaction Details page we need more, hence the need to fetch an parse the data from RPC.

frontend/src/libraries/explorer-wamp/transactions.ts
icerove: I will delete it directly

frontend/src/components/transactions/TransactionAction.tsx
icerove: block height is not in this repo, I will fix the final timestamp first and after merge this, will add block height for another PR.

Issue #305, opened 2020-05-25
Display more information about the validators


frol:
## Story

As an on-call network engineer who monitors the health of the network, I want to visually distinguish the nodes that last block height is too old or producing fewer blocks than expected so I can quickly build some understanding of the current state of the network.

## Acceptance Criteria

- [x] Change the text color based on how far the last block height reported via telemetry from the last block height we know (if the difference is more than 1000, display the last block height in red, >50 - orange)
- [x] Display the produced / expected blocks (display the raw numbers + percentage). Expose the `validators` API to fetch the `current_validators` list via Explorer backend
- [ ] Let's try to write end-to-end tests for these cases using the cypress facility we have configured

icerove:
for step 3, test part. Would you mind tell more details and hint? @frol

frol:
@icerove see into https://github.com/near/near-explorer/blob/74b1cb84f4821b54a2b8e5fe20c06e1a16f15c10/frontend/cypress/integration/dashboard.spec.js. You run end-to-end tests with `npm run test:ci`. Otherwise, learn more about Cypress and read `frontend/package.json` scripts.

PR #331, opened 2020-06-12
feat(frontend): display stake of the validation nodes

### LINKED ISSUE #309: Show stake for validators
In the list of validators, it would be good to show the current stake of the validators.

Also, if the account confirmed (signed), which is true for validators - make it clickable for details.

icerove:
fix #309 and #305

### DIFF
--- a/frontend/src/components/dashboard/DashboardHeader.tsx
+++ b/frontend/src/components/dashboard/DashboardHeader.tsx
@@ -132,10 +132,7 @@ export default class extends React.Component<State> {
               text={
                 details ? (
                   <Link href="/nodes/[role]" as={`/nodes/validators`}>
-                    <a
-                      id="node-page"
-                      style={{ textDecoration: "none", color: "#24272a" }}
-                    >
+                    <a id="node-page" className="dashboard-link">
                       {details.validatorsCount.toLocaleString()}/
                       {details.onlineNodesCount.toLocaleString()}
                     </a>
@@ -222,7 +219,7 @@ export default class extends React.Component<State> {
               text={
                 details ? (
                   <Link href="/accounts">
-                    <a style={{ textDecoration: "none", color: "#24272a" }}>
+                    <a className="dashboard-link">
                       {details.accountsCount.toLocaleString()}
                     </a>
                   </Link>
@@ -298,6 +295,11 @@ export default class extends React.Component<State> {
 
           .button-search {
             border-radius: 25px;
+          }
+
+          .dashboard-link {
+            text-decoration: none;
+            color: #24272a;
           }
         `}</style>
       </div>
--- a/frontend/src/components/dashboard/DashboardHeader.tsx
+++ b/frontend/src/components/dashboard/DashboardHeader.tsx
@@ -132,7 +132,10 @@ export default class extends React.Component<State> {
               text={
                 details ? (
                   <Link href="/nodes/[role]" as={`/nodes/validators`}>
-                    <a style={{ textDecoration: "none", color: "#24272a" }}>
+                    <a
+                      id="node-page"
+                      style={{ textDecoration: "none", color: "#24272a" }}
+                    >
                       {details.validatorsCount.toLocaleString()}/
                       {details.onlineNodesCount.toLocaleString()}
                     </a>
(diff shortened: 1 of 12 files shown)


frontend/src/libraries/explorer-wamp/nodes.ts
```diff
@@ -39,22 +40,34 @@ export default class NodesApi extends ExplorerApi {
...
+        ]),
+        this.queryValidators(),
       ]);
+
+      nodes.map((node: NodeInfo) => {
+        validators.map((val: any) => {
+          if (val.account_id === node.accountId) {
+            node.stake = val.stake;
+          }
+        });
```
bowenwang1996, line 68: this is too inefficient :(

icerove: do you have other option to do so? @bowenwang1996

icerove: will change to map

frontend/src/components/nodes/NodeRow.tsx
```diff
@@ -32,7 +33,9 @@ export default class extends React.PureComponent<Props> {
         <Col md="7" xs="7">
           <Row>
             <Col className="node-row-title">
-              @{node.accountId}{" "}
+              @{node.accountId}
+              {"   "}
+              <span>${node.stake && <Balance amount={node.stake} />}</span>
```
frol, line 38: 1. I think we don't need that `$` sign
2. I would like to have some indication that it is a stake, not an account balance or anything else. @corwinharrell @kcole16 any suggestions on how to make this clear?

frontend/src/libraries/explorer-wamp/nodes.ts
```diff
@@ -39,22 +40,34 @@ export default class NodesApi extends ExplorerApi {
...
+              ORDER BY node_id DESC
+              LIMIT :limit
+          `,
+          {
+            limit,
+            paginationIndexer,
+            validatorIndicator,
+          },
+        ]),
+        this.queryValidators(),
```
frol, line 60: Let's decouple fetching the list of nodes based on telemetry from the RPC query. Nodes List page is quite important and we should not block the display when RPC is unreachable.

frontend/src/components/nodes/NodeRow.tsx
kcole16: Something like "Staking 100,000 N"?

frol: @kcole16 keep it stupid simple! I love it! :smile:

frontend/src/components/nodes/NodeRow.tsx
```diff
@@ -20,8 +25,28 @@ export default class extends React.PureComponent<Props> {
...
     ["NoSync", ""],
   ]);
+
+  state: State = { blockDiff: 0 };
+
+  componentDidMount() {
+    new BlocksApi().getLatestBlockHeight().then((lastBlockHeight) =>
+      this.setState({
+        blockDiff: Math.abs(
+          lastBlockHeight.lastBlockHeight - this.props.node.lastHeight
```
bowenwang1996, line 35: what is this for?

icerove: https://github.com/near/near-explorer/issues/305

frol: There is no point in saving it to state, it is actually more expensive to save it than to compute. Also, this latest block height should be fetched only once, and passed via React Context, so we don't query it for every single node. Also, make sure that it takes into account the auto-refresh logic.

icerove:
Frol, I learned a lot about cypress test and find out it is really useful for end to end test. I recommend merge this PR for the feature of validators and create a new issue and PR for the cypress test since on local net, there is no node list on node page and on https://explorer.betanet.near.org/ there is no new feature listed. After this PR merge, I will write end-to-end test for most of explorer pages.
@frol

frontend/src/components/dashboard/DashboardHeader.tsx
```diff
@@ -132,7 +132,10 @@ export default class extends React.Component<State> {
               text={
                 details ? (
                   <Link href="/nodes/[role]" as={`/nodes/validators`}>
-                    <a style={{ textDecoration: "none", color: "#24272a" }}>
+                    <a
+                      id="node-page"
```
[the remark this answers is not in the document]
icerove: for simulate test

PR #332, opened 2020-06-15
feat(backend): Added a new feature to disable automatic backups on network reset to avoid spending too much disk space

frol:
This will be used on production for devnet since we do heavy load there and that produces 5GB database file every day.

### DIFF
--- a/backend/src/sync.js
+++ b/backend/src/sync.js
@@ -10,7 +10,6 @@ const { nearRpc } = require("./near");
 const { Result, delayFor } = require("./utils");
 
 let genesisHeight;
-let genesisTime;
 
 async function saveBlocks(blocksInfo) {
   try {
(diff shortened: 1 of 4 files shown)


backend/src/sync.js
```diff
@@ -428,22 +427,8 @@ async function syncGenesisState() {
   const genesisConfig = await retry(10, () =>
     nearRpc.sendJsonRpc("EXPERIMENTAL_genesis_config")
   );
-  if (
```
frol, line 431: This has been extracted in a separate routine into `index.js` file.

backend/src/sync.js
```diff
@@ -454,7 +439,7 @@ async function syncGenesisState() {
     const genesisRecords = await retry(10, () =>
       nearRpc.sendJsonRpc("EXPERIMENTAL_genesis_records", [pagination])
     );
-    await saveGenesis(parsedGenesisTime, genesisRecords.records, offset);
+    await saveGenesis(genesisTime, genesisRecords.records, offset);
```
bowenwang1996, line 442: why is it indexed by genesis time instead of genesis height?

icerove: We need genesis time recorded and display on the frontend. We use genesis height other place to check and query blocks.

frol: It is not indexed by that. We use genesis time to indicate the creation time of Accounts.

PR #348, opened 2020-07-10
Feature(frontend)/add publish and subscritbe system to display aggregate data

### LINKED ISSUE #341: fix(frontend): Fresh page load of pages with autoRefreshHandler should load fine on initial load
## Story

As a user who opens Blocks / Accounts / Dashboard / Nodes by a direct link, I experience problems with page loads as the rendered content appears for a brief moment and immediately replaced with a loading indicator, and the list of data gets finally rendered only in 10 seconds (when the next iteration of autoRefreshHandler happens).

## Acceptance Criteria

- [ ] The pages load fine when opening them via direct link
- [ ] The pages load fine when opening them via client-side routing (it works fine now, but we should make sure we don't introduce degradation)

icerove:
fix #341 

1. delete all the auto refresh for the list pages(nodes,blocks, transactions, accounts and dashboard)
2. add wamp publish in the backend/src/index.js. The publish information includes all dashboard information and finalized block timestamp. 
3. add query final block timestamp and validating and proposal nodes methods in the near.js. 

3. add all wamp subscribe method in the frontend/src/context folder. this one need more concern to review.

4. rename autoRefreshHandler into ListHandler and all list page also change this except all nodes page, node page are lists all and do not pagination anymore


backend/src/wamp.js
```diff
@@ -101,6 +103,8 @@ function setupWamp() {
         return;
       }
     }
+
+    wamp_session = session;
```
frol, line 107: [According to the source code](https://github.com/crossbario/autobahn-js/blob/9097e868e2bc46e599b1c16e6ba25c3dc14bb644/packages/autobahn/lib/connection.js#L416-L420), `wamp` instance should have `.session` property, so you don't need to save it separately, just use `wamp.session`

frontend/src/components/utils/autoRefreshHandler.tsx
```diff
@@ -33,15 +33,7 @@ export default (
...
     componentDidMount() {
-      this.timer = setTimeout(this.regularFetchInfo, 0);
-    }
-
-    componentWillUnmount() {
-      const timer = this.timer;
-      this.timer = null;
-      if (timer !== null) {
-        clearTimeout(timer);
-      }
```
frol, line 44: Won't this break pages like Account Details, Block Details?

backend/src/sync.js
```diff
@@ -232,6 +233,33 @@ function promiseResult(promise) {
   });
 }
 
+function blocksPublish(blocks) {
+  if (blocks.length > 0) {
+    wampPublish("blocks", blocks);
+    let accountAmount = 1;
```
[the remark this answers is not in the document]
icerove: what do you mean?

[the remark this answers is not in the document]
icerove: my fault, it should be 0. I use 1 to test the content.

backend/src/sync.js
```diff
@@ -232,6 +233,33 @@ function promiseResult(promise) {
...
+            ) {
+              return 1;
+            } else {
+              return 0;
+            }
+          })
+        )
+      )
+      .flat();
+    actions = [...actions];
```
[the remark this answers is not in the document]
icerove: yes, it will send the info like [[0]] and [[1]] or other number if this block has more than one create account txs.

backend/src/sync.js
```diff
@@ -305,6 +333,7 @@ async function saveBlocksFromRequests(requests) {
     )
   ).filter((block) => block !== null);
 
+  blocksPublish(blocks);
```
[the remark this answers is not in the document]
icerove: this is reasonable place to publish the new syncing blocks, otherwise we have to wait blocks stored in the database and then fetch from database.

backend/src/wamp.js
icerove: I tried, `wamp.session` is null

frontend/src/components/utils/autoRefreshHandler.tsx
icerove: We never apply autoRefreshHandler to details pages, we only use it on list pages

backend/src/wamp.js
frol: There is something wrong with your experiment since [according to the source code](https://github.com/crossbario/autobahn-js/blob/9097e868e2bc46e599b1c16e6ba25c3dc14bb644/packages/autobahn/lib/connection.js#L292) `onopen` is called with the same self._session as it returns from `wamp.session`. `wamp.session` will be null when it is not connected or was disconnected, in your case you keep even dead `wamp_session`, which may lead to many problems in the future. Use `wamp.session` and let me know if you still cannot get it working.

backend/src/wamp.js
```diff
@@ -113,4 +117,16 @@ function setupWamp() {
   return wamp;
 }
 
+function wampPublish(topic, args) {
```
frol, line 120: It is suboptimal as this function relies on the fact that someone will call `setupWamp`, and only once. It is currently the case, and this implementation will work most of the time, but I wanted to let you know that ideally we should avoid such assumptions and dependencies on singletones.

backend/src/sync.js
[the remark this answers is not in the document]
icerove: I change it and I wonder why I used to think this is correct

backend/src/wamp.js
icerove: not very sure what do you mean by this, but I will change the whole logic and let you review

backend/src/sync.js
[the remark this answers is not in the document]
icerove: I changed it a little bit, it is not straight to just select the number of total and then get increment in the provider.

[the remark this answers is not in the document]
icerove: I tried it many many time to get increment of blocks and transactions and accounts. Either tried it on frontend or backend. Need your help here. The best method is still want you give, but when to fetch `last_published_block_timestamp` I still could not figure out. @frol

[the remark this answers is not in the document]
icerove: this part need your more concern and help.

frontend/src/components/blocks/BlockDetails.tsx
```diff
@@ -14,151 +15,175 @@ export interface Props {
...
-                  <Term title={"Hash"}>
-                    {"Unique identifier (hash) of this block. "}
-                  </Term>
-                }
-                text={block.hash}
-                className="border-0"
-              />
-            </Col>
-          </Row>
+    <RpcConsumer>
```
icerove, line 18: this file only add consumer.

frontend/cypress/integration/nodepage.spec.js
```diff
@@ -2,7 +2,7 @@
 
 context("Dashboard", () => {
   beforeEach(() => {
-    cy.visit("http://localhost:3000");
+    cy.visit("https://explorer.betanet.near.org");
```
icerove, line 5: local one has some weird issue and can not pass.

frontend/src/components/blocks/BlocksRow.tsx
```diff
@@ -17,107 +18,115 @@ export default class extends React.PureComponent<Props> {
...
-                style={{ width: "20px" }}
-              />
-            </Col>
-            <Col md="7" xs="6">
-              <Row>
-                <Col className="transaction-row-title">#{block.height}</Col>
-              </Row>
-              <Row>
-                <Col className="transaction-row-text">
+      <RpcConsumer>
```
icerove, line 21: this file only add consumer

frontend/src/components/dashboard/DashboardBlocksBlock.tsx
```diff
@@ -4,134 +4,149 @@ import React from "react";
...
-              className="dashboard-blocks-block"
-              style={{ background: block.isFinal ? "#fff" : "#f8f7f8" }}
-            >
-              <Card.Title className="dashboard-blocks-block-title">
-                #{block.height}
-              </Card.Title>
-              <Card.Body className="dashboard-blocks-block-content">
-                <p className="dashboard-blocks-block-content-p">
-                  {block.isFinal ? (
+      <RpcConsumer>
```
icerove, line 16: add consumer for finalTimestamp

frontend/src/libraries/explorer-wamp/blocks.ts
```diff
@@ -41,9 +40,8 @@ export default class BlocksApi extends ExplorerApi {
     paginationIndexer?: number
   ): Promise<BlockInfo[]> {
     try {
-      const [blocks, finalHeight] = await Promise.all([
-        this.call<any>("select", [
-          `SELECT blocks.*, COUNT(transactions.hash) as transactionsCount
+      const blocks = await this.call<any>("select", [
```
icerove, line 43: delete query finalTimestamp method since we already have this from provider

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -272,9 +272,4 @@ export default class TransactionsApi extends ExplorerApi {
       throw error;
     }
   }
-
-  async queryFinalTimestamp(): Promise<any> {
```
icerove, line 276: no need anymore since finalTimestamp is from Rpc provider

frontend/cypress/integration/nodepage.spec.js
frol: This is not correct. The tests should test the local deployment, not production. What is the issue you have with your local deployment?

frontend/src/components/dashboard/DashboardHeader.tsx
```diff
@@ -69,240 +59,225 @@ export default class extends React.Component<State> {
...
+                          {context.onlineNodeAmount.toLocaleString()}
+                        </a>
+                      </Link>
+                    ) : (
+                      ""
+                    )
+                  }
+                  className="border-0"
+                  loading={
+                    !!!context.validatorAmount && !!!context.onlineNodeAmount
```
frol, line 99: Why `!!!`? Wouldn't one `!` be enough?

frontend/src/components/utils/Update.tsx
```diff
@@ -0,0 +1,38 @@
...
+export default ({ category, count }: Props) => {
+  if (count === 0) {
+    return null;
+  }
+  return (
+    <div className="update">
+      {count > 1 ? (
+        <p>
+          {count} new {category}s. Refresh or Click to view the latest{" "}
+          {category}s.
```
frol, line 15: I find string interpolation easier to maintain as it is a string:

```suggestion
          {`${count} new ${category}s. Refresh or Click to view the latest ${category}s.`}
```

frontend/src/components/utils/Update.tsx
```diff
@@ -0,0 +1,38 @@
...
+        </p>
+      ) : (
+        <p>
+          {count} new {category}. Refresh or Click to view the latest {category}
+          .
```
frol, line 20:
```suggestion
          {`${count} new ${category}. Refresh or Click to view the latest ${category}.`}
```

backend/src/index.js
```diff
@@ -94,6 +104,96 @@ async function main() {
...
+        `SELECT COUNT(*) as total FROM transactions
+        WHERE block_timestamp > (strftime('%s','now') - 60 * 60 * 24) * 1000`,
+      ]),
+    ]);
+    return [
+      totalAccounts.total,
+      totalBlocks.total,
+      totalTransactions.total,
+      lastDayTxCount.total,
+    ];
```
frol, line 150: Whenever you return more than a single value, let's prefer named values as it is very easy to mess things up in JS and left it unnoticed for quite some time.

```suggestion
    return {
      totalAccounts: totalAccounts.total,
      totalBlocks: totalBlocks.total,
      totalTransactions: totalTransactions.total,
      lastDayTxCount: lastDayTxCount.total,
    };
```

backend/src/index.js
```diff
@@ -94,6 +104,96 @@ async function main() {
   const wamp = setupWamp();
   console.log("Starting WAMP worker...");
   wamp.open();
+
+  const wampPublish = (topic, args) => {
+    const uri = `com.nearprotocol.${wampNearNetworkName}.explorer.${topic}`;
+    wamp.session.publish(uri, args);
+  };
+
+  const wampCall = async (args) => {
```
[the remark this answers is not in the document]
icerove: change to wampSqlSelectQuery

frontend/cypress/integration/nodepage.spec.js
icerove: will change back and check carefully

frontend/src/components/dashboard/DashboardHeader.tsx
```diff
@@ -7,28 +7,18 @@ import { Button, Col, FormControl, InputGroup, Row } from "react-bootstrap";
 import AccountsApi from "../../libraries/explorer-wamp/accounts";
 import BlocksApi from "../../libraries/explorer-wamp/blocks";
 import TransactionsApi from "../../libraries/explorer-wamp/transactions";
-import DetailsApi, * as D from "../../libraries/explorer-wamp/details";
+import { SubConsumer } from "../../context/SubscriptionProvider";
```
frol, line 10: Please, don't create shortcut names that are hard to read. It makes the review and maintenance process harder.

frontend/src/context/SubscriptionProvider.jsx
```diff
@@ -0,0 +1,181 @@
...
+    }
+    if (nodeInfo.proposalAmount !== proposalAmount) {
+      dispatchNode({ type: "proposals", proposalAmount, proposals });
+    }
+  };
+
+  const Subscription = useCallback(() => {
+    new ExplorerApi().subscribe("data-stats", fetchNewStats);
+    new ExplorerApi().subscribe("final-timestamp", fetchFinalTimestamp);
+    new ExplorerApi().subscribe("nodes", fetchNodeInfo);
```
frol, line 161: Why did you choose to merge all of these into a single context? This way a user opening ANY page of Explorer will get flooded with irrelevant data (e.g. opening a Block details page one will still get online nodes and data-stats)

I suggest we split the context into smaller chunks. Currently, we have a monolith and I hardly see we can move forward with it adding new features.

frontend/src/components/nodes/Nodes.tsx
```diff
@@ -1,36 +1,27 @@
...
-    }
-    if (!nodes) {
-      return <PaginationSpinner hidden={false} />;
-    }
-    return <>{nodes}</>;
+    return (
+      <SubConsumer>
+        {(context) => (
+          <>
+            {context.nodeInfo.onlineNodes ? (
```
frol, line 15: If I open a Nodes list page, do I need to wait 10 seconds to see any data? That is suboptimal.

frontend/src/components/blocks/BlockDetails.tsx
```diff
@@ -14,151 +15,175 @@ export interface Props {
...
+                    title={
+                      <Term title={"Transactions"}>
+                        {"Number of transactions in this block. "}
+                        <a
+                          href={
+                            "https://docs.nearprotocol.com/docs/concepts/transaction"
+                          }
+                        >
+                          docs
+                        </a>
```
frol, line 35: Note, we need to update `docs.nearprotocol.com` with `docs.near.org`

icerove: let's make this a new PR

frontend/src/context/SubscriptionProvider.jsx
icerove: I had tried to separate these subscribe function into different component and the result is that only can subscribe successfully and others will be out of work. If you have other options to do so, please write the hint down. @frol

backend/src/index.js
```diff
@@ -91,9 +97,151 @@ async function main() {
...
+  const addNodeInfo = async (nodes) => {
+    const accountArray = nodes
+      .map((node) => '"' + node.account_id + '"')
+      .join(",");
+    let nodesInfo = await wampSqlSelectQueryRows([
+      `SELECT ip_address as ipAddress, account_id as accountId, node_id as nodeId, 
+        last_seen as lastSeen, last_height as lastHeight,status,
+        agent_name as agentName, agent_version as agentVersion, agent_build as agentBuild
+              FROM nodes
+              WHERE account_id IN (${accountArray})
```
frol, line 188: You may easily introduce security issues when you construct SQL queries with inline values. Use named replacement here.

backend/src/index.js
```diff
@@ -91,9 +97,151 @@ async function main() {
...
+        });
+      }
+    }
+
+    for (let i = 0; i < nodes.length; i++) {
+      let nodeInfo = nodeMap.get(nodes[i].account_id);
+      if (nodeInfo) {
+        nodes[i].nodeInfo = nodeInfo;
+      }
+      nodes[i].nodeInfo = undefined;
```
frol, line 215:
```suggestion
      nodes[i].nodeInfo = nodeMap.get(nodes[i].account_id);
```

backend/src/index.js
```diff
@@ -91,9 +97,151 @@ async function main() {
...
+          agentName: nodesInfo[i].agentName,
+          agentVersion: nodesInfo[i].agentVersion,
+          agentBuild: nodesInfo[i].agentBuild,
+          nodeId: nodesInfo[i].nodeId,
+          lastHeight: nodesInfo[i].lastHeight,
+          lastSeen: nodesInfo[i].lastSeen,
+          ipAddress: nodesInfo[i].ipAddress,
+          accountId: nodesInfo[i].accountId,
+          status: nodesInfo[i].status,
+        });
```
frol, line 206:
```suggestion
        const { accountId, ...nodeInfo } = nodesInfo[i];
        nodeMap.set(accountId, nodeInfo);
```

icerove:
> It seems that the initial display of the list can take up to 10 seconds since, which is very suboptimal. We need to pull the initial version of the data immediately on page load.

I tried and do not have this problem and I have question here: in function `regularCheckNodeStatus`, we use call method `select` here. If  `regularCheckNodeStatus` is also a call method here, I kind of confusing about this. Will there be some trouble here?
Also, I also find the node page loading burden is sooooo heavy and the nodenav is strange when click to change table. I am thinking of not exposed validators list, will try to separate amount and list first.

frontend/src/pages/_app.jsx
```diff
@@ -74,18 +57,17 @@ export default class extends App {
...
-          <RpcProvider
-            finalStamp={this.state.finalStamp}
-            lastBlockHeight={this.state.lastBlockHeight}
-          >
-            <div className="app-wrapper">
-              <Header />
-              <div className="page">
-                <Component {...pageProps} />
+          <StatsDataProvider>
+            <NodeProvider>
```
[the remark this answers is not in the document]
icerove: I change node provider to other pages, but I still keep stats data one since this provider is  used so many places. what do you think?

backend/src/index.js
icerove: I tried: 
image

but sql query does not deal with it  correctly and gives this:
image
it have a ' ' out of string and incorrect

frontend/src/pages/_app.jsx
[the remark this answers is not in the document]
frol: That sounds reasonable to me :+1:

backend/src/index.js
frol: You should just pass a JS array of account IDs instead of a comma-separated string.

frontend/src/components/dashboard/DashboardHeader.tsx
```diff
@@ -69,168 +60,135 @@ export default class extends React.Component<State> {
...
-        </Row>
+                      <p>
+                        <a href="https://docs.near.org/docs/concepts/overview">
+                          {"Learn more about the key concepts"}
+                        </a>
+                      </p>
+                    </Term>
+                  }
+                  imgLink="/static/images/icon-m-height.svg"
+                  text={context.dashState.lastBlockHeight.toLocaleString()}
```
frol, line 122: Is it going to display "0" when the data has not been loaded yet? We should not do that. We need to display some neutral value (N/A / loading / "..." / "-" / some other loading indication)

frontend/src/components/nodes/NodeNav.tsx
```diff
@@ -3,138 +3,107 @@ import Link from "next/link";
...
-            font-size: 12px;
-            font-weight: 500;
-            letter-spacing: 1.38px;
-            color: #24272a;
-            text-transform: uppercase;
-            padding: 10px;
-            text-decoration: none;
-          }
+      <NodeStatsConsumer>
+        {(context) => (
```
frol, line 17: In some cases, you may consider unpacking the context immediately. I don't have a strong preference, just wanted to mention.

```suggestion
        {({ nodeInfo }) => (
```

frontend/src/context/StatsDataProvider.jsx
```diff
@@ -0,0 +1,123 @@
...
+        lastDayTxCount: action.lastDayTxCount,
+      };
+    default:
+      return stateInit;
+  }
+};
+
+const StatsDataContext = createContext({
+  finalTimestamp: 0,
+  dashState: stateInit,
```
frol, line 52: > The dash is a punctuation mark that is similar in appearance to the hyphen and minus sign 

[Wiki](https://en.wikipedia.org/wiki/Dash)

Let's use `dashboardStats` instead

frontend/src/components/nodes/NodesMap.tsx
```diff
@@ -391,6 +345,7 @@ export default class extends React.Component<State> {
         }}
       />
     );
+    let value = this.context;
```
frol, line 348: This does not make the code any better. Let's either use `const { nodeInfo } = this.context` or just use `this.context.nodeInfo` in place without renaming it to "value"

frontend/src/components/nodes/NodesMap.tsx
```diff
@@ -307,22 +259,24 @@ export default class extends React.Component<State> {
...
         nodesType: "validators",
         newNodes: [],
       },
-      () => this.fetchGeo()
+      () => this.fetchGeo(value.nodeInfo.validators, value.nodeInfo.onlineNodes)
     );
   };
 
   changeToOnlineNodes = () => {
+    let value = this.context;
```
frol, line 273: ^^^

frontend/src/components/nodes/NodesMap.tsx
```diff
@@ -307,22 +259,24 @@ export default class extends React.Component<State> {
   };
 
   changeToValidators = () => {
+    let value = this.context;
```
frol, line 262: ^^^

frontend/src/components/nodes/NodesMap.tsx
```diff
@@ -55,50 +43,16 @@ export default class extends React.Component<State> {
   };
 
   componentDidMount() {
-    this.fetchGeo();
+    let value = this.context;
```
frol, line 46: ^^^

frol:
@icerove It seems that functionality is in place, so make your own judgment when to merge it :+1:

PR #349, opened 2020-07-14
feat(frontend): rework Nodes pages

### LINKED ISSUE #339: Rework Validators page
Validators page should provide info about all validators not just validators that are currently reporting telemetry validators. Telemetry should just augment information from `validators` RPC (accessible via `provider.validators()`).

Additionally, there should be information about epoch progress on the top (similar to main page dashboard), avg. block time, total amount of stake, current stake price.

Also in the list of validators show ones that are going to be kickedout in the next epoch / new coming and also upcoming proposals. This can be shown in the single list indicating with color and icons on the status.

A really nice UI: http://solanabeach.io/

Specifically for epoch progress:
Screen Shot 2020-06-26 at 10 22 22 AM

For listing validators and their stake distribution:
Screen Shot 2020-06-26 at 10 22 57 AM

icerove:
fix #339 
- [x] 1. change  the query method for validators.
- [x] 2. delete the pagination for node page and validator page.
- [x] 3. change page into validators page and online nodes page.
- [x] 4. change dash board node explanation to total validating nodes and total online nodes.
- [x] 5. add proposal table and display kickout nodes table.
- [x] 6. add test for new components


frontend/src/components/nodes/Nodes.tsx
```diff
@@ -2,60 +2,62 @@ import React from "react";
...
+    if (this.props.role === "online-nodes") {
+      let onlineNodes = await this.getNonValidatingNodes();
+      this.setState({ onlineNodes });
+    }
   };
 
   componentDidUpdate(prevProps: any) {
     if (this.props.role !== prevProps.role) {
-      this.autoRefreshNodes = autoRefreshHandler(Nodes, this.config);
+      this.getNodes();
```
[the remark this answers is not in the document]
icerove: I remember we discuss this before and the answer is delete the auto refresh somehow. However, if you re think we need auto refresh, here are two methods:
1. use pub  and lib since I actually make this in the PR 
2. write the auto refresh logic in this component

icerove:
this PR is not finished, why do you approve now? need at least add kickout and proposal table in the node page. @frol

frol:
@icerove I suggest we add those extra pages in separate PRs. WDYT?

frontend/src/components/nodes/Nodes.tsx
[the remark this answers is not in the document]
frol: I find it confusing to see `last seen X minutes ago` if I keep the tab open for a while, so this needs to be resolved one way or another. I leave the implementation detail decision on you. Let me know if you need to brainstorm it (keep in mind that some nodes may go offline and back online, and new ones can get online after an initial page load, etc).

icerove:
I am sorry, I do not read your last comment and already add  proposal table to it.

frontend/src/components/nodes/NodeNav.tsx
```diff
@@ -66,21 +66,66 @@ export default class extends React.Component<Props, State> {
...
+            >
+              <Col className="node-selector align-self-center">
+                {nodeStats
+                  ? `${nodeStats.proposalsCount} Proposal-nodes`
+                  : `- Proposal-nodes`}
+              </Col>
+            </a>
+          </Link>
+        </Row>
+        <Row className="node-icon">
```
frol, line 98: I suggest we just add these hints inline, so the user can just hover over the circle on the left to the node id and read the hint.

frontend/src/components/nodes/Nodes.tsx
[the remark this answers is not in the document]
frol: @icerove this really needs to be addressed ^^^. Let me know if you want to do this in a separate PR or whatnot.

frontend/src/components/nodes/Nodes.tsx
```diff
@@ -1,61 +1,101 @@
...
+    }
+    if (onlineNodes && this.props.role === "online-nodes") {
+      nodes = onlineNodes.map((node: N.NodeInfo) => (
+        <NodeRow key={node.nodeId} node={node} />
+      ));
+    }
+    if (proposals && this.props.role === "proposals") {
+      nodes = proposals.map((node: N.Proposal) => (
+        <ProposalRow key={node.account_id} node={node} />
+      ));
```
frol, line 94: BTW, you have re-invented routing in this component :man_shrugging: 

It might make more sense to have different components and even different pages (I mean different files in `pages/nodes/` folder instead of `[role].jsx`)

frontend/src/components/nodes/Nodes.tsx
```diff
@@ -1,61 +1,101 @@
...
+      let nextValidators = nodes.next_validators;
+      let diffEpochValidators = VA.diffEpochValidators(
+        currentValidators,
+        nextValidators
+      );
+      console.log(diffEpochValidators);
+      let newValidators = this.signNewValidators(
+        diffEpochValidators.newValidators
+      );
+      this.signRemovedValidators(diffEpochValidators.removedValidators);
```
frol, line 54: From the call-sight, it seems that these two methods operate differently while they are almost identical. If you need to change the objects in place, I suggest you do that instead of making an illusion that a new array is returned from `signNewValidators`.

```suggestion
      const { newValidators, removedValidators } = nearApi.validators.diffEpochValidators(
        currentValidators,
        nextValidators
      );
      this.signNewValidators(newValidators);
      this.signRemovedValidators(removedValidators);
```

frontend/src/components/nodes/Nodes.tsx
icerove: will make them different page

icerove:
for auto refresh nodes, I want to make a separate PR to work.

frontend/src/components/nodes/Nodes.tsx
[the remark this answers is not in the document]
icerove: in a seperate PR

frontend/src/components/nodes/Proposals.tsx
```diff
@@ -0,0 +1,37 @@
...
+    let nodes;
+    if (proposals) {
+      nodes = proposals.map((node: N.Proposal) => (
+        <ProposalRow key={node.account_id} node={node} />
+      ));
+    }
+    if (!nodes) {
+      return <PaginationSpinner hidden={false} />;
+    }
+    return <>{nodes}</>;
```
frol, line 35: This can be simplified to (please, update the other components accordingly):

```suggestion
    if (!proposals) {
      return <PaginationSpinner hidden={false} />;
    }
    return <>
      {proposals.map((node: N.Proposal) => (
        <ProposalRow key={node.account_id} node={node} />
      ))}
    </>;
```

frontend/src/components/nodes/ValidatorRow.tsx
```diff
@@ -0,0 +1,235 @@
...
+      <RpcConsumer>
+        {(context) => (
+          <Row className="node-row mx-0">
+            <Col md="auto" xs="1" className="pr-0">
+              {node.new ? (
+                <OverlayTrigger
+                  placement={"right"}
+                  overlay={
+                    <Tooltip id="new">
+                      next epoch upcoming validating nodes
```
frol, line 55: @kcole16 may I ask you to bikeshed the wording in the tooltips (search for `<Tooltip`) here? They are visible on the ["Validation Nodes" page](https://near-explorer-frontend-pr-349.onrender.com/nodes/validators) when you hover over the green (and other colors) dots on the left from the node ID:

image

frontend/src/libraries/explorer-wamp/nodes.ts
```diff
@@ -15,72 +15,68 @@ export interface NodeInfo {
...
+  async getOnlineNodes() {
     try {
       let nodes = await this.call<NodeInfo[]>("select", [
         `SELECT ip_address as ipAddress, moniker, account_id as accountId, node_id as nodeId, signature, 
           last_seen as lastSeen, last_height as lastHeight, last_hash as lastHash,
           agent_name as agentName, agent_version as agentVersion, agent_build as agentBuild,
           peer_count as peerCount, is_validator as isValidator, status
               FROM nodes
-              ${whereClause}
+             WHERE last_seen > (strftime('%s','now') - 60) * 1000
```
frol, line 53:
```suggestion
              WHERE last_seen > (strftime('%s','now') - 60) * 1000
```

PR #362, opened 2020-07-21
set up access to postgres

icerove:
1. not disclose our password and not hard code the config.
2. leave the test part in the src/index.js
3. the start command should be `env NEAR_NETWORK_INDEXER_EXPLORER_PASSWORD=password npm start`


backend/config/database.js
```diff
@@ -14,4 +16,11 @@ module.exports = {
     dialect: "sqlite",
     storage: `db/${dbPrefix || "production"}-database.sqlite`,
   },
+  postgres: {
+    username: "explorer",
+    password: dbPassword,
+    database: "explorer_testnet",
+    host: "postgres.render.com",
+    dialect: "postgres",
+  },
```
frol, line 25: I suggest the following refactoring:

```js
const dbPrefix =
  process.env.NEAR_DB_PREFIX || process.env.WAMP_NEAR_NETWORK_NAME || "development";

module.exports = {
  sqlite: {
    dialect: "sqlite",
    storage: `db/${dbPrefix}-database.sqlite`,
  },
  postgres: {
    username: "explorer",
    password: dbPassword,
    database: `explorer_${dbPrefix}`,
    host: "postgres.render.com",
    dialect: "postgres",
  },
};
```

backend/models/index.js
```diff
@@ -8,6 +8,8 @@ const env = process.env.NODE_ENV || "development";
 const config = require(__dirname + "/../config/database")[env];
 const db = {};
 
+const postgresConfig = require(__dirname + "/../config/database")["postgres"];
+
```
frol, line 12:
```suggestion
const dbConfig = require(__dirname + "/../config/database");
const db = {};

```

Then, use `dbConfig.sqlite` and `dbConfig.postgres` where applicable.

backend/models/index.js
```diff
@@ -18,6 +20,19 @@ const sequelize = new Sequelize(
...
+const sequelizePostgres = new Sequelize(
+  postgresConfig.database,
+  postgresConfig.username,
+  postgresConfig.password,
+  {
+    host: postgresConfig.host,
+    dialect: postgresConfig.dialect,
+    dialectOptions: {
+      ssl: true,
+    },
```
frol, line 32: Is there a way to enforce read-only mode on connection? We should not let people to alter our databases via WAMP calls, you know :smile:

frol: Just for the track, we decided to enforce that with postgres read-only access user

PR #363, opened 2020-07-24
Remove occurences of devnet since it's being deprecated

janedegtiareva:


icerove:
I already remove it from server, I want to know your opinion about delete it from code. @frol

frol:
Yes, we don't need them anymore if devnet is deprecated. Good catch!

PR #369, opened 2020-07-30
fix: Fixed pagination of transactions in a single block

icerove:
the transaction query failure is from all the transactions have same block_timestamp which we use to indicate the pagination In the future, this maybe more easily to happen and I suggest just like what we do to accounts table, add one more index column and query both

### DIFF
--- a/backend/models/Transaction.js
+++ b/backend/models/Transaction.js
@@ -16,6 +16,10 @@ module.exports = (sequelize, DataTypes) => {
         type: DataTypes.BIGINT,
         allowNull: false,
       },
+      transactionIndex: {
+        type: DataTypes.BIGINT,
+        allowNull: false,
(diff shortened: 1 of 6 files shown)


icerove:
> NOTE: We will need to re-sync the data on production once we lend this PR.

Do you mean delete the render backend disk? @frol

frol:
@icerove It would be better to rename instead of removing completely, so yes:

1. Go to Render.com and open shell
2. `mv db/testnet-database.sqlite db/testnet-database.sqlite.2020-07-30` (and do the same for mainnet and betanet). FYI, It is safe operation, and will not break the current deployment as those files remain to be open by the backend processes, and file renaming does not change anything for them, they keep working with the old file.
3. Merge this PR

This should be it.

PR #374, opened 2020-08-03
fix(frontend): Clarified the documentation message for the locked state of a contract

frol:
This is a quite techy explanation, so feel free to suggest a better one


kcole16:
```
Note: The contract itself may be implemented to re-deploy itself or re-add FullAccess keys. 
```

PR #378, opened 2020-08-06
fix: Change the way GeoIP is handled for node map

### LINKED ISSUE #368: Node map is not displaying online validating node directly
Go to https://explorer.testnet.near.org/nodes/map and observe an empty map
Users have to click the button to see the node on the map

Expected: not empty map on first glance

icerove:
fix #368 

definitely sure about ip geolocation establishment. not very sure about the UI. 
when merge this, we will need to rename database on render.com again


frontend/src/pages/api/nodes.js
```diff
@@ -1,16 +1,54 @@
+import IPData from "ipdata";
 import { ExplorerApi } from "../../libraries/explorer-wamp";
 
+const ipdata = new IPData(
+  "c8b80db74e687b7471135f099c0332c36899f40df0a2716e7a66d57e"
```
frol, line 5: Just for future reference, NEVER commit code with real credentials (even if you believe that there is no harm can be made), https://medium.com/@selvaganesh93/what-happens-if-you-accidentally-commit-your-aws-access-token-to-public-github-be50d378b4c7

The very first moment you introduce some key, create a config for it immediately, and gitignore it.

frontend/src/components/utils/DatamapsExtension.tsx
```diff
@@ -33,9 +29,10 @@ export default class Datamap extends React.Component<Props> {
     this.drawMap();
   }
 
-  componentDidUpdate() {
-    this.clear();
-    this.drawMap();
+  componentDidUpdate(prevProps: Props) {
+    if (this.props.bubbles.length !== prevProps.bubbles.length) {
+      this.drawMap();
+    }
```
frol, line 35: 1. If you don't want to re-render, it makes sense to implement `shouldComponentUpdate`
2. Are you sure it is safe to draw without clearing and also skip some drawing if the length of the arrays is the same? What if one node goes offline and another one goes online (the length will be the same as far as I can tell)?

frontend/src/libraries/explorer-wamp/nodes.ts
```diff
@@ -10,6 +8,9 @@ export interface NodeInfo {
   agentVersion: string;
   agentBuild: string;
   status: string;
+  latitude: number;
+  longitude: number;
```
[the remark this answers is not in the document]
icerove: it could not represent as string, otherwise the map could not location their latitude and longitude.

frontend/src/pages/api/nodes.js
icerove: got it

frontend/src/components/utils/DatamapsExtension.tsx
icerove: not very sure, need your help here. @frol

chain.json
```diff
@@ -0,0 +1,9 @@
+{
```
frol, line 1: Is it a debug-only helper? Should it be removed?

agent.json
```diff
@@ -0,0 +1,5 @@
+{
```
frol, line 1: Is it a debug-only helper? Should it be removed?

frontend/src/components/nodes/NodeNav.tsx
```diff
@@ -26,7 +26,7 @@ export default class extends React.Component<Props> {
                 >
                   <Col className="node-selector align-self-center">
                     {context.nodeInfo.validatorAmount
-                      ? `${context.nodeInfo.validatorAmount}  Validating`
+                      ? `${context.nodeInfo.validatorAmount}  Validating & New Upcoming`
                       : `- Validating`}
```
frol, line 30: Here you got bitten by the code duplication. Your messages are now different during the loading and the loaded states.

frontend/src/components/nodes/NodesMap.tsx
```diff
@@ -36,83 +33,38 @@ interface State {
 class NodesMap extends React.Component<State> {
   state: State = {
     nodesData: [],
-    nodesType: "validators",
+    nodesType: "",
```
frol, line 36: 1. Why?
2. What is the semantics of "" node type? It looks cryptic to me

frontend/src/components/nodes/NodesMap.tsx
```diff
@@ -36,83 +33,38 @@ interface State {
...
+      this.state.nodesType === "validators" ? validators : nodeInfo.onlineNodes;
+
+    const bubbles: IBubble[] = nodes.map((node: N.NodeInfo & N.Validating) => {
+      return {
+        radius: 4,
+        fillKey: "validatorBubbleFill",
+        nodeInfo: node,
+        nodeId: node.nodeId || node.nodeInfo?.nodeId || "Unknown",
+        latitude: node.latitude || node.nodeInfo?.latitude,
+        longitude: node.longitude || node.nodeInfo?.longitude,
```
frol, line 63: Can we improve the data layout so we don't need to do all these hops and dances?

frontend/src/components/nodes/NodesMap.tsx
```diff
@@ -174,36 +126,44 @@ class NodesMap extends React.Component<State> {
     // Prettier ruined the entire indentation that i made for this
     return `<div className="hoverinfo" style="border: none; text-align: left; padding: 20px 0px 0px; box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.16); border-radius: 8px; color: white; background-color: #343A40; max-width: 300px">
         <div style="color: #8DD4BD; font-size: 14px; line-height: 14px; letter-spacing: 0.4px; font-weight: bold; font-family: BwSeidoRound; padding:0 20px 8px"> @${
-          data.nodeInfo.accountId
+          data.nodeInfo.account_id || data.nodeInfo.accountId
```
frol, line 129: Why did we introduce this inconsistency? Can we avoid it?

frontend/src/components/utils/DatamapsExtension.tsx
```diff
@@ -33,8 +29,15 @@ export default class Datamap extends React.Component<Props> {
     this.drawMap();
   }
 
+  shouldComponentUpdate(nextProps: Props) {
+    if (nextProps.pins.length !== this.props.pins.length) {
```
frol, line 33: Let's do deep comparison here just to be on the safe side of things. Use https://www.npmjs.com/package/fast-deep-equal (we already depend on it through other dependencies, so it will not introduce extra code)

frontend/src/pages/api/nodes.js
```diff
@@ -2,15 +2,23 @@ import { ExplorerApi } from "../../libraries/explorer-wamp";
 
 export default async function (req, res) {
   try {
-    await new ExplorerApi(req).call("node-telemetry", [
+    let ip_address =
+      req.headers["x-forwarded-for"] ||
+      req.connection.remoteAddress ||
+      req.socket.remoteAddress ||
+      req.connection.socket.remoteAddress;
```
frol, line 9: Are those different?

frontend/src/pages/api/nodes.js
```diff
@@ -2,15 +2,23 @@ import { ExplorerApi } from "../../libraries/explorer-wamp";
...
+
+    res.send("node telemetry received.");
+
+    return await new ExplorerApi(req).call("node-telemetry", [
```
frol, line 13: I don't see a point in returning anything. HTTP 200 with blank response should be enough.

frontend/src/components/nodes/NodesMap.tsx
icerove: if we change here to the same format, we will change backend with this to match the same format.

icerove: see here: https://github.com/near/near-explorer/blob/7356915e75d9c937c8fe49335ef1f26456bcc6d7/backend/src/db-utils.js#L63 
the online node and validating node have different attrs, so it is different format.

frontend/src/components/nodes/NodesMap.tsx
icerove: ^^^

frontend/src/pages/api/nodes.js
icerove: I search and find this for some browser may cause problem if we do not have all these. this will make sure all browser can give correct location.

frontend/src/components/nodes/NodesMap.tsx
icerove: because the validators or onlinenodes data will fetching for a while and it will be weird  if nothing happen when the validators are selected. But then you click each of them instead, it will be correct to display.

frontend/src/components/nodes/NodeNav.tsx
```diff
@@ -26,8 +26,8 @@ export default class extends React.Component<Props> {
                 >
                   <Col className="node-selector align-self-center">
                     {context.nodeInfo.validatorAmount
-                      ? `${context.nodeInfo.validatorAmount}  Validating`
-                      : `- Validating`}
+                      ? `${context.nodeInfo.validatorAmount}  Validating & New Upcoming`
+                      : `- Validating & New Upcoming`}
```
frol, line 30: Here is how you can avoid such bugs from code duplication:

```js
{`${context.nodeInfo.validatorAmount ?? '-'} Validating & New Upcoming`}
```

Read more about ["#3: Nullish Coalescing"](https://www.freecodecamp.org/news/javascript-new-features-es2020/)

icerove: thanks!

frontend/src/components/nodes/NodesMap.tsx
icerove: sorry to say this, but I tried to add componentDidUpdate and shoudComponentUpdate method here, but it still load too much data and can not be control. context component gives data first time is empty and componentdidmount is over. After only click button will trigger setstate method to new data. @frol

frol: Well, I don't have time to dive into it this week, so I suggest we merge it as is now and address it in the next iteration.

PR #385, opened 2020-08-12
fix(frontend): Display transaction fee tooltip when amount is larger than 0.00001 N

### LINKED ISSUE #383: Transaction fee popup should not be empty
https://explorer.near.org/transactions/7aZRyvfRnZzMQtxmxvMnTGmrdR44osAnRhady5ZE39EY

Hover over the transaction fee and observe that currently it is empty.

image

icerove:
fix #383 
also add mainnet script in the package.json in frontend


frontend/src/components/utils/Balance.jsx
```diff
@@ -43,6 +43,7 @@ export const showInYocto = (amountStr) => {
   if (amount.lte(YOCTO_NEAR_THRESHOLD)) {
     return formatWithCommas(amountStr) + " yoctoⓃ";
   }
+  return formatWithCommas(amountStr) + " yoctoⓃ";
```
frol, line 46: Why do we need the condition? It seems that the code is the same

PR #387, opened 2020-08-18
fix(frontend): Dashboard page should not do an overwhelming amount of WAMP requests

### LINKED ISSUE #386: Dashboard page should not do an overwhelming amount of WAMP requests
## Story

Currently, when I open the Dashboard and monitor the `/ws` endpoint I see that it sends a ton of requests all the time! There is no need to do that.

## Acceptance Criteria

- [ ] The amount of requests is sane (~10 requests on load is fine and at most 2 requests per second going forward)
- [ ] Tested

icerove:
fix #386


frol:
@icerove  I don't see any improvement:

image

icerove:
check this again @frol 
I delete componentdidupdate on the dashboard for transactions and blocks and the overwhelming problem is solved. But I found there is always two request send at same time, one is what we send like total blocks, another is a long number and I do not know the meaning, I think it is the wamp format or some thing similar?

frol:
Recent blocks and transactions now don't receive any new items after the initial render (autorefresh is dead).

> another is a long number and I do not know the meaning, I think it is the wamp format or some thing similar?

yes, that is the case.

icerove:
check again @frol

frontend/src/context/DatabaseProvider.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+const DatabaseContext = createContext({
+  finalTimestamp: 0,
+  lastBlockHeight: 0,
+  totalBlocks: 0,
+  totalTransactions: 0,
+  totalAccounts: 0,
+  lastDayTxCount: 0,
+  transactions: [],
+  blocks: [],
```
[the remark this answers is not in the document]
icerove: In that case, I will create a new provider only for those two lists.

frontend/src/components/dashboard/DashboardTransactions.tsx
```diff
@@ -4,7 +4,7 @@ import React from "react";
 import { Row, Col } from "react-bootstrap";
 
 import * as T from "../../libraries/explorer-wamp/transactions";
-import { DatabaseConsumer } from "../../context/DatabaseProvider";
+import { ListConsumer } from "../../context/ListProvider";
```
frol, line 7: List of what? Please-please-please, give descriptive names. `DashboardDataProvider` or such is much easier to reason about.

backend/src/index.js
```diff
@@ -129,11 +129,8 @@ async function main() {
       if (wamp.session) {
         const dataStats = await aggregateStats();
         const { transactions, blocks } = await queryDashboardBlocksAndTxs();
-        wampPublish(
-          "database-stats",
-          [{ dataStats, transactions, blocks }],
-          wamp
-        );
+        wampPublish("database-stats", [{ dataStats }], wamp);
+        wampPublish("database-list", [{ transactions, blocks }], wamp);
```
frol, line 133: Please, spend more time on naming things as it is hard to review and maintain.

database stats is somewhat meaningful, but database list has no reasonable meaning to me. I still suggest considering the high-level picture and see that `database-stats` is hard to reason about while `network-stats` or `chain-stats` are more specific, and later we may add `economics-stats`.

P.S. Whenever you change semantics, make sure that the parent function/class/module names are still describing the implementation accurately.

frontend/src/context/ListProvider.jsx
```diff
@@ -0,0 +1,35 @@
...
+  blocks: [],
+});
+
+export default (props) => {
+  const [transactions, dispatchTransactions] = useState([]);
+  const [blocks, dispatchBlocks] = useState([]);
+
+  const fetchList = function (stats) {
+    let transactions = stats[0].transactions;
+    let blocks = stats[0].blocks;
```
frol, line 15:
```suggestion
    const { blocks, transactions } = stats[0];
```

frontend/src/context/ListProvider.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+    dispatchBlocks(blocks);
+    dispatchTransactions(transactions);
+  };
+
+  const Subscription = useCallback(() => {
+    new ExplorerApi().subscribe("database-list", fetchList);
```
frol, line 22: We need to unsubscribe from the topic when the provider gets unmounted, otherwise, it will keep receiving the traffic.

PR #398, opened 2020-08-25
fix unsubscribe issue 

icerove:
I find current one has some issues, let's revert unsubscribe part and I will create new PR for this.


frontend/src/context/ListProvider.jsx
```diff
@@ -9,6 +9,7 @@ const ListContext = createContext({
 export default (props) => {
   const [transactions, dispatchTransactions] = useState([]);
   const [blocks, dispatchBlocks] = useState([]);
+  let sub;
```
frol, line 12: What is this for?

PR #399, opened 2020-08-25
feat(backend): Migrate away from EXPERIMENTAL_genesis_records RPC API to raw genesis.json file

### LINKED ISSUE #393: Migrate away from EXPERIMENTAL_genesis_records
nearcore wants to shrink down the RAM usage and genesis records are huge to keep in memory all the time, so nearcore wants to deprecate the `EXPERIMENTAL_genesis_records`. Instead of getting the genesis records from the RPC we can download the genesis.json file for the relevant network from S3 and load the records from it.

* https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore-deploy/betanet/genesis.json
* https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore-deploy/testnet/genesis.json
* https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore-deploy/mainnet/genesis.json (not available at the moment, but I believe it will be /cc @bowenwang1996)

Notes:
* The genesis file can be gigabytes of JSON, and that may introduce some issues due to the RAM usage (just be ready). Ideally, we want to have a way to scan it through and iterate over the records without reading all of them into memory all at once
* There can be the case that there is no genesis.json for some networks available on our S3 (e.g. when we run a local node network), Explorer backend should just log a warning and keep going

icerove:
fix #393

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,13 +1,13 @@
 version: '2'
 
 services:
-  # wamp:
-  #   build: ./wamp
-  #   environment:
-  #     - WAMP_NEAR_EXPLORER_PORT=8080
-  #     - WAMP_NEAR_EXPLORER_BACKEND_SECRET=back
-  #   ports:
-  #     - 8080:8080
+  wamp:
+    build: ./wamp
+    environment:
+      - WAMP_NEAR_EXPLORER_PORT=8080
+      - WAMP_NEAR_EXPLORER_BACKEND_SECRET=back
+    ports:
+      - 8080:8080
 
   backend:
     build: ./backend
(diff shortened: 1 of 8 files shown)


ailisp:
Is that possible nodejs garbage collection isn't triggered? 2.5G is still a small proportion of your total ram, have a try on config nodejs gc options / or run it in a docker with limited memory: https://docs.docker.com/config/containers/resource_constraints/#limit-a-containers-access-to-memory

icerove:
give docker  memory limit to 1000M and test with original code, it works fine with testnet and betanet. I think it is your computer do not trigger  garbage collection with 2.5G memory since it is still enough to work. But when memory is small, nodejs will frequently garbage collect.

icerove:
Try limit memory for docker and test with docker image. Unfortunately, this stream json method is not working as it said using small memory to stream. The implementation works, test with docker is recommended. @frol

icerove:
I already change to new library and do you think this is the wamp insert issue? The streaming json might be true but insertion is not controlled as expected? @frol

docker-compose.yml
```diff
@@ -14,14 +14,16 @@ services:
...
-    ports:
-      - 3000:3000
+  # frontend:
+  #   build: ./frontend
+  #   environment:
+  #     - PORT=3000
+  #     - WAMP_NEAR_EXPLORER_URL=ws://localhost:8080/ws
+  #     - WAMP_NEAR_EXPLORER_INTERNAL_URL=ws://wamp:8080/ws
+  #   ports:
+  #     - 3000:3000
```
frol, line 29:
```suggestion
  frontend:
    build: ./frontend
    environment:
      - PORT=3000
      - WAMP_NEAR_EXPLORER_URL=ws://localhost:8080/ws
      - WAMP_NEAR_EXPLORER_INTERNAL_URL=ws://wamp:8080/ws
    ports:
      - 3000:3000
```

icerove: I am still figure out memory problem and after all the fixes, I will return docker file to original one.

icerove:
I mean sequelize, sorry @frol

icerove:
test this version with docker, I tried 500M and out of memory but 1G is enough. @frol

icerove:
I will keep this PR not merged right now, since mainnet has different json file path and cannot work as testnet and betanet

frol:
@icerove Please, revert the irrelevant changes to the docker-compose.yml file

PR #412, opened 2020-09-04
Fix/genesis query sequence

icerove:
I see the new instance for testnet memory usage is under 1G, and always under 1G. Thinking nodejs assign these several nets together maybe too heavy to load. I give sequence for every different net (mainnet is the first and then betanet and then guildnet and last one is testnet) and each one having 5 minutes to load genesis, since despite testnet genesis, every one will be really quick to finish genesis querying.

### DIFF
--- a/backend/src/config.js
+++ b/backend/src/config.js
@@ -47,3 +47,8 @@ exports.wampNearExplorerBackendSecret =
   process.env.WAMP_NEAR_EXPLORER_BACKEND_SECRET || "back";
 
 exports.genesisRecordsUrl = process.env.NEAR_GENESIS_RECORDS_URL;
+
+exports.genesisQuerySequence = process.env.NEAR_GENESIS_QUERY_SEQUENCE || 0;
+
+exports.genesisWaitingTime =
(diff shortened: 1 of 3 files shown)


frol (changes requested):
I don't think it will help. I observe that testnet on a dedicated 1GB of RAM limit hits the limit and dies, so when I bumped it to 2GB (only testnet is running on that instance), it managed to sync, but during the process it is consuming up to 2.5GB of RAM (I don't know how that is allowed by Render, but that is what they report on the metrics page).

We can just give it up and use 2GB instance for testnet for now, but the problem is that every push to master brings testnet down by at least 9 minutes since the genesis sync routine is so tight that almost no other requests are handled during this time. We should add some delays between batches, so other jobs inside the backend have a chance to do their work OR we should add a way to check that we already have the genesis fully synced, and thus we don't need to download it and insert again (create a new model/table GenesisInfo with `genesis_time`, `genesis_height`, and `chain_id` columns; we are going to only add a single record to that table when we finish syncing genesis records and this will be the indicator that we don't need to download and sync genesis again when we restart the backend)

PR #414, opened 2020-09-08
Refactor/give delay query genesis

icerove:
fix genesis sync

### DIFF
--- a/backend/models/Genesis.js
+++ b/backend/models/Genesis.js
@@ -1,0 +1,30 @@
+"use strict";
+module.exports = (sequelize, DataTypes) => {
+  const Genesis = sequelize.define(
+    "Genesis",
+    {
+      genesisTime: {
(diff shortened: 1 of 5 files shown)


frol (approved):
I have slightly refactored the code to be more robust. Please, test it on your end.

icerove:
I think there is some issue left, when there is genesis row is fetched(means the genesis records already fetched) the backend log will be this:
image
and exit.
Is this what expected? 
I think it is not really correct since you see the height and time is exactly the same but backend think it is different and reset.

frol:
> Is this what expected?

@icerove It is clearly not. Do you use an old DB file or a fresh one? I had changed the model to use BIGINT for the timestamp, so it might cause an issue on your end. If it is not the case, please, print the `typeof` of those values we compare and let's see how to deal with that.

PR #415, opened 2020-09-08
change gas display

### LINKED ISSUE #413: Confusing gas unit on transaction page
On the transaction page for this transaction https://explorer.mainnet.near.org/transactions/BsocFPKcyeGgzAVahQSWwuNf5pp5M3RVtfohXhJGv8PR, it shows that gas used is 460052 Mgas, which seems a bit confusing as I would expect to see 460 Ggas here. As an analogy, while technically correct, we would say that our total supply is 1 billion instead of 1000 million.

icerove:
fix #413

### DIFF
--- a/frontend/src/components/utils/Gas.tsx
+++ b/frontend/src/components/utils/Gas.tsx
@@ -5,7 +5,7 @@ interface Props {
 }
 
 const MGAS = new BN(10 ** 6);
-const GGAS = new BN(10 ** 9);
+const GGAS = new BN(1024).mul(MGAS);
 export const TGAS = new BN(10 ** 12);
 
 export default ({ gas }: Props) => {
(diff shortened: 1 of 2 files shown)


frontend/src/components/utils/Gas.tsx
```diff
@@ -5,12 +5,12 @@ interface Props {
 }
 
 const MGAS = new BN(10 ** 6);
-export const TGAS = new BN(10 ** 12);
+export const GGAS = new BN(10 ** 9);
```
bowenwang1996, line 8: We should still have TGAS

frol (approved):

frontend/src/components/utils/Gas.tsx
```diff
@@ -5,12 +5,15 @@ interface Props {
 }
 
 const MGAS = new BN(10 ** 6);
+const GGAS = new BN(1024).mul(MGAS);
```
frol, line 8: Why 1024?

I like the idea of defining the next const using the previous one multiplied by 1000.

frol (changes requested):
^

frol (approved):

PR #420, opened 2020-09-14
have vote function

### LINKED ISSUE #416: Display Phase 2 vote current status
### Overview

To indicate to community members the progress of the vote towards Phase 2, we can visualize the current status of the vote in a progress bar.

The progress can be represented as a % of the vote needed to enable Phase 2. 

**Example**
- Total stake = 600M
- Total votes = 200M (1/3 of total stake)
- Amount needed to enable Phase 2 (2/3 of total stake) = 400M

```
Current Status of the Vote to Enable Phase 2 (2/3 of Total Stake is Needed)

|---------->                      |              
        200M Voted           400M Needed (2/3)              

```

**Technical Implementation**

Two functionCalls to the [voting contract](https://github.com/near/core-contracts/blob/cd221798a77d646d5f1200910d45326d11951732/voting/src/lib.rs)
- Total stake = `get_total_voted_stake()`
- Total votes = `get_votes()`

Calculations
- Total votes needed = `total_stake * (2/3)`
- % of total stake currently voted = `(total_votes / total_stake) * 100`

### Acceptance Criteria

- [x] Total stake (in NEAR) is displayed
- [x] Total votes (in NEAR) is displayed
- [x] Total votes needed (in NEAR) is displayed (2/3 of total stake)
- [x] % of the votes needed to launch Phase 2 is visualized in the progress bar
- [x] Mobile ok
- [x] Tested

icerove:
fix #416


backend/src/near.js
```diff
@@ -41,6 +41,35 @@ const getCurrentNodes = (nodes) => {
...
 
+const testnetConfig = {
+  nodeUrl: "https://rpc.testnet.nearprotocol.com",
+  networkId: "testnet",
+  contractName: "transfer-vote.near",
+};
+const mainnetConfig = {
+  nodeUrl: "https://rpc.mainnet.nearprotocol.com",
+  networkId: "mainnet",
+  contractName: "vote.f863973.m0",
```
bowenwang1996, line 52: you had it backwards. For mainnet the contractName is `transfer-vote.near` whereas for testnet it is `vote.f863973.m0`. Can you test that with the config you put down, you can successfully query the contract state?

frontend/package.json
```diff
@@ -40,6 +48,7 @@
   "dependencies": {
     "@types/enzyme": "^3.10.3",
     "autobahn": "19.7.3",
+    "big.js": "^5.2.2",
```
frol, line 51: Can we `bn.js` instead of pulling another package (`big.js`)? (we already use `bn.js` in many other places)

frontend/src/context/DatabaseProvider.jsx
```diff
@@ -17,10 +19,16 @@ export default (props) => {
   const [totalTransactions, dispatchTotalTransactions] = useState(0);
   const [totalAccounts, dispatchTotalAccounts] = useState(0);
   const [lastDayTxCount, dispatchLastDayTxCount] = useState(0);
+  const [phase2TotalStake, dispatchPhase2TotalStake] = useState(
+    "11416584959186517242479124953463"
+  );
+  const [phase2TotalVotes, dispatchPhase2TotalVotes] = useState(
+    "1272506343149920375853886366252"
```
frol, line 26: What are these magic numbers?

frol (approved):
@icerove I had refactored the backend implementation, but I have not tested it at all. Please, check if that it works end-to-end given the renamings I had done.

/cc @kcole16

corwinharrell:
Screen Shot 2020-09-24 at 10 19 41 AM

Looking ok so far. Can we make the following changes the margins and padding of the different elements to achieve something closer to the mockup?

- Increase left and right padding of the container
- Increase margin between the progress bar and the percentage
- Add a space between the labels and NEAR values
- Decrease the left and right padding of the counter in the upper right.
- Make sure the counter is aligned to the right side of the container.

I'd also like to propose the following change. Rather than saying `17%/100%` and displaying this percentage as a separate measurement from the NEAR amount in the upper right, let's just display the percentage in parenthesis to the right of the counter like so:

Screen Shot 2020-09-24 at 10 42 34 AM

frontend/package.json
icerove: Sorry we cannot(pity), bn.js cannot show decimal but integer. @frol

frontend/src/context/DatabaseProvider.jsx
icerove: it is from testnet vote contract, I use them as testing and later I think it is okay to be initial state.

frol: I would prefer either null or zero values as initial

frontend/package.json
frol: Do we really need decimal? We seem to only care about yoctoNEAR (integer) up until we display the values (millions of NEAR?)

icerove: we display the decimal for Votes/Stake, it is now 0.17%, the bn.js will give 0 which is really weird...

frontend/src/context/DatabaseProvider.jsx
icerove: okay

frontend/package.json
frol: I would compute the percent value x100, then convert the result to regular JS Number (given it cannot be higher than 100%, the max x100 value would be 10000), divide by 100 and use toFixed(2).

Proceed at your own judgement here

icerove: let me have a try

frontend/src/context/DatabaseProvider.jsx
icerove: after re thinking, it could not be null or zero from start

frol:
Is there a link where users can learn more about the phase 2 voting if they are out of context?

ilblackdragon:
@frol https://near.org/blog/mainnet-phase-i/

there is no anchor but "The vote to launch Phase II"

also https://near.org/blog/mainnet-roadmap/

frol:
@icerove let's add the link to the banner (I see that the design expected "Phase 2" to be a link)

PR #435, opened 2020-09-29
feat(frontend): shorten displayed long name

### LINKED ISSUE #427: Design display for long account IDs (e.g. implicit accounts)
## Story

In some places, we don't have enough space to display such long account ids:

image

image

icerove:
fix #427

### DIFF
--- a/frontend/src/components/accounts/AccountRow.tsx
+++ b/frontend/src/components/accounts/AccountRow.tsx
@@ -34,6 +34,10 @@ export default class extends React.Component<Props, State> {
   render() {
     const { accountId, createdAt } = this.props;
     const { amount } = this.state;
+    let Id =
+      accountId.length > 25
+        ? accountId.slice(0, 8) + "..." + accountId.slice(accountId.length - 7)
+        : accountId;
     return (
       <Link href="/accounts/[id]" as={`/accounts/${accountId}`}>
         <a style={{ textDecoration: "none" }}>
@@ -46,7 +50,7 @@ export default class extends React.Component<Props, State> {
             </Col>
             <Col md="7" xs="6">
               <Row>
-                <Col className="transaction-row-title">@{accountId}</Col>
+                <Col className="transaction-row-title">{Id}</Col>
               </Row>
             </Col>
             <Col md="3" xs="4" className="ml-auto text-right">
--- a/frontend/src/components/accounts/AccountRow.tsx
+++ b/frontend/src/components/accounts/AccountRow.tsx
@@ -8,7 +8,7 @@ import { Row, Col } from "react-bootstrap";
 import AccountsApi from "../../libraries/explorer-wamp/accounts";
 
 import Balance from "../utils/Balance";
-import { shortAccountId } from "../utils/AccountLink";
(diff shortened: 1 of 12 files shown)


frontend/src/components/accounts/AccountRow.tsx
```diff
@@ -34,6 +34,10 @@ export default class extends React.Component<Props, State> {
   render() {
     const { accountId, createdAt } = this.props;
     const { amount } = this.state;
+    let Id =
+      accountId.length > 25
+        ? accountId.slice(0, 8) + "..." + accountId.slice(accountId.length - 7)
+        : accountId;
```
frol, line 40: Let's have it as a helper function so we can use `{shortAccountId(accountId)}` in-place.

frontend/src/components/utils/AccountLink.tsx
```diff
@@ -2,11 +2,16 @@ import Link from "next/link";
...
-export default ({ accountId, children }: Props) => (
-  <Link href="/accounts/[id]" as={`/accounts/${accountId}`}>
-    <a className="account-link">{children || `@${accountId}`}</a>
-  </Link>
-);
+export default ({ accountId }: Props) => {
+  let Id =
+    accountId.length > 25
+      ? accountId.slice(0, 8) + "..." + accountId.slice(accountId.length - 7)
+      : accountId;
```
frol, line 11: Reuse the same helper extracted from AccountRow

frol (changes requested):
Once you extract the helper, cover it with a test (feed it with short and long account ids)

frontend/src/components/accounts/AccountRow.tsx
```diff
@@ -34,6 +35,7 @@ export default class extends React.Component<Props, State> {
   render() {
     const { accountId, createdAt } = this.props;
     const { amount } = this.state;
+    let Id = accountId.length > 25 ? shortAccountId(accountId) : accountId;
```
frol, line 38: Let's move the check whether the accountId is too long inside the `shortAccountId` helper (let's have an optional second argument `lengthThreshold = 25`). I want to use it as:

```jsx
<Col className="transaction-row-title">{shortAccountId(accountId)}</Col>
```

P.S. `shortAccountId` is not a perfect naminng, so if you have a better option, please suggest

frontend/src/components/utils/AccountLink.tsx
```diff
@@ -2,11 +2,21 @@ import Link from "next/link";
...
+    </Link>
+  );
+};
+
+export function truncateAccountId(
+  accountId: string,
+  lengthThreshold: number = 25
+) {
+  return accountId.length > lengthThreshold
+    ? accountId.slice(0, 8) + "..." + accountId.slice(accountId.length - 7)
```
frol, line 20: Looking at the result of abbreviation of `b7df2090560a225dc4934aed43db03a6c674c2d4.lockup.near` to `b7df2090...up.near`, I think we should do better than that, e.g. `b7df2…2d4.lockup.near` (also, consider using "ellipsis" character … if it looks better in UI)

frol (approved):

PR #447, opened 2020-10-08
feat(frontend): display detailed balance information on AccountDetails page

### LINKED ISSUE #291: Display locked token balance breakdown on Account Detail page
### Overview

Related issue in Wallet: https://github.com/near/near-wallet/issues/481

We display "locked" balance, which is the amount staked. We should also display the other types of locked balances:
- Vesting
  - Several community members (NEAR team, contributors) have tokens that vest over time, and will be unlocked linearly, block by block
- State stake (Minimum Balance)
   - There is a minimum required balance ([NEP #40](https://github.com/nearprotocol/NEPs/issues/40)) for all accounts. This can only be transferred by deleting the account
- Staking
- Delegating

### Story

As a user, I want to view both my active and locked (vesting, state staked, delegated, staked) token balances, so I can understand the quantity of tokens that are fully usable and transferrable.

### Acceptance Criteria

- [ ] Total NEAR balance (available+locked) is visible
- [ ] Available NEAR balance is visible (total - locked)
- [ ] Locked NEAR balance is visible, and broken down into
   - [ ] Staked
   - [ ] Minimum Balance (`storage_used` * `cost_per_byte`)
   - [ ] Vesting
   - [ ] Delegating
- [ ] Difference between locked and available is explained
- [ ] Minimum balance is explained
- [ ] Mobile ok
- [ ] Tested

icerove:
fix #291 
Total Balance = Account balance + Lockup balance (wallet is + unlockup balance, but frol said it is wrong)
Available Balance = Account Balance - Minimum Balance
Minimum Balance = Bytes used * cost per byte
Lockup Balance = get_lockup_balance (wallet usage)
Staked Balance = Current “Locked Balance”


icerove:
7fb82f4 commit is current version which uses RPC and our database to get lockup balance and timestamp, which is workable. But I also tried to use viewLockupState function in the https://github.com/near/account-lookup/blob/master/script.js which is really weird. I want to show up here to to get some help.

icerove:
491f774 commit is current version which uses RPC and our database to get lockup balance and timestamp, which is workable. But I also tried to use viewLockupState function in the https://github.com/near/account-lookup/blob/master/script.js which is really weird. I want to show up here to to get some help.

icerove:
I find out it is the issue for account-lookup not from mine. That one cannot be used since I think our function to decode uint8array is weird .

frol (changes requested):
https://github.com/near/near-explorer/issues/291#issuecomment-706133026

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -97,4 +116,37 @@ export default class AccountsApi extends ExplorerApi {
...
+  async queryLockupAccountInfo(accountId: string): Promise<any> {
+    // copied from https://github.com/near/near-wallet/blob/f52a3b1a72b901d87ab2c9cee79770d697be2bd9/src/utils/wallet.js#L601
+    const lockupAccountId =
+      sha256(Buffer.from(accountId)).substring(0, 40) +
+      "." +
+      LOCKUP_ACCOUNT_ID_SUFFIX;
+    try {
+      const lockupAccount = await this.queryAccount(lockupAccountId);
+      if (lockupAccount) {
+        const account = new nearAPI.Account({ provider: nearRpc });
```
icerove, line 129: need your attention to failed test which is here.

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -1,53 +1,114 @@
...
+        if (lockupInfo) {
+          this.setState({
+            lockupAccountId: lockupInfo.lockupAccountId,
+            lockupAmount: lockupInfo.lockupAmount,
+            unlockupAmount: lockupInfo.unlockupAmount,
+          });
+        }
+      })
+      .catch((err) => console.error(err));
+  };
```
frol, line 41: Move all the state manipulations away from here into `src/libraries/explorer-wamp/accounts.ts` `getAccountInfo`. This will also resolve your issue with the test.

frol (changes requested):

frontend/src/libraries/explorer-wamp/accounts.ts
frol: Ah, I see that is a TypeScript error. Pass an empty string as a second argument there, it is fine.

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -1,53 +1,114 @@
...
+      .sub(new BN(minimumBalance))
+      .toString();
+    let totalBalance;
+    if (lockupAmount && unlockupAmount) {
+      totalBalance = new BN(account.amount)
+        .add(new BN(account.locked))
+        .add(new BN(lockupAmount))
+        .add(new BN(unlockupAmount))
+        .toString();
+    }
```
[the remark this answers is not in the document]
frol: near-api-js had a bug, btw https://github.com/near/near-api-js/pull/435

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -97,4 +116,37 @@ export default class AccountsApi extends ExplorerApi {
...
+        const lockupAmount = await account.viewFunction(
+          lockupAccountId,
+          "get_locked_amount",
+          {}
+        );
+        const unlockupAmount = await account.viewFunction(
+          lockupAccountId,
+          "get_owners_balance",
+          {}
+        );
```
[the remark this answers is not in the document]
icerove: if we do not expose this, we will not get the total balance as wallet has. Total balance here is Account balance + Lockup balance + unlockup balance, I remember you said it is wrong? >_< @frol

frontend/src/libraries/explorer-wamp/accounts.ts
icerove: image

frol:
@icerove I have moved the logic to the backend since fetching the genesis config and doing a view call to the lockup contract needs to be done through it anyway (instead of using near-api-js on the client; we should not use RPC from the client side since we may end up providing inconsistent data from the backend and the direct RPC calls from the browser, so it is better to use backend to do all the calls to NEAR RPC). The changes are half-backed, please, finish them up (mostly on the frontend side, I believe).

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -18,36 +20,67 @@ export interface Props {
 export default class extends React.Component<Props> {
   render() {
     const { account } = this.props;
+    const stakedBalanceBN = new BN(account.locked);
```
icerove, line 23: @kcole16 see these calculation here and give your comments. Since there is liquid balance which maybe refer to unlocked one

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -18,36 +19,55 @@ export interface Props {
 export default class extends React.Component<Props> {
   render() {
     const { account } = this.props;
+    console.log(account);
```
frol, line 22: Please remove the log

frol (approved):

PR #456, opened 2020-10-16
feat(backend): Configure the backend to use Indexer for Explorer database

icerove:
fix first step of migration

### DIFF
--- a/backend/src/db-utils.js
+++ b/backend/src/db-utils.js
@@ -1,8 +1,7 @@
 const models = require("../models");
-const { wampNearNetworkName } = require("./config");
 
-const query = async ([query, replacements]) => {
-  if (wampNearNetworkName === "testnet") {
+const query = async ([query, replacements], from_indexer = false) => {
+  if (from_indexer) {
     return await models.sequelizePostgres.query(query, {
       replacements,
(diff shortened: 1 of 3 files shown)


backend/src/db-utils.js
```diff
@@ -1,6 +1,13 @@
 const models = require("../models");
+const { wampNearNetworkName } = require("./config");
 
 const query = async ([query, replacements]) => {
+  if (wampNearNetworkName === "testnet") {
+    return await models.sequelizePostgres.query(query, {
+      replacements,
+      type: models.Sequelize.QueryTypes.SELECT,
+    });
+  }
```
frol, line 10: You have brought a valid point here, but we cannot merge this change as it will break testnet Explorer immediately (given we don't have the fixes for the frontend yet). I am considering to have duplicate implementations on the backend for querying the data from Postgres and publishing that data to separate Wamp topics (create clones of the regular jobs pulling the data from Postgres instead of Sqlite and publishing to `chain-stats:from-indexer` instead of `chain-stats`).

frol (changes requested):

frol (approved):

PR #461, opened 2020-10-28
fix name matching and create frontend provider

icerove:
fix some unmatching names for current backend and indexer
1. fix the progresql issue with extract function
2. transactions table do not have transactionIndex column ( it is index_in_chunk in the indexer )
3. In indexer, the name for actions table is named as transaction_actions instead of actions 
4. In indexer,  transaction_actions table do not action_index but index.


frontend/src/context/IndexerDatabaseProvider.jsx.jsx
```diff
@@ -0,0 +1,72 @@
+import React, { createContext, useEffect, useCallback, useState } from "react";
```
frol, line 1: Let's have it in a separate PR. I don't understand why we cannot use the old provider and just subscribe to a different topic based on the configuration option. Please, make sure the backend sends the data of the same shape, so we can just have a thin tweak to the `DatabaseProvider` changing the WAMP topic

frol (changes requested):

backend/src/db-utils.js
```diff
@@ -149,14 +158,38 @@ const queryDashboardBlocksAndTxs = async (from_indexer = false) => {
...
     transactions.map(async (transaction) => {
       const actions = await queryRows(
         [
           `SELECT transaction_hash, action_index, action_type as kind, action_args as args
-      FROM actions
-      WHERE transaction_hash = :hash
-      ORDER BY action_index`,
+        FROM actions
+        WHERE transaction_hash = :hash
+        ORDER BY action_index`,
```
frol, line 192: Apply the same suggestion I had above: https://github.com/near/near-explorer/pull/461#discussion_r515196362

Avoid copy-pasting as it is really getting hard to maintain. Here you have a bug now since if `from_indexer` is true, then you will fetch the actions from the indexer DB first, and then from the old DB right after that.

P.S. Also, you can re-write the `map` of queries (it is almost always a bad idea to do queries in a loop) into a single query which selects `where transaction_hash IN :transaction_hashes` (see [`IN` operator](https://www.w3schools.com/sql/sql_in.asp))

frol (changes requested):

backend/src/db-utils.js
```diff
@@ -158,52 +158,34 @@ const queryDashboardBlocksAndTxs = async (from_indexer = false) => {
...
-      });
-    })
+  const actionsArray = await queryRows(
+    [query, { transactionHashes }],
+    from_indexer
   );
+  transactions.map((transaction) => {
+    let actions = actionsArray.filter(
+      (action) => action.transaction_hash === transaction.hash
+    );
```
frol, line 181: nit: Currently, it is O(N^2).

You can collect the `actionsArray` into a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) of `transactionHash -> Array of actions` once before the `.map()`.

frol (approved):
Well done! I will merge/deploy it tomorrow morning my time.

Issue #350, opened 2020-07-14
Display information and statistic on dashboard


icerove:
### Goals
- Top level stats are prominent
- Search is at least as prominent as top level stats
- Transactions are either removed entirely, or not very prominent (just a few displayed)
- Blocks are not prominent

For specific details see the [attached issues on ZenHub](https://app.zenhub.com/workspaces/app-layer-5e2b7ac4f336abcdc02917e8/issues/near/near-explorer/350)

frol:
I want to sum up all the issues we track here.

The current dashboard has:

* The number of Network Nodes:
  * the first number is the number of nodes that do block production and validation
  * the second number is the total number of nodes that send their telemetry to us that they are online (these are the validation nodes and "read-only" nodes that are maintained by partners, enthusiasts, service providers, etc)
* The most recent Block Height (on average this number gets incremented by 1 every second)
* The total number of transactions through the whole history of the chain
* The number of transactions during the last 24 hours (it is a moving window, but I think it makes more sense to just display a graph with daily stats for the last, let's say, 14 days)
* The total number of accounts
* The list of recent transactions (we consider it is useless already)
* The list of recent blocks (it is not really convenient to use that list as it gets a new item every second, so we either constantly update it or update every 10 seconds replacing the whole list with the new records; I think, the heatmap [github contributions style of graph] or a moving column chart is a better way to present the data)

We have requests to display more stats:

* [#189] Total supply over time [huge NEAR token amounts needs to be reasonably presented] (supply breakdown: Liquid tokens, Staked tokens, Locked tokens, unclaimed / etc tokens)
* [#189] Epoch reward for given epoch [measured in NEAR tokens]
* [#189] Amount of fees burnt over time [measured in NEAR tokens]
* [#122] Gas spent / gas limit over time [measured in Gas/TeraGas/etc] (we will possibly like to add a breakdown across shards)
* [#334] Gas price over time [measured in NEAR/Gas] (there seem to be a separate page for a breakdown, see the issue for details)
* [#189] State size over time [measured in bytes/megabytes/etc] (also across shards)

Overall, we probably want to display all the stats with some track of history, so I guess, we should have all of the "current state" stats backed with a graph of a historical change for that number, for example, a total number of accounts can be displayed as the total number + a graph showing how many accounts were back a few days / months / years ago.

icerove:
Please leave your comments and estimation for the new design @corwinharrell

bowenwang1996:
[...]

PR #466, opened 2020-11-04
feat: New Display information and statistic on dashboard without chart

icerove:
partially addressed #350 
2. fix the node and block card
3. fix the transaction card without chart


frol:
@icerove given the plan @corwinharrell [suggests](https://github.com/near/near-explorer/discussions/470#discussioncomment-125801):

>  We shouldn't ship any changes from the dashboard refresh until we feel like confident that the design changes applied site-wide are consistent, and that it doesn't feel like the dashboard is too divergent from the other views.

I suggest we land all the changes to the backend in a separate PR, so we can land it first, so we have the Render PR URL working fine. Could you do that?

frol:
On the other hand, in that case we will be required to make sure we don't break the currently deployed frontend (the stats should be additive instead of replacing one with another)

icerove:
remember to pull the latest change I made. @frol

frol:
@icerove I have deployed the backend and frontend to new instances on Render, so we can test it online with @corwinharrell and anyone willing to give it a try., but it currently fails due to TypeScript errors. Please, fix those and you should be able to see the deployed version on https://near-explorer-frontend-redesign.onrender.com/

frol:
The tests fail due to:

```
ERROR in /near-explorer/frontend/src/components/dashboard/DashboardHeader.tsx(88,33):
Property 'totalTransactions' does not exist on type '{ finalTimestamp: number; latestBlockHeight: number; latestGasPrice: string; numberOfLastMinuteBlocks: number; transactionCountArray: never[]; lastDayTxCount: number; last2DayTxCount: number; }'.
```

Please, address that

PR #468, opened 2020-11-06
feature(backend): Improved the configuration to support the legacy syncing backend and the new indexer infrastructure

frol:


icerove:
I would say great work but it has conflicts with my new Dashboard PR, would you consider merge the dashboard one and then merge this one? @frol

icerove (approved):

PR #483, opened 2020-11-17
feat(frontend): New design for the header and the footer and background

icerove:
*Partially address #350*
New design for the header and the footer and background with font change

See more details about the design along with the Figma: https://github.com/near/near-explorer/issues/350#issuecomment-700006882


corwinharrell:
@jakestutzman can you please provide @icerove with the most up to date NEARkat illustration for use in the explorer footer?

frontend/src/components/utils/Footer.jsx
```diff
@@ -80,16 +79,30 @@ export default () => (
       }
 
       .footer-help {
+<<<<<<< HEAD
```
frol, line 82: You are supposed to resolve merge conflicts, you know :smile:

corwinharrell (changes requested):
### Issues on Desktop
- [x] Search bar in nav is not centered
- [x] Nav items are not vertically aligned
- [x] Logo should not scale dynamically. It should be one size for desktop, and then a smaller size for tablet/mobile
- [x] There should be some padding between the icons and titles for each nav item.
- [x] Each network should be represented by a different color
- [x] "ChainInfo" is an abbreviation. It is also two words, not one. The original designs use the words "Explore", as this was a suggestion from internal team members. If you have received feedback that this is not an effective title, please share the feedback so we can determine a better alternative.
- [x] The hover and active states for the dropdown menus in the nav are not consistent with the designs
- [x] Styles for the search bar in the navigation are not consistent with the design. (hover/focus states in particular)

### Issues on Mobile
- [x] Logo and network selector are not left and right aligned.
- [x] Primary navigation is not nested within a Menu, but is instead placed under the search bar, which is inconsistent with the designs.

@icerove it seems that most of these issues could be avoided if the implementation would more closely adhere to the designs provided. Could you articulate why there are so many things here that still need to be addressed? What resources or details have you felt are missing in order to be able to accurately implement the designs that were proposed for these changes?

frol (changes requested):
## Desktop and mobile issues:

The footer is not attached to the bottom of the window:

Screenshot from 2020-11-27 17-23-07

## Mobile issues:

There is no network dropdown at all.

NEARkat should be together with the "join the community":

Screenshot from 2020-11-27 17-17-13

Top navbar dropdown has different corner styles on hover and without hover:

image

image

icerove:
@frol 
almost finish all feature for the header and footer.
Here are something missing:
1. cannot get the innerwidth so search bar is not showing on the dashboard on mobile.
2. on hover of nav dropdown, the icon color is not changed.

jakestutzman:
@corwinharrell @icerove @frol I'd like us to make a few tweaks to accommodate a small 'window' for the nearkats to hang out of. Here's what I'm thinking. Any red flags here? 

image


https://www.figma.com/file/tW2iLtA1psHki0D1vKExf1/Dashboard-Refresh?node-id=0%3A1

jakestutzman:
Here's the mobile view 
image

corwinharrell:
@jakestutzman I like the little window idea, but don't know if the border is necessary. We're already calling a lot of attention to this component, so imo, the more subtle the better. Are you opposed at all to omitting it?

icerove:
@corwinharrell I fix the footer for new design

jakestutzman:
> @jakestutzman I like the little window idea, but don't know if the border is necessary. We're already calling a lot of attention to this component, so imo, the more subtle the better. Are you opposed at all to omitting it?

I think the outline helps balance the line art of the nearkats. I tried it without and it didn't seem to 'contain' it as well. Feel free to make a final decision on it tho.

frontend/src/components/utils/Search.jsx
```diff
@@ -1,141 +1,128 @@
...
-          padding-bottom: 5px;
-        }
+          .search-field {
+            background: ${this.props.dashboard ? "#FFFFFF" : "#FAFAFA"};
+            border: 2px solid #eaebeb;
+            border-left: ${this.props.dashboard ? "2px solid #eaebeb" : "none"};
+            box-sizing: border-box;
+            border-radius: 8px;
+            height: 100%;
+            font-weight: lighter;
```
frol, line 115: Please, remove this rule since it is too light, and it is really hard to read. The default font-weight matches the Figma design.

frol (changes requested):
Here are a few discrepancies I notice:

* [ ] The networks dropdown does not match the design when open (the border should not stand out, the width of the dropped down element should match the width of the button, if possible)
* [ ] the border when clicked on the "Explore" dropdown should not be there
* [x] the font-weight on the search field in the top navigation is too light (see the comment about font-weight)
* [x] the footer in mobile version is broken
image

frol:
* [ ] the footer is below the screen when the content fits the screen just fine - the footer should be visible if the main content fits the page. Reproduction: https://near-explorer-frontend-pr-483.onrender.com/blocks/qwe

frol:
@icerove I had unchecked the items that still do not match the design as you asked.

Here is one that is the follow-up to the last item about the footer:

* [ ] The footer on a page with little content should not cause the page to have a vertical scroll

frontend/src/pages/_app.jsx
```diff
@@ -144,23 +120,13 @@ export default class extends App {
...
-          }
-
-          .page {
-            background: white;
-          }
-
           .app-wrapper {
             position: relative;
-            min-height: calc(100vh - 120px);
+            min-height: calc(100vh - 118px);
```
frol, line 129: 118px is not enough to avoid vertical scroll on a page where there is not enough content, while 119px works fine

frol (approved):
Great job!

@corwinharrell could you review the header/footer redesign? https://near-explorer-frontend-pr-483.onrender.com/ How do you feel about releasing this version to production [without the new dashboard] (the new dashboard is in a separate PR at the moment)?

frol:
@icerove I have reviewed the initial comments from @corwinharrell, and it seems that we still miss a few items. Please, review them and let me know if we cannot/unclear how to address

frontend/src/components/utils/HeaderNavDropdown.jsx
```diff
@@ -0,0 +1,111 @@
...
+          background: #000000;
+          color: white;
+        }
+      `}</style>
+    </Dropdown.Item>
+  );
+};
+
+export default () => (
+  <Dropdown>
```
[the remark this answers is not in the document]
icerove: need help for this. @frol 
file is HeaderNavDropdown.js

frontend/src/components/utils/Header.jsx
```diff
@@ -1,156 +1,94 @@
...
+            </Col>
+
+            <Col xs="2" className="align-self-center text-right d-md-none">
+              <MobileHeaderNavDropdown />
+            </Col>
+          </Row>
+        </Col>
+
+        <Col className="align-self-center text-center" md="8" lg="8" xs="12">
+          {router.pathname !== "/" ? <Search /> : isMobile ? <Search /> : null}
```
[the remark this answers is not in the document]
icerove: I change it with your comment, but weird thing happen as on dashboard page, the search box for mobile one is not fit for design on other page. @frol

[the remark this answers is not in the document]
frol: @icerove Could you provide screenshots? I am not exactly following your concern. It seems to be working fine for me except that on mobile the search box is higher on the dashboard page than on any other page:

image

(the search input is 306x49px)

image

(the search input is 306x40px)

NOTE: The second style is what we want to have everywhere.

frontend/src/components/utils/HeaderNavDropdown.jsx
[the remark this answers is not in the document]
frol: It looks good to me now. How can I help there?

frontend/src/components/utils/HeaderNavDropdown.jsx
```diff
@@ -0,0 +1,125 @@
+import { Dropdown } from "react-bootstrap";
+
+import IconAccounts from "../../../public/static/images/icon-accounts.svg";
+import IconBlocks from "../../../public/static/images/icon-blocks.svg";
+import IconTransactions from "../../../public/static/images/icon-transactions.svg";
+import IconNodes from "../../../public/static/images/icon-nodes.svg";
+
+const HeaderNavItem = ({ link, imgLink, text }) => {
+  return (
+    <Dropdown.Item className="header-nav-item" href={link}>
```
frol, line 10: We need to use the `<Link>` tag from next.js so the transitions are client-side instead of full reload.

frontend/src/components/utils/Search.jsx
```diff
@@ -1,141 +1,142 @@
...
+            border: 2px solid #cdcfd1;
+          }
+
+          .search-field:disabled,
+          .search-field[disabled] {
+            background: #eaebeb;
+            border: 1px solid #eaebeb;
+          }
+
+          .button-search {
```
frol, line 127: Please, fix the `:hover` button background. It is currently somewhat green:
image

frol (changes requested):
These are the final review notes from me.

@corwinharrell I would love to hear your review. See the deployed staging version here: https://near-explorer-frontend-pr-483.onrender.com/

frol (changes requested):
* [ ] Please, resolve the merge conflict
* [ ] There are way too many borders on the search input (both on `:hover` and `:active` state): 
image

frol (changes requested):
I wish I could just merge it, but I have to send another changes request :confounded: 

* [ ] image

the border does not match the design

image
* [ ] 
image

the border shifts the input a few pixels on hover and focus and also does not match the design
* [ ] 
image

when you click on the Search button, it goes dark green. @corwinharrell what should be the background color for the pressed button?

@icerove please, pair up with @corwinharrell to iterate faster.

frol:
@icerove Hints about the borders styling:

* use `.parent-block:focus-within .button` to style the border on the block that needs to be styled with the input is focused
* use `.parent-block:hover .button` to style the whole search box when a user hovers over it (instead of only capturing the input hover event itself

corwinharrell:
@icerove Please let me know if you have any questions regarding the styling issues @frol mentioned above. I'm happy to help in providing the necessary guidance to ensure that the implementation reaches parity with the design.

corwinharrell:
@icerove I've gone ahead and [styled a version on codepen](https://codepen.io/corwinharrell/pen/XWjQbMd) if you'd like to use it as reference for how this should be implemented.

If you'd like me to go ahead and do this for the search input inside of the header as well, just let me know.

frol (approved):

frol (approved):

corwinharrell (changes requested):
There are still issues with the implementation of the new inputs.

- [ ] The borders and border radii are not consistent and aligned, which results in gaps between the inputs and border at the corners. This is most visible in the focused state:

Screen Shot 2021-01-25 at 10 26 59 AM

- [ ] In addition, the search input nested in the header also has a different background color than the rest of the input. It's very slight but its there.

Screen Shot 2021-01-25 at 10 27 07 AM

corwinharrell (changes requested):
Almost there. There is a remnant style from the old search input that adds a teal shadow/outline of sorts on the search button when pressed. This should be removed.

Screen Shot 2021-01-26 at 12 55 18 PM

corwinharrell (approved):

PR #485, opened 2020-11-19
feat(frontend): Introduced an API endpoint for circulating supply

frol:
Here is the link to test it out: https://near-explorer-frontend-pr-485.onrender.com/api/circulating-supply

Once it is live, it will be available as https://explorer.mainnet.near.org/api/circulating-supply

## Testing Plan

* [x] Manually test the endpoint for mainnet and confirm that the endpoint return 404 when non-mainnet network is queried.


evgenykuzyakov (approved):

icerove (approved):

icerove:
the data is listed as row json and could not change, can we change to function calculating these data. Once it changes somehow, it is hard to maintain

frol:
@icerove It is in the CSV, which @eriktrautman provided me with. I don't know how the values were acquired.

@evgenykuzyakov @eriktrautman it is a function that we can use instead of the computed values?

evgenykuzyakov:
I think it's more complicated due to the some cliffs, e.g. 1 year cliff

PR #488, opened 2020-11-23
feat(frontend): Display yoctoN/gas in the gas price tooltip

frol:


frontend/src/components/utils/GasPrice.jsx
```diff
@@ -10,8 +10,11 @@ export default ({ gasPrice }) => {
   let show = formatNEAR(TerraGas);
   let precise = showInYocto(gasPrice);
   return (
-    <OverlayTrigger placement={"bottom"} overlay={<Tooltip>{precise}</Tooltip>}>
-      <span>{show} Ⓝ/ Tgas</span>
+    <OverlayTrigger
+      placement={"bottom"}
+      overlay={<Tooltip>{precise}/Tgas</Tooltip>}
```
chadoh, line 15: The precise value is _per gas unit_, not per Tgas, which would be 1¹² gas units

At minimum gas price, we show "0.0001Ⓝ/Tgas". We can work this back and verify that we should be showing "100,000,000 yN/gas unit" in the tooltip:

0.0001 N/Tgas
= 10²⁰ yN / Tgas
= 10²⁰ yN / 10¹² gas
= 10⁸ yN / gas
= 100,000,000 yN / gas

chadoh: Can we link to our documentation page about gas in the tooltip? Something like "100,000,000 Ⓝ/gas unit – [more info](https://docs.near.org/docs/concepts/gas)

frol: Ouch, great call. I have misread the code.

The link to the docs are available when you click on the "(i)" sign next to "Gas Price" title when you hover over it.

chadoh (approved):
Thanks!

icerove (approved):

PR #497, opened 2020-12-09
feat(frontend): Brought back the native balances

frol:


khorolets (approved):

ilblackdragon:
Looks reasonable.

I would still leave a notice on top explaining where all the balances go for a week or two.

PR #477, opened 2020-11-13
feat(frontend): Migrate Explorer to the Indexer for Explorer

icerove:
migration plan frontend part.
finish #450 2 3 and 4 part

PR #499, opened 2020-12-09
fix(backend): Aligned the indexer support on the backend to produce the same events as the legacy-sync backend

frol:
While working on #477, I hit a few discrepancies with the Indexer support on the backend.

### DIFF
--- a/backend/src/db-utils.js
+++ b/backend/src/db-utils.js
@@ -246,8 +246,8 @@ const queryDashboardBlocksAndTxs = async ({ dataSource }) => {
           ) AS recent_blocks
           LEFT JOIN blocks ON blocks.${blockHashColumnName} = recent_blocks.${blockHashColumnName}
           LEFT JOIN transactions ON transactions.${transactionBlockHashColumnName} = recent_blocks.${blockHashColumnName}
-          GROUP BY hash
-          ORDER BY timestamp DESC`,
+          GROUP BY blocks.${blockHashColumnName}
+          ORDER BY blocks.${blockTimestampColumnName} DESC`,
       ],
       { dataSource }
     ),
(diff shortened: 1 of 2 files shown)


backend/src/wamp.js
```diff
@@ -161,6 +163,10 @@ wampHandlers["get-account-details"] = async ([accountId]) => {
     nearRpc.sendJsonRpc("EXPERIMENTAL_genesis_config", {}),
   ]);
 
+  if (accountInfo === null) {
+    return null;
+  }
```
frol, line 168: It is easier to handle the missing account information when it returns `null` instead of raising some cryptic error message

backend/src/db-utils.js
```diff
@@ -322,11 +346,13 @@ const aggregateStats = async (options) => {
...
-    totalTransactions: totalTransactions.total,
-    lastDayTxCount: lastDayTxCount.total,
-    lastBlockHeight: lastBlockHeight ? lastBlockHeight.block_height : 0,
+    totalAccounts: parseInt(totalAccounts.total),
+    totalBlocks: parseInt(totalBlocks.total),
+    totalTransactions: parseInt(totalTransactions.total),
+    lastDayTxCount: parseInt(lastDayTxCount.total),
+    lastBlockHeight: lastBlockHeight
+      ? parseInt(lastBlockHeight.block_height)
+      : 0,
```
frol, line 355: Postgres returns these as strings (I assume it uses int64 numbers here, which don't fit into JS number in the worst-case scenario, but we don't expect any of the values to overflow over 2^53 any time soon)

backend/src/db-utils.js
```diff
@@ -284,6 +301,13 @@ const queryDashboardBlocksAndTxs = async ({ dataSource }) => {
       });
     }
   });
+
+  if (dataSource === DS_INDEXER_BACKEND) {
+    blocks.forEach((block) => {
+      block.timestamp = new BN(block.timestamp).divn(10 ** 6).toNumber();
+    });
+  }
```
frol, line 309: We need to adjust the timestamp to be the same style as the legacy backend used to provide

backend/src/db-utils.js
```diff
@@ -270,7 +281,13 @@ const queryDashboardBlocksAndTxs = async ({ dataSource }) => {
...
     }
   });
-  transactions.map((transaction) => {
+
+  transactions.forEach((transaction) => {
+    if (dataSource === DS_INDEXER_BACKEND) {
+      transaction.block_timestamp = new BN(transaction.block_timestamp)
+        .divn(10 ** 6)
+        .toNumber();
+    }
```
frol, line 290: Same legacy vs indexer adjustment

frol (commented):
Here are some review tips

icerove (approved):

PR #504, opened 2020-12-16
Closes #489: Compute Circulating Supply and expose via backend

### LINKED ISSUE #489: Compute Circulating Supply and expose via backend
*This is only applicable to mainnet*

*This issue is blocked on #478*

`Circulating Supply = Total Supply - Tokens locked in the Lockup contracts`

## Implementation Notes

Here is how to compute the value:

1. Total Supply is available though RPC `block` method (see `header->total_supply`)
2. Tokens locked in the lockup contracts needs to be queried individually:
2.1. Query Indexer for Explorer database for all the lockup accounts: `SELECT account_id FROM accounts WHERE account_id LIKE '%.lockup.near'`
2.2. Use [account-lookup code](https://github.com/near/account-lookup) to query the still locked balance
3. Sum up all the locked balances and subtract it from the Total Supply

NOTE: Make sure you pin a specific block height when you query the Total Supply and every individual contract state (through the account-lookup)

Compute the value every 2 hours on the backend and store the value in memory (if the computation takes too long, consider storing it in SQLite). Expose the precomputed value via WAMP `get-circulating-supply` method.

## Acceptance Criteria

- [x] The value is computed correctly (compare with the projection)
- [x] Non-mainnet networks return undefined
- [x] The value is re-computed every 2 hours (note: adjust the number if necessary)

khorolets:
Closes #489 

As for Acceptance Criteria

> The value is re-computed every 2 hours (note: adjust the number if necessary)

The value is re-computed every 1 hour instead of 2 because it takes ~20s to run.

> Non-mainnet networks return undefined

returns 404, didn't change that


frol (commented):
Overall looks good. We need to confirm that the values are in line with the expectations

backend/src/aggregations.js
```diff
@@ -0,0 +1,220 @@
...
+const readOption = (reader, f) => {
+  let x = reader.read_u8();
+  if (x == 1) {
+    return f();
+  }
+  return null;
+};
+
+// viewLockupState function taken from https://github.com/near/account-lookup/blob/master/script.js
+const viewLockupState = async (contractId, blockHeight) => {
```
MaximusHaximus, line 22: Just something that I noticed and was curious about; is it safe for `viewLockupState()` to retry _indefinitely_ when it encounters _any_ type of error, with no boundary to how many times it retries, and no error reporting if it has to retry many times to finish? Feels like this could permanently deadlock if fed e.g. a bad contractId?  Same with `while(true)` loops on L126 & L152  of this file.

frol: @MaximusHaximus Thanks for pointing those out! This is a draft PR that we needed to collect the figures in the first place. Since then we have done more exploration of the space, and @telezhnaya will take it over from here solidifying the implementation and having graceful retry strategies.

frol:
Closing in favor of #636

PR #505, opened 2020-12-18
feat(api): Added some internal temporary APIs to stats for Town Halls

frol:
These are for internal use only. The queries are heavy and may take up to a few minutes to resolve. We don't plan to support them after we implement dedicated statistics pages on Explorer frontend.

### DIFF
--- a/backend/src/db-utils.js
+++ b/backend/src/db-utils.js
@@ -356,12 +356,12 @@ const aggregateStats = async (options) => {
   };
 };
 
-const queryTransactionsByDate = async () => {
+const queryTransactionsCountAggregatedByDate = async () => {
   return await queryRows(
     [
       `SELECT
           TIMESTAMP 'epoch' + DIV(DIV(blocks.block_timestamp, 1000000000), 60 * 60 * 24) * INTERVAL '1 day' AS "date",
-          COUNT(*) AS transactions_by_date
+          COUNT(*) AS transactions_count_by_date
         FROM transactions
         JOIN blocks ON blocks.block_hash = transactions.included_in_block_hash
         GROUP BY "date"
@@ -371,7 +371,7 @@ const queryTransactionsByDate = async () => {
   );
 };
 
-const queryTeragasUsedByDate = async () => {
+const queryTeragasUsedAggregatedByDate = async () => {
   return await queryRows(
     [
       `SELECT
@@ -386,12 +386,12 @@ const queryTeragasUsedByDate = async () => {
   );
 };
 
-const queryNewAccountsByDate = async () => {
+const queryNewAccountsCountAggregatedByDate = async () => {
   return await queryRows(
     [
       `SELECT
         TIMESTAMP 'epoch' + DIV(DIV(blocks.block_timestamp, 1000000000), 60 * 60 * 24) * INTERVAL '1 day' AS "date",
-        COUNT(*) as new_accounts_by_date
+        COUNT(*) as new_accounts_count_by_date
       FROM accounts
       JOIN receipts ON receipts.receipt_id = accounts.created_by_receipt_id
       JOIN blocks ON blocks.block_hash = receipts.included_in_block_hash
@@ -402,12 +402,12 @@ const queryNewAccountsByDate = async () => {
   );
 };
 
-const queryNewContractsByDate = async () => {
+const queryNewContractsCountAggregatedByDate = async () => {
   return await queryRows(
     [
       `SELECT
         TIMESTAMP 'epoch' + DIV(DIV(receipts.included_in_block_timestamp, 1000000000), 60 * 60 * 24) * INTERVAL '1 day' AS "date",
-        COUNT(distinct receipts.receiver_account_id) AS new_contracts_by_date
+        COUNT(distinct receipts.receiver_account_id) AS new_contracts_count_by_date
       FROM action_receipt_actions
       JOIN receipts ON receipts.receipt_id = action_receipt_actions.receipt_id 
       WHERE action_receipt_actions.action_kind = 'DEPLOY_CONTRACT'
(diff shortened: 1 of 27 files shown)


icerove:
nice work!
there is some updates for this: 
Hey, have an update on the metrics needed on explorer Dashboard
# of smart contracts deployed (w/ chart)
# of accounts created (chart? explorer just has fixed #)
Growth in stake (w/ chart)
Gas paid (w/ chart)
optional bonus — GDP
Weekly + Daily Active Accounts <Number of>
Weekly + Daily Active Contracts <Number of and List of Names>
Number of Transactions
Transactions Volume <all transactions amount aggregation>
Number of Staking
Staking Volume
Weekly +Daily Active Validators <Number of and List of Names>
13. Average Transaction Size
@frol

frol:
@icerove as we discussed, we will need proper support on the backend with caching. The implementation of all the charts will require:

1. Craft the queries
2. Designing the framework on the backend for caching the precomputed values (it will quickly grow into something we cannot maintain efficiently if we just keep adding the metrics ad-hoc, so we need some structure, at least, having a `stats` folder on the backend)
3. Implementation on the backend
4. Design the framework to query the data on the frontend
5. Implement the charts

Which parts of the scope are you comfortable leading?

icerove:
important updates: 
these are most important metrics:
# of smart contracts deployed (w/ chart)
# of accounts created (chart? explorer just has fixed #)
Gas paid (w/ chart)
Weekly + Daily Active Accounts <Chart + Most active accounts list>
Weekly + Daily Active Contracts <Chart + Most active contracts list>
Number of Transactions Daily <Chart>
Staking Volume <Chart>

icerove:
definitely can leading for 3, 4, 5 since really familiar with those. 1 and 2 I would like to sync with you together. @frol

frol:
@khorolets I would like to ask you to help with (1) and (2). As to (1), let's focus on the important metrics first. As to (2), I see it somewhat similar to the approach you took for circulating supply caching, just come up with some project layout and pattern, so adding more metrics will be just a matter of following the same flow (ideally, with as little boilerplate as possible).

frol:
P.S. The queries for the following stats can be extracted from this PR:

* \# of smart contracts deployed
* \# of accounts created
* ~Gas paid~ Gas used (we need to clarify the intent of this metric and whether sum of `gas_used` in chunks is accurate to estimate the paid gas or should we account the `gas` attached to function calls with Evgeny Kuzyakov after holidays, so far, `gas_used` should be enough)
* Number of Transactions Daily

backend/src/db-utils.js
```diff
@@ -356,6 +356,68 @@ const aggregateStats = async (options) => {
...
+        FROM transactions
+        JOIN blocks ON blocks.block_hash = transactions.included_in_block_hash
+        GROUP BY "date"
+        ORDER BY "date"`,
+    ],
+    { dataSource: DS_INDEXER_BACKEND }
+  );
+};
+
+const queryTeragasUsed = async () => {
```
frol, line 374:
```suggestion
const queryTeragasUsedByDate = async () => {
```

frontend/src/libraries/explorer-wamp/stats.ts
```diff
@@ -0,0 +1,39 @@
+import { ExplorerApi } from ".";
+
+export interface TransactionsByDate {
+  date: string;
+  transactions: number;
```
frol, line 5:
```suggestion
  transactionsCount: number;
```

Apply the same naming for accounts and contracts

frontend/src/libraries/explorer-wamp/stats.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export interface TeragasUsedByDate {
+  date: string;
+  teragas: number;
```
frol, line 10:
```suggestion
  teragasUsed: number;
```

bowenwang1996:
> whether sum of gas_used in chunks is accurate to estimate the paid gas

It is equal to the amount of gas burnt. It is not clear to me why you want to know the amount of gas attached given that it is generally the case that people attach more than needed.

icerove:
finish current frontend implementation and wait for new queries. @frol @khorolets

frol:
@bowenwang1996 I was not following closely the topic after we discussed to burn all the attached tokens. So just to double-check, can you confirm that the sum of `gas_used` in the chunk headers is what we want to present here? (If that is the case, it is done correctly now)

backend/src/db-utils.js
```diff
@@ -356,6 +356,68 @@ const aggregateStats = async (options) => {
   };
 };
 
+const queryTransactionsByDate = async () => {
```
frol, line 359: @khorolets This `queryTransactionsByDate` misleads the reader to assume that it can be used to query a list of transactions by a given date. I suggest we rename it to something like `queryTransactionsCountAggregatedByDate`

backend/src/stats.js
```diff
@@ -0,0 +1,98 @@
...
+  queryNewContractsByDate,
+} = require("./db-utils");
+const { formatDate } = require("./utils");
+
+let TRANSACTIONS_BY_DATE = null;
+let TERAGAS_USED_BY_DATE = null;
+let NEW_ACCOUNTS_BY_DATE = null;
+let NEW_CONTRACTS_BY_DATE = null;
+
+const calculateTransactionsByDate = async function () {
```
frol, line 14: @khorolets Consider renaming to

```suggestion
const aggregateTransactionsCountByDate = async function () {
```

frol (commented):
@khorolets I bikeshed some naming. Consider other places like wamp-endpoint names as well

@icerove I want to mention that all the bikeshedding about the code style and project structure I do is what I think can help you to be more productive and us (as a team) be more efficient going forward, so these nits are not to bother you all the time instead, I try to share my reasoning framework you can apply next time you face a similar situation

bowenwang1996:
> So just to double-check, can you confirm that the sum of gas_used in the chunk headers is what we want to present here?

Yes

icerove:
I do not see the api function for staking one, are you still working on that or I will add that directly? @frol @khorolets

icerove:
Note: 
1. for active accounts and contracts, they also want weekly data, should we add query or aggregate the data from frontend

2. also for active accounts and contracts, they want most active list, we should also have those query and make it table I think. 

@frol @khorolets

icerove:
we need active accounts and contracts for both daily and weekly. And need list of 10 most active guys

icerove:
@khorolets I will finish and merge this PR today and will add staking in a separate PR tomorrow. Since this is in urgent need somehow.

icerove (approved):

PR #507, opened 2020-12-18
feature(frontend): Redesigned dashboard

### LINKED ISSUE #350: Display information and statistic on dashboard

icerove:
This is a follow-up PR to #483 to resolve #350.

add transaction Count History chart on dashboard


frol (changes requested):
I guess we are blocked on #483 here.

* [x] The font family and the font-weight do not seem to match the one that is used on Figma
* [x] The arrows are not aligned and the font color does not match the design: 
image

icerove:
1. font family is already changed in the PR #483 
2. yes, it is different from design since the design do not consider the maker for the term explanation and all cell will link to new page will cover the maker that for term explanation.

frontend/src/components/utils/ListHandler.tsx
```diff
@@ -158,9 +158,9 @@ export default (
                   <Update
                     count={
                       config.category === "Block"
-                        ? context.lastBlockHeight
+                        ? context.latestBlockHeight
```
frol, line 161: Revert these changes, I had fixed it in master already

frontend/src/components/utils/Update.tsx
```diff
@@ -9,13 +9,13 @@ export default ({ category, count }: Props) => {
   }
   return (
     <div className="update">
-      {count > 1 ? (
+      {category === "Block" ? (
```
frol, line 12: Fixed in master

frontend/src/context/DatabaseProvider.jsx
```diff
@@ -5,49 +5,50 @@ import { ExplorerApi } from "../libraries/explorer-wamp";
 
 const DatabaseContext = createContext({
   finalTimestamp: 0,
-  lastBlockHeight: 0,
-  totalBlocks: 0,
-  totalTransactions: 0,
-  totalAccounts: 0,
+  latestBlockHeight: 0,
```
frol, line 8: I believe it is already fixed in master

frol (changes requested):
Mobile version needs more love:

image

frol (approved):
I approve it to mark it ready, but we should not merge it before #483. Once #483 is merged, we need to rebase/merge master, and ensure that the look matches the design.

* [ ] The transactions graph needs some styling in terms of font size, color, padding, etc.

icerove:
please have a look and give comments. @corwinharrell @frol

frol (changes requested):
* [ ] These are not good defaults: image
* [ ] We should adjust the graph height, font color, and other minor details that we can tweak, otherwise, explicitly state which parts we cannot tweak

icerove:
The chart x and y axis and label and tooltip date format since it is the same as x label cannot be exact the same as design. @frol @corwinharrell

frol:
FYI I am fixing some general layout styling issues here. I will push them later today

frol:
I have mostly fixed the styling and layout issues, but I broke the colors of the font and the right arrow during the refactoring:

image

It is quite late here, so I am heading to bed.

@corwinharrell It would be great if you can take a glance at the latest version now.

corwinharrell (changes requested):
This is looking better. Here are some inconsistencies that I identified:

### Desktop

- [x] Search bar width should be the same as the container that wraps the cards

Screen Shot 2021-01-29 at 11 44 35 AM

- [x] On hover, the metric number should be blue.
- [ ] The padding around the cards should be consistent on all sides. -----Not very sure what this refer to @frol @corwinharrell 

Screen Shot 2021-01-29 at 11 45 07 AM

- [ ] There is a strange flash of a black outline on the network menu onMouseLeave ---- This is from `.show > .btn-secondary.dropdown-toggle` part. I am not sure why this happened. Have a look. 

Kapture 2021-01-29 at 11 46 43

- [x] The styling on the graph does not match the designs. Primarily the borders and step markers. See designs for reference. -----change a lot, the y label cannot be the same and the vertical line is for tooltip and cannot remove.   

### Mobile

- [x] The logo in the header uses the full version instead of the icon version
- [x] The network dropdown is not centered.
- [x] Letter-spacing for the nav items is inconsistent
- [x] The expanded menu is not positioned correctly (no space between menu and search bar.
- [x] There is excessive padding on the left and right sides of the expanded menu

Screen Shot 2021-01-29 at 11 50 35 AM

- [x] The card headers are broken and wrap on mobile

Screen Shot 2021-01-29 at 11 53 21 AM

- [x] The footer design is incorrect.
- [x] Between certain breakpoints, there is not spacing between the dashboard cards

Screen Shot 2021-01-29 at 11 55 11 AM

- [x] Borders are inconsistent at different breakpoints within the Transactions card

This is what I've caught for now. I will add more as I am able.

icerove:
will touch again and fix some issues.

thisisjoshford:
awesome job y'all! 💪  looks like this fixes #527 as well 🙂

corwinharrell (changes requested):
Looking much better. 2 things I noticed.

1. I'd make a pass over the font-sizes to make sure they are correct. For example, the "View All" links are bigger than the card headings.

2. The indicators in the network dropdown are all the same color, so unless you actually select one, the unique color associated with each ENV is not visible. Each network should be represented by its unique color at all times, including in the dropdown.

frol (approved):
I believe we can address the last item in a separate PR. I want to release this PR now to address a jumping page issue and unnecessary scroll bars.

PR #510, opened 2021-01-05
Add staking volume

icerove:
Add total stake and stake bar chart to the stats page
add tooltip for bar chart
fix the date showing


icerove:
remind me here to fix the date showing

frol (changes requested):
Well done!

There are two nits:

* [x] I think it will be more useful if the validators are ordered by their stake (in descending order).
* [ ] It seems that given the volume of the cumulative stake, we can ignore the sub-NEAR amounts (ignore the fractional part)

frol (approved):

PR #514, opened 2021-01-12
Stats/partner stats on a new partner page

icerove:
this PR is build under the requirements under business team, they want our five partner transactions numbers:
1. use query to find the account that with company name -- backend
2. use query to get total transactions count and first 3 month transactions count for list of partners -- backend
3. generate /partner page for stats display -- frontend


icerove:
@frol 
Please check the query for each part since it is important to make sure it is correct and then let's discuss with Sasha whether it is okay or not.

frol (approved):
Overall, looks good, just update the query and let's ship it!

PR #517, opened 2021-01-16
Make search field less annoying on mobile

vgrichina:
Auto-capitalization messes with ability to type account name like `bananaswap.near` easily


icerove:
Thanks for fixing this, I also fix it on the new PR for new dashboard search

vgrichina:
> Thanks for fixing this, I also fix it on the new PR for new dashboard search

yeah I think it pays off a lot to set these properties and relatively low lift

it also often helps to set granular input type where it makes sense:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types

vgrichina:
@icerove do you see any blockers to merge this?

frol (approved):

PR #528, opened 2021-01-27
[frontend]: Fix sorting in stats

### LINKED ISSUE #511: Display top active accounts and contract in descending order
* [x] Sort the top active accounts and contracts in from most active to least active (update the SQL query on the backend)
* [x] Make sure that we display all the names on the left (currently we only display 5 out of 10):
image

shelegdmitriy:
Fixes #511

### DIFF
--- a/frontend/src/components/stats/ActiveAccountsList.tsx
+++ b/frontend/src/components/stats/ActiveAccountsList.tsx
@@ -10,6 +10,11 @@ export default () => {
   useEffect(() => {
     new StatsApi().activeAccountsList().then((accounts: Account[]) => {
       if (accounts) {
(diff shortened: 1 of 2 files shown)


frol (approved):
This already partially addresses the issue, but I still only see 5 out of 10 names on the left:

image

P.S. I use Firefox 85 on Arch Linux (Gnome)

shelegdmitriy:
@frol , I'm still trying to reproduce this issue. I also checked this in Firefox but for for me it's still visible 10 names. I'll try to figure out what is going on

frol:
I think it makes sense to address it on the backend (reverse the sorting order)

PR #530, opened 2021-02-05
fix the link issue

icerove:
fix the issue for docs link in the modal do not work


frontend/src/components/utils/Term.tsx
```diff
@@ -59,6 +59,8 @@ export default class extends React.Component<Props, State> {
               margin-left: 5px;
               width: 16px;
               cursor: pointer;
+              postion: relative;
+              z-index: 10;
```
frol, line 63: Hmm, what was the issue with this? I believe, `stopPropagation` is indeed the fix, but these two are not clear to me yet.

frol (approved):

frontend/src/components/utils/Term.tsx
icerove: will remove this, this is what I try to see whether css can help.

PR #531, opened 2021-02-05
delete old query for dashboard

icerove:
remove unused  query from backend for old dashboard


frol:
Reminder: We need to finish the proper deployment of testnet and mainnet vs betanet and guildnet before merging this.

frol (approved):
Well, it was related only to the old dashboard, so it is fine to remove it indeed.

PR #538, opened 2021-02-09
test: Add tests for the main dashboard page

### LINKED ISSUE #533: Add advanced end-to-end tests for the main dashboard
## Story

As a regular user who uses Explorer, I want to use a reliable service so I can access the data at any time.

## Acceptance Criteria

- [x] There are end-to-end (cypress) test(s) covering:
  - [x] Initial page load
  - [x] The stats are loaded (in timely manner, let's try 5 seconds limit)
  - [x] The stats are updated over time (let's check in 5-10 seconds)

shelegdmitriy:
Resolves #533

### DIFF
--- a/frontend/cypress/integration/dashboard.spec.js
+++ b/frontend/cypress/integration/dashboard.spec.js
@@ -10,4 +10,71 @@ context("Dashboard", () => {
   it("Logo is visible", () => {
     cy.get(".header-container .near-main-logo").should("exist");
   });
+
+  it("Search bar is visible", () => {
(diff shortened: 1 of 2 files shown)


shelegdmitriy:
@frol , please, could you tell me am I on the wright way or not?

frol (approved):
@shelegdmitriy great job!

Overall, it looks exactly what we want to have, but I think we can be less specific in our selector rules, e.g. skip bootstrap classes (`.container`, `.row`, `.no-gutters`, etc), and only use those classes that are relevant to the test case ignoring everything irrelevant.

PR #542, opened 2021-02-12
fix(frontend): Properly handle the finalizedTimestamp event from WAMP (use named arguments instead of positional)

frol:


frontend/src/context/DatabaseProvider.tsx
```diff
@@ -41,42 +41,36 @@ export default (props: Props) => {
...
-    finalTimestamp: string;
-  }) => {
-    dispatchFinalTimestamp(new BN(finalTimestamp));
+  const storeFinalTimestamp = (
+    _positionalArgs: any,
+    namedArgs: {
+      finalTimestamp: string;
+    }
+  ) => {
+    dispatchFinalTimestamp(new BN(namedArgs.finalTimestamp));
```
frol, line 73: @icerove this is the core of the change. After my previous refactoring, I failed to update this method to expect the data in the *second* function parameter (namedArgs)

icerove (approved):

PR #545, opened 2021-02-16
feat(frontend): Configure redirects for upper-case account IDs

### LINKED ISSUE #493: Configure redirects for upper-case account IDs
## Story

As a regular Explorer user who types account ID, I want to navigate to a relevant page without caring about lower/upper-case ([only lower-cased account IDs are valid](https://nomicon.io/DataStructures/Account.html#account-id-rules)).

These should be the same:

* https://explorer.mainnet.near.org/accounts/frol.near
* https://explorer.mainnet.near.org/accounts/FROL.near (this should redirect to `frol.near`)
* https://explorer.mainnet.near.org/accounts/FROL.NEAR (this should redirect to `frol.near`)
* https://explorer.mainnet.near.org/accounts/FRol.neAR (this should redirect to `frol.near`)

## Implementation notes

When a user navigates to the link directly, we need to check if it is all lower-case, and otherwise, just permanently redirect ([see next.js docs](https://nextjs.org/docs/basic-features/data-fetching#getserversideprops-server-side-rendering)) to the same page with the account ID in the lower case.

## Acceptance Criteria

- [ ] Search is lowering case of the input before doing the account query
- [x] Search bar has [auto-capitalization disabled for mobile devices](https://stackoverflow.com/a/5171812/341267)
- [ ] Direct URLs with upper-case account ID should redirect to the lower-cased URL
- [ ] End-to-end tested:
  - [ ] Search from the dashboard page should work for accounts in lower/upper/mixed case
  - [ ] Direct navigation to `/accounts/:accountId` page should work for accounts in lower/upper/mixed case
  - [ ] Direct navigation to `/accounts/:accountId` page with account ID in upper/mixed case should return HTTP 301 (Permanent Redirect)

shelegdmitriy:
Resolves #493


icerove (approved):

frol (approved):
LGTM.

@shelegdmitriy I refactored our test suite, please, take a look, especially at the use of `Cypress.Promise.all`, removed `http://localhost:3000/` in favor of `cypress.json` configuration, and simplified logic around unnecessary `visit`s and `wait`s.

PR #556, opened 2021-02-19
feat(frontend): Refactor Link and Term Component

icerove:
Refactor Link and Term Component 
Currently, Mixpanel is delayed by some reason but the refactor for this PR I think make sense to review and merge. 
What do you think? @frol


icerove:
@frol the unit test checking is weird and could you mind have a look?

frontend/mixpanel/index.ts
```diff
@@ -0,0 +1,31 @@
+import mixpanel from "mixpanel-browser";
+
```
frol, line 2: Let's put this file into `frontend/src/libraries/mixpanel.ts`

frontend/src/components/utils/Footer.jsx
```diff
@@ -1,154 +1,176 @@
...
-        width: 368px;
-        height: 92px;
-      }
-
-      @media (max-width: 780px) {
+      </div>
+      <div className="nearkat-wrapper">
+        <a
+          className="footer-help-link"
+          href="http://near.chat"
```
frol, line 49: We lost `target="_blank"` here. Please, add `target="_blank"` and `rel="noopener"` ([read more](https://mathiasbynens.github.io/rel-noopener/)) to the external links

frontend/src/components/utils/Link.tsx
```diff
@@ -0,0 +1,21 @@
...
+  href: string;
+  as?: string;
+  children: React.ReactNode;
+}
+
+export default ({ href, as, children }: Props) => {
+  const clicked = () =>
+    Mixpanel.track("Click Link", { href: href, as: as ? as : "" });
+  return (
+    <div onClick={clicked}>
```
frol, line 15: `div` is a block-displaying element, which breaks text when a link is inline, e.g. the text message on the [transaction row](https://near-explorer-frontend-pr-556.onrender.com/transactions):

image

Consider using `span` instead

frontend/src/components/utils/Term.tsx
```diff
@@ -45,7 +49,17 @@ export default class extends React.Component<Props, State> {
             <Modal.Header closeButton>
               <Modal.Title>{title}</Modal.Title>
             </Modal.Header>
-            <Modal.Body>{children}</Modal.Body>
+            <Modal.Body>
+              {text}
+              {href && (
+                <a
+                  href={href}
```
frol, line 56: reminder about the `target="_blank" rel="noopener"`

frontend/src/components/utils/WalletLink.tsx
```diff
@@ -1,15 +1,20 @@
...
+  const clicked = () =>
+    Mixpanel.track("Click for wallet profile", {
+      accountId: accountId,
+      walletPrefix: nearWalletProfilePrefix,
+    });
   return (
-    <>
+    <div onClick={clicked}>
       <a
-        target="_blank"
```
frol, line 12: Why? The intention was to open a new tab

frontend/src/components/contracts/ContractDetails.tsx
```diff
@@ -122,13 +120,15 @@ export default class extends React.Component<Props, State> {
...
+                  <Term
+                    title={"Locked?"}
+                    text={`
+                      Locked contract means that there are no access keys allowing the
                       contract code to be re-deployed (e.g. even a single FullAccess
-                      permission access key casts the locked status to "No").`}</p>
-                    <p>{`Note: The contract itself may be implemented to re-deploy itself
-                      or re-add FullAccess keys.`}</p>
-                  </Term>
+                      permission access key casts the locked status to "No"). \n
```
frol, line 128: We lost the paragraphs here

image

frol (changes requested):

frol:
@icerove How do we distinguish mainnet vs testnet usage?

icerove:
> @icerove How do we distinguish mainnet vs testnet usage?

The url of the website will distinguish already for mainnet and testnet

frol:
> the unit test checking is weird and could you mind have a look?

You have the following unguarded code in the `mixpanel/index.ts`:

```js
mixpanel.init(BROWSER_MIXPANEL_TOKEN);
mixpanel.register({ timestamp: new Date().toString() });
```

This gets triggered on every import of the file. `Link` imports this file, and this ends up blocking things, It should only be initialized when executed in a browser context.

icerove:
> This gets triggered on every import of the file. Link imports this file, and this ends up blocking things, It should only be initialized when executed in a browser context.

How should I fix this?

frol:
@icerove try wrapping it with `if (typeof window !== 'undefined')`

frol:
image

Please, don't update styles. Let's keep this change on topic.

frontend/cypress/integration/nodes-page.spec.js
```diff
@@ -9,14 +9,13 @@ context("Nodes", () => {
...
-    cy.get("#online-node").find(".node-selector").contains("Online-nodes");
-    cy.get("#proposal-node").find(".node-selector").contains("Proposal-nodes");
+    cy.get("#validator-node").contains("Validating");
+    cy.get("#online-node").contains("Online-nodes");
+    cy.get("#proposal-node").contains("Proposal-nodes");
   });
 
   it("Check validators tab", () => {
     cy.url().should("include", "/nodes/validators");
-    cy.get("#validator-node").should("have.class", "node-selected");
```
frol, line 19: Can we update the test instead of just removing it?

icerove:
> @icerove try wrapping it with if (typeof window !== 'undefined')
I tried, but the e2e test not pass this time since error for `prop undefined`

frontend/cypress/integration/nodes-page.spec.js
icerove: sure

frol:
> I tried, but the e2e test not pass this time since error for prop undefined

Then you should dig into it and find why the prop is undefined and either define it or avoid calling the function.

frontend/src/components/utils/WalletLink.tsx
```diff
@@ -1,13 +1,19 @@
...
 
 const WalletLink = ({ accountId, nearWalletProfilePrefix }: Props) => {
+  const clicked = () =>
+    Mixpanel.track("Click for wallet profile", {
+      accountId: accountId,
+      walletPrefix: nearWalletProfilePrefix,
+    });
   return (
-    <>
+    <div onClick={clicked}>
```
frol, line 16: I think it should be possible to apply `onClick` directly on `a` tag.

frontend/src/libraries/mixpanel.ts
```diff
@@ -0,0 +1,48 @@
+import mixpanel from "mixpanel-browser";
+
+const BROWSER_MIXPANEL_TOKEN = "7cc6cbaab18d00de1b06ca9b4e773ed7";
+
+if (process.env.NEAR_EXPLORER_DATA_SOURCE) {
```
frol, line 5: Since `process` is not available in a browser context, I believe this is going to be never `true` on the browser side.

frontend/src/libraries/mixpanel.ts
```diff
@@ -0,0 +1,48 @@
...
+  people: {
+    set_once: (props: object) => {
+      console.log(props);
+    },
+  },
+};
+
+const Mixpanel = process.env.NEAR_EXPLORER_DATA_SOURCE
+  ? MixpanelReal
+  : MixpanelMock;
```
frol, line 46: The same concern about browser context.

frontend/src/pages/accounts/[id].jsx
```diff
@@ -32,7 +34,7 @@ export default class extends React.Component {
 
   render() {
     const { account, accountFetchingError, currentNearNetwork } = this.props;
-
+    Mixpanel.track("View Individual Account", { accountId: account.accountId });
```
Comment (frol) [inline frontend/src/pages/accounts/[id].jsx:37]: Should we move it to `componentDidMount`?

frol (changes requested):

frontend/src/components/utils/WalletLink.tsx
icerove: If I add `onClick` on a , `target="_blank" rel="noopener"` will be broken

frontend/src/pages/accounts/[id].jsx
Comment (icerove) [reply frontend/src/pages/accounts/[id].jsx:37]: No, I test it and it works fine?

frontend/src/libraries/mixpanel.ts
icerove: we have this variable on the render environment file and this works for tracking

frontend/src/pages/accounts/[id].jsx
Comment (frol) [reply frontend/src/pages/accounts/[id].jsx:37]: Why do we need to call it on every render event?

frontend/package.json
```diff
@@ -62,9 +62,9 @@
     "echarts": "^4.9.0",
     "echarts-for-react": "^2.0.16",
     "fast-deep-equal": "^3.1.3",
-    "hexy": "^0.3.1",
-    "lodash": "^4.17.21",
-    "moment": "^2.29.1",
+    "hexy": "^0.3.0",
+    "lodash": "^4.17.15",
+    "moment": "^2.26.0",
```
frol, line 67: `master` has newer versions, let's keep them.

frontend/src/components/utils/WalletLink.tsx
```diff
@@ -7,7 +7,7 @@ export interface Props {
 
 const WalletLink = ({ accountId, nearWalletProfilePrefix }: Props) => {
   return (
-    <>
+    <div>
```
frol, line 10: ^

frontend/src/components/utils/Link.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export default ({ href, as, children }: Props) => {
+  return (
+    <span>
+      <Link href={href} as={as}>
+        {children}
+      </Link>
+    </span>
+  );
+};
```
frol, line 18: ^

frol (approved):

PR #574, opened 2021-03-04
feat(frontend): Expose receipts on Transaction Details page

### LINKED ISSUE #512: Expose receipts on Transaction Details page
## Story

As a regular user or developer who browses transactions, I want to understand what exactly happens during a transaction execution so I can build a mental model and better understand the flow.

## Acceptance Criteria

- [ ] Transaction Details page lists receipts and their execution outcomes
- [ ] Mobile ok
- [ ] The dumb (render-only) components are unit-tested

## Notes

- A transaction ALWAYS has at least one receipt (which has the same actions list, the sender account id, and the receiver account id)
- A receipt may have 0 or more followup (outgoing) receipts
- A receipt has exactly one execution outcome
- Some receipts may fail (execution outcome status is `FAILURE`)

shelegdmitriy:
Resolves #512 

Test plan

- [x] Existing tests pass;
- [x] ReceiptRow unit test covering a transaction with all successful receipts pass;
- [x] ReceiptRow unit test covering a transaction with nested receipts pass;
- [x] ReceiptRow unit test covering a transaction with failing receipts pass;
- [x] Manually confirmed the layout on desktop;
- [x] Manually confirmed the layout on mobile


icerove:
Hi @shelegdmitriy 
Great work. We are already on the board with you with this PR for design. 
For the sake, would you mind give more information about the receipts list.
For design research, we always provide list of information for each row and let Corwin better understanding the structure.
Even better is to provide screen shot and easy to have a look since render do not provide this anymore.

shelegdmitriy:
> Hi @shelegdmitriy
> Great work. We are already on the board with you with this PR for design.
> For the sake, would you mind give more information about the receipts list.
> For design research, we always provide list of information for each row and let Corwin better understanding the structure.
> Even better is to provide screen shot and easy to have a look since render do not provide this anymore.

Hi @icerove. Thanks, but Vlad already explained it to me and I'm finishing with the new version that includes this notes.

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -435,6 +454,12 @@ export default class TransactionsApi extends ExplorerApi {
           transactionInfo.hash,
           transactionInfo.signerId,
         ]);
+
+        const transactionReceiptsList = await this.call<any>(
```
icerove, line 458: I do not know whether Frol tell you this but the pattern for this query is actually a little different from current one. We query the list of transactions and outcome for the transactions directly here instead writing on the backend and expose wamp method, please make sure this with Frol to see what is the way he wants

backend/src/db-utils.js
```diff
@@ -240,6 +240,24 @@ const queryDashboardTransactionsStats = async (options) => {
   };
 };
 
+const queryReceiptsList = async (transactionHash) => {
```
icerove, line 243: leave the comment at the frontend/src/libraries/explorer-wamp/transactions.ts file

frontend/src/libraries/explorer-wamp/transactions.ts
frol: @icerove thanks for noticing it. Indeed, it is not following the current pattern, but since we switched from our legacy backend, and have more room on the backend now, I decided we go forward with gradually moving the queries to the backend and eventually forbid the arbitrary `query` requests. Thus, this is our path forward.

frontend/src/components/transactions/ReceiptsOutcomeList.tsx
```diff
@@ -0,0 +1,38 @@
+import * as T from "../../libraries/explorer-wamp/transactions";
+import ReceiptOutcomeRow from "./ReceiptOutcomeRow";
+export interface Props {
+  receipts: T.ReceiptsOutcomeList;
+}
+
+export default ({ receipts }: Props) => {
```
icerove, line 7: please follow new pattern to export component not getting next js warning again

frontend/src/components/transactions/ReceiptOutcomeRow.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+export interface Props {
+  sirReceiptHash: string;
+  receiptsObj: {};
+  receiptsInfoObj: {};
+  senderIsReceiverActions: T.RpcAction[];
+  senderIsReceiver: Boolean;
+}
+
+export default class extends React.Component<Props> {
```
icerove, line 16: follow the name pattern and export without next js warning

frontend/src/components/transactions/ReceiptOutcomeRow.tsx
```diff
@@ -0,0 +1,260 @@
...
+  handleStatusInfo = function (receipt: T.ReceiptLocalOutcome) {
+    let statusInfo;
```
frol, line 18: I feel like you don't need this variable, whenever you assign a value to it in this function, you may just as well return it immediately

frontend/src/components/transactions/ReceiptOutcomeRow.tsx
```diff
@@ -0,0 +1,260 @@
...
+  ) {
+    if (
+      receiptObjInfo?.receipt &&
+      (!senderIsReceiverActions || senderIsReceiverActions.length === 0)
+    ) {
+      const { Action: action = null } = receiptObjInfo?.receipt;
+
+      if (action && action.actions && action.actions.length > 0) {
+        const actionsList = action.actions.map((action, index) => (
+          <Row noGutters key={`${action[Object.keys(action)[0]]}_${index}`}>
```
frol, line 68: Let's extract `Object.keys(action)[0]` into a variable to avoid casting this magic over and over again. I even suggest we preprocess the actions as we do for the [transaction actions](https://github.com/near/near-explorer/blob/35c49e500d6662e70c94d4bab40d3ef72c7ca04d/frontend/src/libraries/explorer-wamp/transactions.ts#L443-L453)

P.S. We should address this weirdness on the JSON API side at some time in the future.

frontend/src/components/transactions/ReceiptsOutcomeList.tsx
```diff
@@ -0,0 +1,38 @@
...
+      [receiptInfoItem.receipt_id]: receiptInfoItem,
+    }),
+    {}
+  );
+
+  return (
+    <ReceiptOutcomeRow
+      sirReceiptHash={converted_into_receipt.receipt_id}
+      receiptsObj={receiptsOutcomeObj}
+      receiptsInfoObj={receiptsInfoObj}
```
frol, line 33: `Obj` is not a great naming since all the variables in JS are objects. I suggest we call these `receiptOutcomesById` (both in props and in local variables here)

frontend/src/components/transactions/ReceiptsOutcomeList.tsx
```diff
@@ -0,0 +1,38 @@
+import * as T from "../../libraries/explorer-wamp/transactions";
+import ReceiptOutcomeRow from "./ReceiptOutcomeRow";
```
frol, line 2: It seems that we display the full Receipt information, not only outcomes, so let's name these components as `ReceiptsList`, `ReceiptRow`, etc (leave out "outcome" unless you feel it reflects the situation.

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -105,6 +109,29 @@ export interface ReceiptOutcome {
...
   receiptsOutcome?: ReceiptOutcome[];
 }
+export interface ReceiptOutcomeInfo {
+  predecessor_id: string;
+  receiver_id: string;
+  receipt_id: string;
+  receipt?: ReceiptOutcomeLocalAction;
+}
+
+export interface ReceiptLocalOutcome {
```
frol, line 119: I don't think we need to have a separate "local" entity. We should be able to construct a local receipt from a transaction and transaction outcome (the receipt ID goes from there).

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -435,24 +463,36 @@ export default class TransactionsApi extends ExplorerApi {
...
         transactionInfo.receiptsOutcome = transactionExtraInfo.receipts_outcome as ReceiptOutcome[];
+        transactionInfo.receiptsList = {
+          receipts: transactionExtraInfo.receipts as ReceiptOutcomeInfo[],
+          receipts_outcome: transactionExtraInfo.receipts_outcome as ReceiptOutcome[],
+          converted_into_receipt: {
+            actions: actions,
+            block_hash: transactionExtraInfo.receipts_outcome[0].block_hash,
+            outcome: transactionExtraInfo.receipts_outcome[0].outcome,
+            receipt_id: transactionExtraInfo.receipts_outcome[0].id,
+          } as ReceiptLocalOutcome,
```
frol, line 494: Let's just push the converted receipt directly into the `receipts` and `receipts_outcome`, and be done with it.

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -435,24 +463,36 @@ export default class TransactionsApi extends ExplorerApi {
...
+            }
           }
-        });
-
+        );
+        transactionInfo.actions = actions;
         transactionInfo.receiptsOutcome = transactionExtraInfo.receipts_outcome as ReceiptOutcome[];
+        transactionInfo.receiptsList = {
+          receipts: transactionExtraInfo.receipts as ReceiptOutcomeInfo[],
+          receipts_outcome: transactionExtraInfo.receipts_outcome as ReceiptOutcome[],
```
frol, line 488: `receiptsList` feels unnecessary (it sticks out immediately to me due to its naming which does not reflect the object that it defines)

frol (changes requested):

frol (changes requested):
Bug report from @icerove:

image

image

The UI says that the transfer goes to berryclub.ek.near instead of mg88.near

frontend/src/components/transactions/__tests__/common.tsx
```diff
@@ -189,3 +337,194 @@ export const TRANSACTIONS: T.Transaction[] = [
     },
   },
 ];
+
+export const RECEIPT_OUTCOMES_BY_ID = [
+  // for those no action has deposit
```
frol, line 342: Instead of having a list of groups of receipt outcomes leading to quite a mysterious usage of `RECEIPT_OUTCOMES_BY_ID[0]` (`0` is not an id, and I was quite confused), I suggest you have separate constants `RECEIPT_OUTCOMES_BY_ID_WITH_NO_DEPOSIT_ACTIONS`, `RECEIPT_OUTCOMES_BY_ID_WITH_SMALL_DEPOSIT_ACTIONS`, ...

frontend/src/components/transactions/ReceiptRow.tsx
```diff
@@ -48,45 +66,229 @@ class ReceiptRow extends React.Component<Props> {
...
-                  {`${receipt.id.substr(0, 7)}...`}
+                  {tokensBurnedByConvertation ? (
+                    <Balance amount={tokensBurnedByConvertation.toString()} />
+                  ) : (
+                    "..."
+                  )}
                 </Col>
               </Row>
+            </>
+          )}
```
frol, line 132: This should be moved to a separate component (`TransactionOutcome`) and just used directly from `frontend/src/pages/transactions/[hash].jsx`, so there is no need to have `convertedReceipt` in this component.

frol (changes requested):

frontend/src/components/transactions/ReceiptsOutcomeList.tsx
shelegdmitriy: @icerove thanks, I fixed it

frontend/src/components/transactions/ReceiptOutcomeRow.tsx
shelegdmitriy: @icerove , and this one too

frontend/src/pages/transactions/[hash].jsx
```diff
@@ -55,13 +57,16 @@ class TransactionDetailsPage extends React.Component {
             />
           </Content>
         )}
-        {this.props.receiptsOutcome && (
+
+        {receiptsOutcome && (
```
Comment (frol) [inline frontend/src/pages/transactions/[hash].jsx:61]: You should check for `this.props.receipts`, otherwise, it looks odd since you don't use `receiptsOutcome`

frontend/src/components/transactions/ReceiptRow.tsx
```diff
@@ -1,18 +1,24 @@
...
 }
 
 class ReceiptRow extends React.Component<Props> {
-  render() {
-    const { receipt } = this.props;
-
+  renderRow = (
+    receipt: T.ExecutionOutcomeReceipts,
+    convertedReceipt = true
+  ) => {
```
frol, line 21: There is no need to make it a method. Just put the implementation right in the `render` function and replace `this.renderRow()` with `<ReceiptRow receipt={executedReceipt} />`.

frontend/src/components/transactions/ReceiptRow.tsx
```diff
@@ -1,18 +1,24 @@
...
 import { Row, Col } from "react-bootstrap";
 
-import * as T from "../../libraries/explorer-wamp/transactions";
+import ActionRow from "./ActionRow";
 
+import * as T from "../../libraries/explorer-wamp/transactions";
+import Gas from "../utils/Gas";
+import Balance from "../utils/Balance";
+import { truncateAccountId } from "../../libraries/formatting";
 import { displayArgs } from "./ActionMessage";
```
frol, line 11: Moved the first line to the last line and merged it with the group according to the CONTRIBUTING.md:

```suggestion
import * as T from "../../libraries/explorer-wamp/transactions";
import Gas from "../utils/Gas";
import Balance from "../utils/Balance";
import { truncateAccountId } from "../../libraries/formatting";
import { displayArgs } from "./ActionMessage";
import ActionRow from "./ActionRow";
```

frontend/src/components/transactions/ReceiptRow.tsx
```diff
@@ -48,45 +54,152 @@ class ReceiptRow extends React.Component<Props> {
...
-                    "No logs"
-                  ) : (
-                    <pre>{receipt.outcome.logs.join("\n")}</pre>
-                  )}
-                </Col>
-              </Row>
+      <Row
+        noGutters
+        className={
+          !convertedReceipt ? "receipt-row pl-4 mx-0" : "receipt-converted-row"
```
frol, line 68: I believe there is no need in the `convertedReceipt` flag anymore.

frol (changes requested):

frol:
@corwinharrell This new detailed transaction execution plan needs your beautiful design touch!

We put everything we need to display there in terms of data, so you can have a better idea of what needs to be fit in there. We currently put the structure in a nested tree view (and most of the transactions if not all, use it this way), but in the essence of NEAR Protocol, it could be an acyclic graph where a single receipt may produce three receipts, where the third will be awaiting for the first two. Currently, it is going to look like:

```
- Receipt 1
|- Receipt 2
|- Receipt 3
|- Receipt 4 (awaited for Receipt 2 and Receipt 3)
```

Receipt 4 was created by Receipt 1, that is correct, but we could also represent it as:

```
            ---- Receipt 2 ----
           /                    \
Receipt 1    ------------------   Receipt 4
           \                    /
            ---- Receipt 3 ----
```

I think this could be a separate initiative as just an alternative representation, and we should stick to the tree-view.

Currently, we display all the receipts the same way, and receipts are all equal, but from a user point of view, there are some special receipts which cost 0 gas and sent to them back from `system` (non-existing) account, which is a refund system in NEAR Protocol (each execution leaf ends with a refund).

Some complex transactions for your reference:

* many receipts:
  * https://near-explorer-frontend-with-indexer-pr-dl95.onrender.com/transactions/FBxPnRgJ7mDbv9nycV7TVgDwhSMLMSALecArs9sb52Em
  * https://near-explorer-frontend-with-indexer-pr-dl95.onrender.com/transactions/FJNg2yQKrkXbUp23pys3Beid4aAAKf9ZfEAMJziUz85i
* big list of args: https://near-explorer-frontend-with-indexer-pr-dl95.onrender.com/transactions/4H2SUehL5q6kDKZs3EvqxiqF4JBQ7QP7VgAQrkbCZJjn
* failing: https://near-explorer-frontend-with-indexer-pr-dl95.onrender.com/transactions/HpHfTgaQsy1AwaKMUQsUH2Eh6UhUuyVf55sfVwqoJiCd
* small: https://near-explorer-frontend-with-indexer-pr-dl95.onrender.com/transactions/BFXM5b3N6UhMt2eD1dMH8N2F3AaCTipK85cwJQp3wbvd
* just a transfer (with a refund receipt): https://near-explorer-frontend-with-indexer-pr-dl95.onrender.com/transactions/48GAZXFuumPAHsMpqVMsorVhYv2M5mxw3EF3YTLUfC4r
* bridge uses binary format instead of JSON, so the args are presented in hex (not really readable as of now :confused:): https://near-explorer-frontend-with-indexer-pr-dl95.onrender.com/transactions/8GNYK2KTXZWnV12jqjzQcYPgVAHoY28kcNP8GkwnsXiL

frol (approved):
The implementation looks good to me. Let's wait for a design for this page and meanwhile switch to the Nodes page redesign task.

frol:
Well, on second thought, the current receipts list is just completely useless, so I think we should merge it as it is now just for the sake of exposing more data and improve the design as a separate PR.

frontend/src/components/transactions/TransactionOutcome.tsx
```diff
@@ -0,0 +1,76 @@
...
+export interface Props {
+  transaction: T.TransactionOutcome;
+}
+
+class TransactionOutcome extends React.Component<Props> {
+  render() {
+    const { transaction } = this.props;
+    const gasBurnt = transaction?.outcome
+      ? new BN(transaction.outcome.gas_burnt)
+      : new BN(0);
```
frol, line 18: Would this work? (also, it seems that `transaction` cannot be undefined or null, so `?` after it is not necessary.

```suggestion
    const gasBurnt = new BN(transaction.outcome?.gas_burnt ?? 0);
```

If it does, update the `tokensBurnt` and other places (if any) as well.

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -105,6 +105,33 @@ export interface ReceiptOutcome {
 export interface ReceiptsOutcomeWrapper {
   receiptsOutcome?: ReceiptOutcome[];
 }
+export interface ReceiptsList {
```
frol, line 108: This type is not a list, so should not be called with "List" suffix. It seems that it is an internal interface just to provide a type to the response we get from the backend, so I suggest we don't `export` it, and call something like `RpcReceipt`.

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -105,6 +105,33 @@ export interface ReceiptOutcome {
...
+  predecessor_id: string;
+  receiver_id: string;
+  receipt_id: string;
+  receipt?: any;
+  actions?: Action[];
+}
+export interface ExecutionOutcomeReceipts {
```
frol, line 115:
```suggestion
export interface NestedReceiptWithOutcome {
```

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -105,6 +105,33 @@ export interface ReceiptOutcome {
...
+  actions?: Action[];
+  block_hash: string;
+  outcome: ReceiptOutcomeReceipts;
+  predecessor_id: string;
+  receipt_id: string;
+  receiver_id: string;
+}
+
+export interface ReceiptOutcomeReceipts {
```
frol, line 124:
```suggestion
export interface ReceiptExecutionOutcome {
```

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -105,6 +105,33 @@ export interface ReceiptOutcome {
...
+  tokens_burnt: string;
+  logs: string[];
+  receipt_ids?: ExecutionOutcomeReceipts[];
```
frol, line 127:
```suggestion
  outgoing_receipts?: NestedReceiptWithOutcome[];
```

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -435,24 +463,90 @@ export default class TransactionsApi extends ExplorerApi {
...
+                  const kind = Object.keys(action)[0] as keyof RpcAction;
+                  return {
+                    kind,
+                    args: action[kind],
+                  };
+                }
+              }
+            );
           }
+          receiptsByIdMap.set(receiptItem.receipt_id, [receiptItem]);
```
frol, line 523:
```suggestion
          receiptsByIdMap.set(receiptItem.receipt_id, receiptItem);
```

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -435,24 +463,90 @@ export default class TransactionsApi extends ExplorerApi {
...
+            outcome: {
+              ...receipts.outcome,
+              receipt_ids:
+                receipts.outcome && receipts.outcome.receipt_ids.length > 0
+                  ? receipts.outcome.receipt_ids.map(
+                      (executedReceipt: string) =>
+                        executionOutcomeReceipt(executedReceipt)
+                    )
+                  : [],
+            },
```
frol, line 541: `receiptOutcome.outcome` cannot be missing, so the following should work fine:

```suggestion
          const receipt = receiptsByIdMap.get(receiptHash);
          const receiptOutcome = receiptOutcomesByIdMap.get(receiptHash);
          return {
            ...receipt,
            ...receiptOutcome,
            outcome: {
              ...receiptOutcome.outcome,
              outgoing_receipts:
                receiptOutcome.outcome.receipt_ids.map(
                      (outgoingReceiptId: string) =>
                        executionOutcomeReceipt(outgoingReceiptId)
                    ),
            },
```

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -435,24 +463,90 @@ export default class TransactionsApi extends ExplorerApi {
...
+          };
+        };
+
+        transactionInfo.actions = actions;
+        transactionInfo.receiptsOutcome = receiptsOutcome;
```
frol, line 546: It seems that we don't need them (update the type accordingly):

```suggestion
```

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -435,24 +463,90 @@ export default class TransactionsApi extends ExplorerApi {
...
+                  };
+                }
+              }
+            );
           }
+          receiptsByIdMap.set(receiptItem.receipt_id, [receiptItem]);
         });
 
-        transactionInfo.receiptsOutcome = transactionExtraInfo.receipts_outcome as ReceiptOutcome[];
+        const executionOutcomeReceipt = (receiptHash: string) => {
```
frol, line 526:
```suggestion
        const collectNestedReceiptWithOutcome = (receiptHash: string) => {
```

frontend/src/components/transactions/__tests__/ReceiptRow.test.tsx
```diff
@@ -31,8 +31,8 @@ describe("<ReceiptRow />", () => {
     expect(
       renderer.create(
         <ReceiptRow
-          receipt={(TRANSACTIONS[1].receiptsOutcome || [])[0]}
-          key={(TRANSACTIONS[1].receiptsOutcome || [])[0].id || "ccc"}
+          receipt={TRANSACTIONS[2].receipt!}
```
frol, line 34: I realized that `TRANSACTIONS[2]` has no failing receipts, which is what this test title states. I suggest we split `TRANSACTIONS` into independent constants (e.g. `TRANSACTION_WITH_FAILING_RECEIPTS`, `TRANSACTION_WITH_HUGE_AMOUNTS`, etc) since otherwise, it is easy to introduce a completely inadequate test.

The test was clearly broken before, but I would like to ask you to fix this situation while we are at it.

frol (changes requested):
I was about to merge it, but suddenly looked into the test suite and found a few discrepancies, which brought me down the line back to the types and naming...

frontend/src/libraries/explorer-wamp/transactions.ts
shelegdmitriy: @frol no, we need as we use them in `TransactionDetails.tsx`

PR #575, opened 2021-03-05
feat(backend): added new stats for weekly active accounts, daily deposit volume, unique users by application

icerove:
1. weekly active accounts
2. aggregate total volume deposit by day
3. partner account list with unique user amount


icerove:
let's wait for adding indexes on action_kind and status. @frol @khorolets

khorolets:
Indexes have been added here https://github.com/near/near-indexer-for-explorer/pull/72 meanwhile migration is already applied

frontend/src/pages/api/stats/total-deposit-amount.js
```diff
@@ -2,9 +2,16 @@ import StatsApi from "../../../libraries/explorer-wamp/stats";
 
 export default async function (req, res) {
   try {
-    const totalDepositAmount = await new StatsApi(req).totalDepositAmount();
-    if (totalDepositAmount) {
-      res.send(totalDepositAmount);
+    const depositAggregatedByDate = await new StatsApi(
```
frol, line 5: Since this is not `.csv`-prefixed variant of the handler, we should just return JSON. Also, we may just drop the method altogether since I only created one when we did not have charts.

backend/src/db-utils.js
```diff
@@ -376,6 +400,27 @@ const queryActiveAccountsList = async () => {
...
+const queryDepositAmountAggregatedByDate = async () => {
+  return await queryRows(
+    [
+      `SELECT 
+          TIMESTAMP 'epoch' + DIV(DIV(receipts.included_in_block_timestamp, 1000000000), 60 * 60 * 24) * INTERVAL '1 day' AS "date",	
+          CAST(SUM((action_receipt_actions.args->>'deposit')::numeric) AS TEXT) AS total_deposit_amount
+        FROM action_receipt_actions
+        JOIN execution_outcomes ON execution_outcomes.receipt_id = action_receipt_actions.receipt_id
+        JOIN receipts ON receipts.receipt_id = action_receipt_actions.receipt_id
+        WHERE action_receipt_actions.action_kind IN ('FUNCTION_CALL', 'TRANSFER')
```
frol, line 412: We should filter out `receipts.predecessor_account_id = 'system'`, so we don't account for transaction refunds.

frol (approved):
@icerove it seems that the queries got sped up, though `queryDepositAmountAggregatedByDate` still takes almost 5 minutes. Please, address my other comments, make sure you test it locally, merge it to master, and confirm that it works correctly on production.

frontend/src/pages/api/stats/total-deposit-amount.js
icerove: will drop the folder.

backend/src/db-utils.js
icerove: @frol 
what is the transaction for predecessor accountId is `system`??? I do not clear that

icerove:
I changed the query somehow and tested it locally. for `queryDepositAmountAggregatedByDate`, it takes 3 minites to finish for now. @frol

backend/src/db-utils.js
```diff
@@ -84,14 +85,14 @@ const queryOnlineNodes = async () => {
...
-//new query for new dashboard
+// query for new dashboard
 const queryDashboardBlocksStats = async (options) => {
   async function queryLatestBlockHeight({ dataSource }) {
     let query;
     if (dataSource === DS_INDEXER_BACKEND) {
       query = `SELECT block_height FROM blocks ORDER BY block_height DESC LIMIT 1`;
     } else {
-      query = `SELECT height AS block_height FROM blocks ORDER BY height DESC LIMIT 1`;
+      query = `SELECT height as block_height FROM blocks ORDER BY height DESC LIMIT 1`;
```
frol, line 95: `AS` is an SQL keyword, so given we opt into using uppercase for all the SQL keywords, let's keep `AS` in uppercase, please.

frol (approved):

Issue #568, opened 2021-03-01
Number of unique contracts


ilblackdragon:
Currently we should number of accounts with contracts and show how many are active.

Given we highly frequently leverage the factory pattern, where the same contract is deployed on many accounts (like staking pools, lockup, multisigs, etc). 

This doesn't allow to understand how many unique contracts does NEAR have.
We can just count dedupped on contract hash to show number of unique contracts on-chain.

This is especially interesting to compare between testnet and mainnet, seeing how much development gets converted into mainnet.

frol:
We can use the `DEPLOY_CONTRACT` action to get the contract binary hash:

```sql
SELECT * FROM action_receipt_actions WHERE action_kind = 'DEPLOY_CONTRACT' LIMIT 1;
```

![image](https://user-images.githubusercontent.com/304265/109519790-55e3cc80-7a8a-11eb-97f4-cb0314b340dc.png)

There are currently 4868 unique `DEPLOY_CONTRACT` actions on testnet, and 117 on mainnet (over the whole history):

```sql
SELECT COUNT(DISTINCT args->>'code_sha256') FROM action_receipt_actions WHERE action_kind = 'DEPLOY_CONTRACT';
```

If we want to have it in a chart over time, we need to use something like this (pseudocode):

```sql
SELECT
    day,
    (SELECT COUNT(DISTINCT args->>'code_sha256')
       FROM action_receipt_actions
       WHERE  receipts.included_in_block_timestamp <= g.day
    ) AS dist_contract_deploys
FROM  (SELECT generate_series('2021-01-01'::date, '2021-01-06'::date, '1d')::date) AS g(day))
```

Basically, we will need to compute the unique value for each date. Alternatively, we may consider to pull all the `DEPLOY_CONTRACT` actions and compute the computation on Explorer backend side keeping the uniqueness accounting rolling.

frol:
We can pull all the `DEPLOY_CONTRACT` events and aggregate them (iterate) on the Explorer backend (and cache it):

```sql
SELECT * FROM action_receipt_actions WHERE action_kind = 'DEPLOY_CONTRACT' ORDER BY included_in_block_timestamp;
```

icerove:
from my understanding, for this unique contracts, we should separate it into two parts:
1. display daily unique deployed contracts amount (even though the contract can be deployed previously, at that day, it is unique deployed)
[...]

PR #577, opened 2021-03-08
feat(backend): aggregate and expose unique contracts deployed count by date

icerove:
add query on backend for unique deployed contracts by day


frontend/src/pages/api/stats/unique-contracts-by-date.csv.js
```diff
@@ -0,0 +1,19 @@
+import StatsApi from "../../../libraries/explorer-wamp/stats";
```
frol, line 1: I think we don't need these CSV wrappers anymore. I only created them when we didn't have the charts, so we could build the charts in excel

frol (changes requested):
See https://github.com/near/near-explorer/issues/568#issuecomment-794438568

backend/src/db-utils.js
```diff
@@ -305,6 +305,22 @@ const queryNewContractsCountAggregatedByDate = async () => {
   );
 };
 
+const queryUniqueContractsAggregatedByDate = async () => {
```
frol, line 308: Well, it is not unique contracts, but unique contract deployments. Unique contracts are the accounts. We can re-deploy different contracts to the same account, and this is what the query returns - the deployments.

backend/src/stats.js
```diff
@@ -80,6 +83,43 @@ async function aggregateNewContractsCountByDate() {
       })
     );
     console.log("NEW_CONTRACTS_COUNT_AGGREGATED_BY_DATE updated.");
+
+    // unique contract part
```
frol, line 87: Please, separate it into a dedicated function. Also, please take my above note into account for the new function name.

frontend/src/libraries/explorer-wamp/stats.ts
```diff
@@ -59,6 +59,12 @@ export default class StatsApi extends ExplorerApi {
     );
   }
 
+  async uniqueContractsCountAggregatedByDate(): Promise<ContractsByDate[]> {
+    return await this.call<ContractsByDate[]>(
+      "unique-contracts-count-aggregate-by-date"
```
frol, line 64: ^

frol (changes requested):

backend/src/stats.js
icerove: rename it but it cannot be separate into a dedicated function since I need the array of date from new contracts count by date

backend/src/stats.js
```diff
@@ -80,6 +83,43 @@ async function aggregateNewContractsCountByDate() {
...
+          );
+        }
+      }
+      if (j > 0) {
+        cumulativeContractsByDate[j].cumulativeContracts = new Set([
+          ...cumulativeContractsByDate[j - 1].cumulativeContracts,
+          ...cumulativeContractsByDate[j].cumulativeContracts,
+        ]);
+      }
+    }
```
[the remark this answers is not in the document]
icerove: no very clear what do you mean but I change it a little bit.

frol (approved):

PR #595, opened 2021-04-01
feat(backend): exposed a method to query token holders of bridge contracts

icerove:
the query is for bridge tracking :
for each token, who is mint and hold this token

### DIFF
--- a/backend/src/db-utils.js
+++ b/backend/src/db-utils.js
@@ -529,17 +529,17 @@ const queryPartnerUniqueUserAmount = async () => {
 };
 
 // bridge query
-const queryBridgeUser = async (accountId) => {
+const queryBridgeTokenHolder = async (accountId) => {
   return await queryRows(
     [
-      `select 
-        distinct transactions.signer_account_id as holder
-      from receipts 
-      join action_receipt_actions on receipts.receipt_id = action_receipt_actions.receipt_id  
-      join transactions on transactions.transaction_hash = receipts.originated_from_transaction_hash 
-      where receipts.receiver_account_id = :account_id
-      and action_receipt_actions.action_kind = 'FUNCTION_CALL' 
-      and action_receipt_actions.args->> 'method_name' = 'mint'`,
+      `SELECT 
+        DISTINCT transactions.signer_account_id AS holder
+      FROM receipts 
+      JOIN action_receipt_actions ON receipts.receipt_id = action_receipt_actions.receipt_id  
+      JOIN transactions ON transactions.transaction_hash = receipts.originated_from_transaction_hash 
+      WHERE receipts.receiver_account_id = :account_id
+      AND action_receipt_actions.action_kind = 'FUNCTION_CALL' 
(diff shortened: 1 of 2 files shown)


backend/src/db-utils.js
```diff
@@ -528,6 +528,24 @@ const queryPartnerUniqueUserAmount = async () => {
   );
 };
 
+// bridge query
+const queryBridgeUser = async (accountId) => {
```
frol, line 532: You lost me here with the naming. To query a bridge user you need to do `SELECT * FROM accounts WHERE account_id = ?`. The implementation clearly does not match it.

backend/src/db-utils.js
```diff
@@ -528,6 +528,24 @@ const queryPartnerUniqueUserAmount = async () => {
...
+  return await queryRows(
+    [
+      `select 
+        distinct transactions.signer_account_id as holder
+      from receipts 
+      join action_receipt_actions on receipts.receipt_id = action_receipt_actions.receipt_id  
+      join transactions on transactions.transaction_hash = receipts.originated_from_transaction_hash 
+      where receipts.receiver_account_id = :account_id
+      and action_receipt_actions.action_kind = 'FUNCTION_CALL' 
+      and action_receipt_actions.args->> 'method_name' = 'mint'`,
```
frol, line 542: 1. You probably only want successful receipts
2. nit: Please use upper-case for the SQL keywords (follow the code style)

frol (changes requested):

backend/src/db-utils.js
icerove: change to `queryBridgeTokenHolder`

backend/src/db-utils.js
icerove: 1. do not need that since query also needs failure one
2. fix the case pro

backend/src/db-utils.js
```diff
@@ -528,6 +528,24 @@ const queryPartnerUniqueUserAmount = async () => {
   );
 };
 
+// bridge query
+const queryBridgeTokenHolder = async (accountId) => {
```
frol, line 532: You seem to want to have a list of accounts (holders) which minted tokens through bridge, right? If that is the case:

1. the name has to be plural (`queryBridgeTokenHolders`)
2. `accountId` in the arguments misleads a reader; I suggest `bridgeTokenContractId`

backend/src/wamp.js
```diff
@@ -305,6 +305,11 @@ wampHandlers["partner-unique-user-amount"] = async () => {
   return await stats.getPartnerUniqueUserAmount();
 };
 
+// bridge query
+wampHandlers["near-bridge-user"] = async ([accountId]) => {
```
frol, line 309:
```suggestion
wampHandlers["get-bridge-token-holders"] = async ([accountId]) => {
```

frol (approved):

frol (approved):

chefsale:
@icerove Resolved the fossa key issue.

PR #600, opened 2021-04-09
queries for bridge token tracking

icerove:
there are 4 queries:
1. query for token symbol and amount from mint
2. query for factory.bridge.near for deposit transaction
3. query for token symbol and amount from withdraw
4. query for factory.bridge.near for finish withdraw transaction

### DIFF
--- a/backend/src/db-utils.js
+++ b/backend/src/db-utils.js
@@ -546,6 +546,77 @@ const queryBridgeTokenHolders = async (bridgeTokenContractId) => {
   );
 };
 
+const queryBridgeMintTransactionAction = async () => {
+  return await queryRows(
+    [
+      `SELECT
+        receipts.receiver_account_id AS symbpl_token,  
+        action_receipt_actions.args ->> 'args_base64' AS serialized_amount, 
+        TO_TIMESTAMP(receipts.included_in_block_timestamp / 1000000000),
+        originated_from_transaction_hash 
+      FROM receipts
+      JOIN action_receipt_actions ON receipts.receipt_id = action_receipt_actions.receipt_id 
+      WHERE
+        receipts.predecessor_account_id = 'factory.bridge.near'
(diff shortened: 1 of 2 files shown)


backend/src/db-utils.js
```diff
@@ -546,6 +546,77 @@ const queryBridgeTokenHolders = async (bridgeTokenContractId) => {
   );
 };
 
+const queryBridgeMintTransactionAction = async () => {
+  return await queryRows(
+    [
+      `SELECT
+        receipts.receiver_account_id AS symbpl_token,
```
frol, line 553: Is "symbpl" a typo?

backend/src/db-utils.js
```diff
@@ -546,6 +546,77 @@ const queryBridgeTokenHolders = async (bridgeTokenContractId) => {
   );
 };
 
+const queryBridgeMintTransactionAction = async () => {
```
frol, line 549: The name implies you only need a single action, but implementation fetches all of them. Do you really need all of them? It seems to me it will grow indefinitely, so this query will cause problems.

We need to limit queries to something manageable.

frol (changes requested):
I want to remind to include the execution time on mainnet and testnet whenever you introduce new queries. I don't think we can get these queries as is as they may potentially return millions of records (they are not limited).

What is the use for these records?

backend/src/db-utils.js
icerove: We really need all of them, but will not query that frequently

frol: @icerove BTW, would direct access to the database be more convenient for these use-cases? See, we now have one: https://github.com/near/near-indexer-for-explorer#shared-public-access

icerove: direct access to the database not very make sense, I have to query first and then deal with my end

icerove: you are right, I will remove these bridge and market query from explorer side and use direct access

PR #601, opened 2021-04-09
query/market data for stake and unstake accounts

icerove:
1. query for stake accounts aggregated by hour
2. query for unstake accounts aggregated by hour


frol (changes requested):
How long does it take for the queries to run? If it is longer than a second, we should cache the results as we do for other stats.

frol:
Feel free to re-open this PR once necessary

PR #604, opened 2021-04-14
feat(frontend): Update charts titles on stats page

### LINKED ISSUE #602: Stats page: Total number of XYZ vs Total amount
On stats page, the titles for graphs should be changed to say "Total number of Transactions", "Total number of accounts", "Total number of contracts", etc. Amount is used when noun that can't be measured (e.g. money, liquid, etc).

shelegdmitriy:
Resolves #602

### DIFF
--- a/frontend/src/components/stats/ActiveAccountsByDate.tsx
+++ b/frontend/src/components/stats/ActiveAccountsByDate.tsx
@@ -125,7 +125,7 @@ const ActiveAccountsByDate = ({ chartStyle }: Props) => {
       <Tab eventKey="daily" title="Daily">
         <ReactEcharts
           option={getOption(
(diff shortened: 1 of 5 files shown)


frontend/src/components/stats/GasUsedByDate.tsx
```diff
@@ -133,7 +133,7 @@ const GasUsedByDate = ({ chartStyle }: Props) => {
       <Tab eventKey="daily" title="Daily">
         <ReactEcharts
           option={getOption(
-            "Daily Amount of Used Tera Gas",
+            "Daily Number of Used Tera Gas",
```
frol, line 136: I believe gas is not countable, so let's revert to "amount" here and one below.

icerove (approved):

PR #608, opened 2021-04-17
Color network icon in options must not depend on current network

ilyar:
actual
expected


frol (approved):

frol:
@ilyar Thank you for your help!

PR #613, opened 2021-04-22
feat(frontend): Implementation of new design on Nodes page

### LINKED ISSUE #370: Design Dashboard for Validators
## Story
node dashboard term will include(with fake data):

Current epoch progress: 90.14%
Time until new epoch 00:21:52
Previous epoch length: 04:19:34
Epoch ID: 9vUWGQnUeDozaiH7L8Vs3ptphg1gtgZTfZoVsRQRto8c
block average time: 0.2s
Last finalized block: 12,344,565
stake total amount: 123M
seat price: 66,455 N

## Acceptance Criteria
- [ ] Mobile ok
- [ ] Tested

## Assets 

A really nice UI: http://solanabeach.io/

shelegdmitriy:
Resolves #370 

### Test plan

- [x] Existing tests pass;
- [x] ValidatorsList unit test covering a list of validators and proposals with all data;
- [x] ValidatorsRow unit test covering a single validator row with all data;
- [x] ValidatorsRow unit test covering a single validator row with partial data;
- [x] NodeEpoch unit test covering a section with information about epoch status;


shelegdmitriy:
@frol, currently I don't know how to get some data. But I have some thoughts of how to do it

backend/src/wamp.js
```diff
@@ -166,6 +166,10 @@ wampHandlers["get-account-details"] = async ([accountId]) => {
       .callViewMethod(lockupAccountId, "get_locked_amount", {})
       .then((balance) => new BN(balance))
       .catch(ignore_if_does_not_exist),
+    nearRpc
+      .callViewMethod(lockupAccountId, "get_reward_fee_fraction", {})
+      .then((fee) => fee)
+      .catch(ignore_if_does_not_exist),
```
shelegdmitriy, line 172: @frol, here suppose I can to fetch "fee" but I'm not sure. May be not here but in the same way.

backend/src/index.js
```diff
@@ -271,8 +303,11 @@ async function main() {
           "node-stats",
           {
             validatorAmount: validators.length,
+            seatPriceAmount: seatPrice,
             onlineNodeAmount: onlineNodes.length,
             proposalAmount: proposals.length,
+            totalStakeAmount: totalStake,
+            epochStartHeightAmount: epochStartHeight,
```
frol, line 310: "Amount" prefix does not make sense here:

```suggestion
            epochStartHeight,
```

backend/src/near.js
```diff
@@ -16,10 +21,37 @@ const queryFinalTimestamp = async () => {
 };
 
 const queryNodeStats = async () => {
-  let nodes = await nearRpc.sendJsonRpc("validators", [null]);
-  let proposals = nodes.current_proposals;
-  let currentValidators = getCurrentNodes(nodes);
-  return { currentValidators, proposals };
+  let genesisConfig = await nearRpc.sendJsonRpc(
```
frol, line 24: It is not a genesis config anymore, we use the latest protocol config for the network, so it makes sense to name it `networkProtocolConfig`:

```suggestion
  let networkProtocolConfig = await nearRpc.sendJsonRpc(
```

backend/src/near.js
```diff
@@ -16,10 +21,37 @@ const queryFinalTimestamp = async () => {
...
+    seatPrice = nearApi.validators
+      .findSeatPrice(epochStatus.current_validators, numSeats)
+      .toString();
+  }
+
+  if (!totalStake) {
+    totalStake = currentValidators
+      .reduce((acc, node) => acc.add(new BN(node.stake)), new BN(0))
+      .toString();
+  }
```
frol, line 45: We should update the seat price once on initialization, and then we can update it when the epoch changes (`epoch_start_height` is different from the one we queried previously)

backend/src/near.js
```diff
@@ -34,9 +66,37 @@ const signRemovedValidators = (removedValidators) => {
...
+    fee = await nearRpc.callViewMethod(
+      account_id,
+      "get_reward_fee_fraction",
+      {}
+    );
+    delegators = await nearRpc.callViewMethod(
+      account_id,
+      "get_number_of_accounts",
+      {}
+    );
```
frol, line 91: You have the cache of `validatorArray`, so we only need to move these updates outside of this method to make its execution independent of this `fetchValidatorsInfo` function, and thus not blocking it on awaiting for the values

frontend/src/components/nodes/CumuativeStakeChart.tsx
```diff
@@ -0,0 +1,50 @@
+interface Props {
+  value: CummulativeStake;
+}
+
+interface CummulativeStake {
```
frol, line 5:
```suggestion
interface CumulativeStake {
```

Also, rename the file to have a proper name "cumulative"

frontend/src/components/nodes/Validators.tsx
```diff
@@ -1,27 +1,116 @@
...
+        ? [
+            " ",
+            "#",
+            "Validator",
+            "Fee",
+            "Delegators",
+            "Stake",
+            "Cumulative Stake",
+          ]
+        : [" ", "#", "Validator", "Fee", "Delegators", "Stake"];
```
frol, line 50: I would avoid all this complexity and "generic" thing, and render the columns directly using `th` below instead of crafting this array first, and then iterate over it, and they apply className conditionally. It seems to be more complex than it is necessary

frontend/src/components/nodes/ValidatorsList.tsx
```diff
@@ -0,0 +1,74 @@
...
+  validators: any;
+  pages: any;
+  cellCount: number;
+  validatorType: string;
+}
+
+class ValidatorsList extends React.PureComponent<Props> {
+  calculateStake = (nodeIndex: number, totalStake: BN) => {
+    let total = new BN(0);
+    let networkHolderIndex = [];
```
frol, line 18: You don't need an array here, you only care about the very first value. Just initialize it as null and only set it if the accumulated stake is over 33% and also this variable is still null.

frontend/src/components/nodes/ValidatorsList.tsx
```diff
@@ -0,0 +1,74 @@
...
+    for (let i = 0; i <= nodeIndex; i++) {
+      total = total.add(new BN(this.props.validators[i].stake));
+
+      if (total.gt(totalStake.div(new BN(3)))) {
```
frol, line 22: Precompute the `totalStake.divn(3)` in a const before the loop, so you don't need to re-compute it over and over again in the loop.

frontend/src/components/utils/Balance.jsx
```diff
@@ -5,19 +5,23 @@ import { OverlayTrigger, Tooltip } from "react-bootstrap";
 
 const FRAC_DIGITS = 5;
 
-const Balance = ({ amount }) => {
+const Balance = ({ amount, label = null, className = "" }) => {
```
frol, line 8: It think it is better to leave className undefined if it was not defined

```suggestion
const Balance = ({ amount, label = null, className }) => {
```

frontend/src/context/NodeStatsProvider.tsx
```diff
@@ -1,41 +1,71 @@
...
+          return;
+        })
+        .catch((err: any) => console.error(err));
+    }
+
     return () => {
       explorerApi.unsubscribe("node-stats");
     };
-  }, []);
+  }, [epochStartHeight]);
```
frol, line 56: Create a separate `useEffect` since otherwise, this code will re-subscribe to `node-stats` for no good reason.

backend/src/index.js
```diff
@@ -285,6 +319,24 @@ async function main() {
   };
   setTimeout(regularCheckNodeStatus, 0);
 
+  // Periodic check of validator's fee and delegators
+  const regularCheckValidatorsExtraInfo = async (validators) => {
```
frol, line 323: Use a dedicated `setTimeout` for this `regularCheckValidatorsExtraInfo`, and don't call it from `regularCheckNodeStatus`. Fetch the validators (current validators and current proposals) inside this regular job, so you do not need to call it from `regularCheckNodeStatus`.

backend/src/index.js
```diff
@@ -320,22 +310,48 @@ async function main() {
...
+          "get_reward_fee_fraction",
+          {}
+        );
+        proposalsExtraInfo[i].delegators = await nearRpc.callViewMethod(
+          account_id,
+          "get_number_of_accounts",
+          {}
+        );
+      }
+    }
```
shelegdmitriy: @frol , I don't know is it a good idea or not... Could you please look at it and tell me your mind?

frontend/src/context/DatabaseProvider.tsx
```diff
@@ -76,13 +78,6 @@ const DatabaseProvider = (props: Props) => {
   useEffect(() => {
     const explorerApi = new ExplorerApi();
 
-    function instrumentTopicNameWithDataSource(topicName: string) {
-      if (explorerApi.dataSource === DATA_SOURCE_TYPE.LEGACY_SYNC_BACKEND) {
-        return topicName;
-      }
-      return `${topicName}:${explorerApi.dataSource}`;
-    }
-
```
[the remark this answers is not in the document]
frol: Yeah, it is fine to move it there. I will remove this thing once I get to refactor Explorer to remove the legacy indexer backend.

frol:
@shelegdmitriy I had to move a lot of things around to hopefully name things a little bit better. There are lots of things that remain to be named in a way that is hard to reason about, but I did not want to block this PR even more.

Bug report:

image

1. missing "M" (million) suffix
2. let's make total stake the same magnitude as total supply ("M")
3. Let's add comma-formatting

frol:
Bugs on mobile:

image

image

frontend/src/components/nodes/ValidatorRow.tsx
```diff
@@ -1,207 +1,422 @@
...
+              >
+                <td colSpan={cellCount}>
+                  <Row noGutters className="validator-nodes-content-row">
+                    {node.num_produced_blocks && node.num_expected_blocks && (
+                      <Col className="validator-nodes-content-cell">
+                        <Row noGutters>
+                          <Col className="validator-nodes-details-title">
+                            <Term
+                              title={"Uptime"}
+                              text={"Uptime explain text"}
```
shelegdmitriy, line 198: @frol , we still need to add some explaining text here

frontend/src/components/nodes/ValidatorRow.tsx
```diff
@@ -1,207 +1,422 @@
...
+                        </Row>
+                      </Col>
+                    )}
+                    {node.nodeInfo && (
+                      <Col className="validator-nodes-content-cell">
+                        <Row noGutters>
+                          <Col className="validator-nodes-details-title">
+                            <Term
+                              title={"Latest block"}
+                              text={"Latest block explain text"}
```
shelegdmitriy, line 229: Explaining text is required here too

frontend/src/components/nodes/ValidatorRow.tsx
```diff
@@ -1,207 +1,422 @@
...
+                        </Row>
+                      </Col>
+                    )}
+                    {node.nodeInfo?.lastSeen && (
+                      <Col className="validator-nodes-content-cell">
+                        <Row noGutters>
+                          <Col className="validator-nodes-details-title">
+                            <Term
+                              title={"Latest Telemetry Update"}
+                              text={"Latest Telemetry Update explain text"}
```
shelegdmitriy, line 263: Explaining text is required here

frontend/src/components/nodes/ValidatorRow.tsx
```diff
@@ -1,207 +1,422 @@
...
+                        </Row>
+                      </Col>
+                    )}
+                    {node.nodeInfo?.agentName && (
+                      <Col className="validator-nodes-content-cell">
+                        <Row noGutters>
+                          <Col className="validator-nodes-details-title">
+                            <Term
+                              title={"Node Agent Name"}
+                              text={"Node Agent Name explain text"}
```
shelegdmitriy, line 284: And here explaining text also needed

frontend/src/components/nodes/ValidatorRow.tsx
```diff
@@ -1,207 +1,422 @@
...
+                        </Row>
+                      </Col>
+                    )}
+                    {node.nodeInfo?.agentVersion && node.nodeInfo?.agentBuild && (
+                      <Col className="validator-nodes-content-cell">
+                        <Row noGutters>
+                          <Col className="validator-nodes-details-title">
+                            <Term
+                              title={"Node Agent Version / Build"}
+                              text={"Node Agent Version / Build explain text"}
```
shelegdmitriy, line 310: And the last one where explaining text is needed

PR #615, opened 2021-04-27
fix(backend): Current week should not be counted for weekly Active Accounts

### LINKED ISSUE #614: Weekly active users current week is cut off
When looking at the chart of the weekly active users, the last data point only shows days in the current week vs showing actually last 7d (similar problem as with daily before).

Screen Shot 2021-04-27 at 4 12 56 AM

frol:
Resolves #614

### DIFF
--- a/backend/src/db-utils.js
+++ b/backend/src/db-utils.js
@@ -264,10 +264,10 @@ const queryTransactionsCountAggregatedByDate = async () => {
   return await queryRows(
     [
       `SELECT
-        DATE_TRUNC('day', TO_TIMESTAMP(DIV(block_timestamp, 1000*1000*1000))) AS date,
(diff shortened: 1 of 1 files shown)


backend/src/db-utils.js
```diff
@@ -371,6 +371,7 @@ const queryActiveAccountsCountAggregatedByWeek = async () => {
       FROM transactions
       JOIN execution_outcomes ON execution_outcomes.receipt_id = transactions.converted_into_receipt_id
       WHERE execution_outcomes.status IN ('SUCCESS_VALUE', 'SUCCESS_RECEIPT_ID')
+      AND transactions.block_timestamp < ((CAST(EXTRACT(EPOCH FROM DATE_TRUNC('week', NOW())) AS bigint)) * 1000 * 1000 * 1000)
```
frol, line 374: This is the fix. The rest in this PR is just an alignment to use the same functions everywhere instead of my tricks with integer divisions

icerove (approved):

PR #616, opened 2021-04-27
backend: delete unused bridge query 

icerove:
since this query is already move to other repo and bridge query that not related to explorer will not put in explorer backend, delete bridge query part


frol (approved):

frol:
@icerove Thanks for cleaning it from the Explorer side!

PR #619, opened 2021-04-30
frontend/feature: adding tracking code for explorer with mixpanel

icerove:
finish tracking code with mixpanel for explorer to get the funnel for user behavior on browser side

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -53,4 +53,7 @@ module.exports = {
 
     return config;
   },
+  env: {
+    NEAR_EXPLORER_DATA_SOURCE: true,
+  },
 };
--- a/frontend/src/pages/accounts/[id].jsx
+++ b/frontend/src/pages/accounts/[id].jsx
@@ -3,6 +3,7 @@ import Head from "next/head";
 import React from "react";
 import { Container } from "react-bootstrap";
 
+import Mixpanel from "../../libraries/mixpanel";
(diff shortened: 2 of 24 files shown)


frontend/next.config.js
```diff
@@ -53,4 +53,7 @@ module.exports = {
 
     return config;
   },
+  env: {
+    NEAR_EXPLORER_DATA_SOURCE: true,
```
frol, line 57: I fail to understand the naming and the need for this configuration. Please, consider either renaming it or adding a detailed comment about it.

frol (changes requested):
@icerove Sorry for the delay in review, I have been ooo. Thanks for bringing Mixpanel integration for Explorer! :tada:

frontend/src/pages/accounts/[id].jsx
```diff
@@ -32,6 +33,9 @@ class AccountDetail extends React.Component {
 
   render() {
     const { account, accountFetchingError, currentNearNetwork } = this.props;
+    Mixpanel.track("Explorer View Individual Account", {
+      accountId: account.accountId,
+    });
```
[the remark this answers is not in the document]
Comment (icerove) [reply frontend/src/pages/accounts/[id].jsx:38]: it works fine

frontend/next.config.js
icerove: https://github.com/near/near-explorer/pull/556
I remember this is for unit test to pass since only this way can separate browser side and client side

frol (approved):

PR #630, opened 2021-05-17
refactor: Upgrade JS dependencies and update Next.js

### LINKED ISSUE #624: Upgrade dependencies
We have individual projects in:

* [ ] `/package.json` (used for prettier)
* [ ] `/backend/package.json`
* [ ] `/frontend/package.json`

Please, update them, and confirm that everything is still working fine.

NOTE: Next.js and React are the biggest projects we use, I believe, and they got some advances since the last upgrade. Pay attention to deprecation warnings are such.

shelegdmitriy:
Resolves #624 

### Test Plan:

- [x] Manually check normal pages loading


shelegdmitriy:
I was trying to upgrade `react 16.14` to `react ^17.0.0` (17.0.2 at now). But after it page loading fails with `React is not defined` text message. There is an open issue [here](https://github.com/facebook/create-react-app/issues/9953). So I'll lease it as it is.

PR #636, opened 2021-05-21
feat(backend): Calculate circulating supply

telezhnaya:
Closes (in one day) #489

PR #641, opened 2021-05-25
feat(backend): add transition logic for circulating supply

telezhnaya:
We need that for https://github.com/near/near-explorer/pull/636


telezhnaya:
You need to review changes only from the last commit, see https://github.com/near/near-explorer/pull/641/commits/eb7a033074267232b4c1cb10e513d57a8b183e2b

telezhnaya:
See https://github.com/near/near-explorer/pull/642

PR #642, opened 2021-05-25
feat(backend): add transition logic for circulating supply

telezhnaya:
We need that for #636

Recreated the PR because it's not possible to change src/dst after creating
The base is not master, it's the branch for #636

### DIFF
--- a/backend/src/index.js
+++ b/backend/src/index.js
@@ -1,7 +1,6 @@
 const moment = require("moment");
 
 const models = require("../models");
-const { wampNearNetworkName } = require("./config");
 
 const {
   isLegacySyncBackendEnabled,
@@ -16,6 +15,7 @@ const {
   regularFetchStakingPoolsInfoInterval,
   regularStatsInterval,
   regularCalculateCirculatingSupplyInterval,
+  wampNearNetworkName,
 } = require("./config");
 
 const { DS_LEGACY_SYNC_BACKEND, DS_INDEXER_BACKEND } = require("./consts");
--- a/backend/src/index.js
+++ b/backend/src/index.js
@@ -303,8 +303,9 @@ async function main() {
         currentProposals = await extendWithTelemetryInfo(
           epochStats.currentProposals
         );
-        const onlineValidatingNodes =
-          pickOnlineValidatingNode(currentValidators);
+        const onlineValidatingNodes = pickOnlineValidatingNode(
+          currentValidators
+        );
         const onlineNodes = await queryOnlineNodes();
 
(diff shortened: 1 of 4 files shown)


frol (approved):

telezhnaya:
@frol do I need to change the transition period? It started a week ago 🙂 
Please also ping me when it would be the time for merge

frol:
@telezhnaya We need to implement some persistence of the computed values as otherwise, we will report some nonsense circulating supply when we restart the backend (e.g. on new deploy)

Given we discussed a design of separate services pushing various stats into the Indexer DB, and us not being there yet, I suggest we go with any hack you feel is good enough to just get the job done (e.g. storing it in the SQLite table we already have on the backend [it might be quite cumbersome] or just plaintext/json file in the `db/` folder).

telezhnaya:
@frol I hate this new piece of code but maybe it's OK as a temporary solution.
Good news: it motivates me to write microservice on Rust faster!

Could you give me access to the server? I want to check logs sometimes.

We need to double-check if the program could create a file on the filesystem, and maybe we should change the filepath.

frol:
Please, fix `/api/circulating-supply-in-near` endpoint as well

frol (approved):

backend/src/index.js
```diff
@@ -302,9 +302,8 @@ async function main() {
         currentProposals = await extendWithTelemetryInfo(
           epochStats.currentProposals
         );
-        const onlineValidatingNodes = pickOnlineValidatingNode(
-          currentValidators
-        );
```
telezhnaya, line 307: It's done by precommit hook

backend/src/index.js
```diff
@@ -427,7 +426,9 @@ async function main() {
     await startStatsAggregation();
   }
 
-  startRegularCalculationCirculatingSupply();
+  if (nearRpc.connection.url.includes("mainnet")) {
```
frol, line 429: https://github.com/near/near-explorer/blob/2d71c17e09c8f379c1174ebacaa5170dc5d0c806/backend/src/config.js#L58

```suggestion
  if (wampNearNetworkName === "mainnet") {
```

Please, add the import and double check that it works fine

backend/src/index.js
frol: Run `npm install` from the root of the near-explorer project, so it installs the exact version of prettier on your side. Otherwise, it fails on CI. We will update the prettier and reformat the codebase in a separate PR.

frol:
@telezhnaya Please, keep an eye on it once it is live on production :pray:

PR #652, opened 2021-06-02
feat(frontend): display up to 120 nodes per page

frol:
Partially address #646.

# Test Plan

* [x] existing tests pass without changes
* [x] the page displays all the current validators on mainnet (59) on a single page
* [x] the regular updates do not crash CPU (without the change in the Context, it did x4 more re-renders since each individual state dispatch triggered re-render)


shelegdmitriy:
@frol, I saw an issue when items per page is fitted only on one page but pagination is still visible, so I fixed it. That was my fail

Issue #631, opened 2021-05-17
When function call arguments are empty, explorer is stuck on "Loading..."


mikroskeem:
For example: https://explorer.testnet.near.org/transactions/BGWDmSzSJ6FEEVuKvWaXViPNbPiNn5V6jDV3EDP5mNkU (and few other recent calls on this dev account).

Trick is to construct a FunctionCallAction with empty args vec. I think explorer assumes that args is never empty, and near-api-js also sets args to `{}` (string) when args aren't supplied.

frol:
@mikroskeem Thanks for reporting it! We will resolve it later this week

PR #653, opened 2021-06-07
fix(frontend): Endless 'Loading...' when arguments are empty

shelegdmitriy:
Resolves: #631

### DIFF
--- a/frontend/src/components/transactions/ActionMessage.tsx
+++ b/frontend/src/components/transactions/ActionMessage.tsx
@@ -88,7 +88,7 @@ const transactionMessageRenderers: TransactionMessageRenderers = {
   }: Props<T.FunctionCall>) => {
     let args;
     if (showDetails) {
-      if (typeof actionArgs.args === undefined) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/transactions/ActionMessage.tsx
```diff
@@ -88,8 +88,13 @@ const transactionMessageRenderers: TransactionMessageRenderers = {
   }: Props<T.FunctionCall>) => {
     let args;
     if (showDetails) {
-      if (!actionArgs.args) {
+      if (typeof actionArgs.args === undefined) {
```
frol, line 91:
```suggestion
      if (typeof actionArgs.args === "undefined") {
```

frol (approved):

frontend/src/components/transactions/ActionMessage.tsx
mikroskeem: Isn't `!!` better way to do this?

Issue #598, opened 2021-04-08
Error: `Result not found` for deleted users


telezhnaya:
It's possible to look at deleted account if I create the link by hands. But, if I put the username into search box, I have `Result not found`.

Example:
https://explorer.near.org/accounts/fa990f6668dd41c16e5cfd7d59f21ade2b8ec7f7.lockup.near
Link works fine

But it's unable to put same username to search field.

PR #656, opened 2021-06-08
fix(frontend): Fix searching for deleted users

shelegdmitriy:
Resolves: #598

### DIFF
--- a/frontend/src/libraries/explorer-wamp/accounts.ts
+++ b/frontend/src/libraries/explorer-wamp/accounts.ts
@@ -197,27 +197,21 @@ export default class AccountsApi extends ExplorerApi {
     }
   }
 
-  async queryDbAccount(accountId: string): Promise<any> {
+  async isAccountIndexed(accountId: string): Promise<boolean> {
     try {
-      let accountDetails;
-      if (this.dataSource === DATA_SOURCE_TYPE.INDEXER_BACKEND) {
-        accountDetails = this.call<any>("select:INDEXER_BACKEND", [
-          `SELECT
+      return this.call<any>("select:INDEXER_BACKEND", [
+        `SELECT
             account_id
           FROM accounts
           WHERE account_id = :accountId`,
-          {
-            accountId,
-          },
-        ]).then((it) =>
-          it.length === 0 || !it[0].account_id ? undefined : it[0].account_id
-        );
-      } else {
-        throw Error(`unsupported data source ${this.dataSource}`);
-      }
-      return accountDetails;
+        {
+          accountId,
+        },
+      ]).then((it) => Boolean(it[0]?.account_id));
     } catch (error) {
-      console.error("AccountsApi.queryDbAccount failed to fetch data due to:");
+      console.error(
+        "AccountsApi.isAccountIndexed failed to fetch data due to:"
+      );
       console.error(error);
       throw error;
     }
--- a/frontend/src/libraries/explorer-wamp/accounts.ts
+++ b/frontend/src/libraries/explorer-wamp/accounts.ts
@@ -199,7 +199,7 @@ export default class AccountsApi extends ExplorerApi {
 
   async isAccountIndexed(accountId: string): Promise<boolean> {
     try {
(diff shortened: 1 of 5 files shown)


frontend/src/libraries/explorer-wamp/contracts.ts
```diff
@@ -56,8 +56,12 @@ export default class ContractsApi extends ExplorerApi {
     };
 
     try {
-      const codeHash = await this.queryCodeHash(id);
-      if (codeHash !== "11111111111111111111111111111111") {
+      // codeHash precheck is requered due to deleted accounts
+      let codeHash;
+      await this.queryCodeHash(id)
+        .then((it) => (codeHash = it))
+        .catch(() => {});
```
frol, line 63: This should work, I believe. Please, double-check

```suggestion
      // codeHash does not exist for deleted accounts
      const codeHash = await this.queryCodeHash(id).catch(() => null);
```

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -197,6 +197,32 @@ export default class AccountsApi extends ExplorerApi {
     }
   }
 
+  async queryDbAccount(accountId: string): Promise<any> {
+    try {
+      let accountDetails;
+      if (this.dataSource === DATA_SOURCE_TYPE.INDEXER_BACKEND) {
+        accountDetails = this.call<any>("select:INDEXER_BACKEND", [
```
frol, line 204: The returned value is not `accountDetails`. Please, be mindful of variable naming since misleading naming is worse than random names.

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -197,6 +197,32 @@ export default class AccountsApi extends ExplorerApi {
     }
   }
 
+  async queryDbAccount(accountId: string): Promise<any> {
```
frol, line 200: Given that the only reason we added this new method is to check if the given account id has ever existed, let's implement the method `isAccountIndexed` and return `Promise<boolean>`

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -197,6 +197,32 @@ export default class AccountsApi extends ExplorerApi {
...
+    try {
+      let accountDetails;
```
frol, line 202: There is no need for this variable, just return the result immediately.

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -197,6 +197,32 @@ export default class AccountsApi extends ExplorerApi {
...
+      if (this.dataSource === DATA_SOURCE_TYPE.INDEXER_BACKEND) {
```
frol, line 203: I have already created the `legacy` branch of near-explorer, so we can strip all these `dataSource` checks in the master branch.

backend/package-lock.json
```diff
@@ -1,1818 +1,8 @@
 {
   "name": "near-explorer-backend",
   "version": "0.0.1",
-  "lockfileVersion": 2,
```
frol, line 4: Please, revert the changes in the `backend/package-lock.json` since they are no-op and not relevant to this PR anyway.

frontend/src/libraries/explorer-wamp/contracts.ts
shelegdmitriy: Yes, you were right!

frol (approved):

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -197,6 +197,26 @@ export default class AccountsApi extends ExplorerApi {
     }
   }
 
+  async isAccountIndexed(accountId: string): Promise<boolean> {
+    try {
+      return this.call<any>("select:INDEXER_BACKEND", [
```
frol, line 202:
```suggestion
      return await this.call<any>("select:INDEXER_BACKEND", [
```

frol (approved):

PR #657, opened 2021-06-10
fix(frontend): Fix confusing information on Nodes page

### LINKED ISSUE #646: Validator nodes screen is confusing
# Details

Screen Shot 2021-05-28 at 4 16 06 AM

1. Validator account names are not full even though they are short. "staked.poolv1.near..." is especially funny, what's behind ... where?
2. key "ed25519... " shown doesn't make any sense. Also it leads to... transaction details of non existent transaction?
3. Why 12 validators per page? We only have 60, they all can be shown on single page. Ideally just let's show all validators on one page until this becomes an issue. All other tools (https://near.zavodil.ru/?pools=) show on a single page fine.
4. https://explorer.near.org/nodes/proposals page doesn't really make sense to me. It shows all the validator that got more delegations together with new validators? I would merge new validators into the list of validators and mark them "new". Not sure if we want to show validators who received more or less delegation but given there is also reward there, would need to query actual transactions to the contract to determine how much was staked / unstaked.
5. Did we sunset the geo map? Is there plan to return it?
6. It would be good to see also not currently validating nodes that are running with staking contract - for people who want to delegate to new nodes, that will be the only to find them given they don't show up in the wallet.

# Steps to reproduce 
1. Go to https://explorer.near.org/nodes/validators and https://explorer.near.org/nodes/proposals

shelegdmitriy:
Partially resolves #646 

It resolves such points 1 and 2.

### DIFF
--- a/frontend/src/components/nodes/ValidatorRow.tsx
+++ b/frontend/src/components/nodes/ValidatorRow.tsx
@@ -10,7 +10,6 @@ import Balance from "../utils/Balance";
 import { TableRow, TableCollapseRow } from "../utils/Table";
 import Term from "../utils/Term";
 import Timer from "../utils/Timer";
-import TransactionLink from "../utils/TransactionLink";
 import ValidatingLabel from "./ValidatingLabel";
 import CumulativeStakeChart from "./CumulativeStakeChart";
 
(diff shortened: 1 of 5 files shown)


frol:
@shelegdmitriy FYI the unittest CI failed and Render link is also dead because of TypeScript guards

shelegdmitriy:
> @shelegdmitriy FYI the unittest CI failed and Render link is also dead because of TypeScript guards

Yeah, I saw it. I'll fix it

shelegdmitriy:
@frol render link operates normally

frol (approved):
LGTM

PR #672, opened 2021-06-22
fix: take hostname from req.headers.host

telezhnaya:
still investigate why we don't have host parameter in request here https://explorer.near.org/api/circulating-supply


frol (approved):

frol:
I will merge it with some other PR to avoid unnecessary downtime.

Issue #509, opened 2020-12-25
Display active access keys on Account Details page


frol:
## Story

As an Explorer user who wants to review the access keys of a given account, I want to see the list of all the active ~and deleted~ keys so I can have a better picture.

## Acceptance Criteria

- [ ] Account Details page has a list of active ~and removed~ access keys (updated on 2021-07-20)
- [ ] Mobile ok
- [ ] The dumb components are unit-tested

shelegdmitriy:
We already have this on beta account activity page

PR #675, opened 2021-06-30
feat(frontend): Display active access keys on Account Details page

shelegdmitriy:
Resolves: #509 

### Test Plan:

- [ ] Add existing test of new component;


frol (changes requested):
Reviewing how the list of active and deleted keys get mixed and presented in action-like format, I believe we should not go that path and instead just display a list of active access keys (full access key should be copyable and details about the permissions should be as descriptive as they are in the current list) with their current nonce.

@reefoh Hey, this is a tech-savy feature which is still quite relevant even for the end-users if they want to review their access keys. Do you think we can have some design decision here that does not require the whole page refactoring just yet? Here is how it currently looks like: https://near-explorer-frontend-with-indexer-pr.onrender.com/accounts/olga.near

Issue #42, opened 2019-07-25
Localization (l10n) and internationalization (i18n)


frol:
It is a placeholder issue tracking our future efforts around multi-lingual support in NEAR Explorer

vgrichina:
@frol do you think https://ryandrewjohnson.github.io/react-localize-redux-docs/ is a solid solution?

frol:
@vgrichina it looks good, but I have not validated alternatives yet. Do you have experience with this particular module or other models for l10n & i18n?

vgrichina:
@frol don't have experience yet, but plan to try it soon for wallet. I like that it allows you to provide HTML/React translations as well.

frol:
Just bumped into another alternative: https://lingui.js.org/

PR #679, opened 2021-07-06
feat(frontend): Internalization (i18n) of Explorer

think-in-universe:
### Description

Resolves: https://github.com/near/near-explorer/issues/42

With the growth of our user base in non-English communities, the requirement to view explorer in local language has been more urgent than before. 

### Current Progress in the PR

1. Added Simplied Chinese language support as an example.
2. Instead of using the `next.js`'s i18n route, here we added i18n support using the same package as NEAR Wallet -- [`react-localize-redux`](https://ryandrewjohnson.github.io/react-localize-redux-docs/)
3. Added language switch for desktop and mobile UI
4. Automatic language detection based on browser language
5. i18n support for the home page
6. More pages' translation are supported by @IsQiao in https://github.com/near/near-explorer/pull/684
   - [X] Accounts and Account Details
   - [X] Blocks and Block Details
   - [X] Part of Transactions and Transaction Details

### To Do

1. [ ] Add i18n support for all the remaining pages 
   - [x] Review the translation of all pages
       - [x] Home
       - [x] Accounts
       - [x] Block
   - [x] Complete Transactions and Transaction Details
   - [x] Contract
   - [ ] Nodes
   - [ ] Stats & Charts
        - [x] Labels
        - [ ] Tooltip
   - [x] Finish all i18n transalations
2. [x] Fix tests failure
3. [ ] Defects
   - [x] The language selection is not preserved between pages and on page reload.
   - [x] Enhahnce `moment` package locales, e.g. `ago`
   - [ ] Language Switcher UI defect in Firefox
   - [x] NEAR symbol Ⓝ is different in local and production
4. [ ] Enhancements
   - [x] Enhance `Term` component
   - [x] Remove unnecessary i18n strings initialization in some pages if possible
   - [ ] Split i18n JSON by pages or components if possible

### Screenshot

Explorer Home Page in Simplified Chinese

image


think-in-universe:
@telezhnaya @shelegdmitriy @frol 

Hey everyone, we have some quite important project in the ecosystem that could benefit a lot from the i18n support if it could be enabled ASAP. Could you please help review the PR once you have time? I'll resume the efforts to add i18n support for all the remaining pages.

shelegdmitriy:
@think-in-universe yeah for sure! I'll review it from my side

think-in-universe:
@shelegdmitriy @frol  Hey Dmitriy and Frol, 

could you please review the i18n PR today first if you get time?  The plan is to make the i18n support in production before end of July 15th if possible. 

We still have several pages that I'm adding the i18n support, and will try to make it ready ASAP.

frol:
@think-in-universe The language selection is not preserved between pages and on page reload. It falls back to the browser local, so I guess most of the users will be fine, but still... We should save the selected locale in the cookies, so server-side rendering can also pick it up

frol:
Unit tests needs to be updated: `npm run test -- -u`

End-to-end tests I believe just glitched, so I restarted the CI.

frol:
@think-in-universe Unit tests needs to be configured to use the translations since otherwise we have `Missing translationId` in the snapshots which is not very useful.

frol:
Also, the "E" letter is partially visible behind the translation dropdown on my Arch Linux, Firefox 89.0.2 (while it works fine in Google Chrome):

image

frontend/src/components/dashboard/DashboardBlock.tsx
```diff
@@ -8,66 +8,95 @@ import LongCardCell from "../utils/LongCardCell";
...
+                          <p>
+                            {translate(
+                              "component.dashboard.DashboardBlock.block_height.desc2"
+                            )}
+                          </p>
+                          <p>
+                            {translate(
+                              "component.dashboard.DashboardBlock.block_height.desc3"
+                            )}
+                          </p>
```
frol, line 58: In `frontend/src/components/accounts/AccountDetails.tsx` you [used `{ renderInnerHtml: true }`](https://github.com/near/near-explorer/pull/679/files?diff=split&w=1#diff-2251c6f194c4bc059c717310df2d85119d9d52070a1004c36ff83519cda22431R138-R142). I feel that is a better option to use since all these `desc*` are part of the same documentation text

frontend/src/components/utils/HeaderNavDropdown.jsx
```diff
@@ -60,84 +62,88 @@ const HeaderNavItem = ({ link, icon, text }) => {
...
+      <Dropdown>
+        <Dropdown.Toggle className="chain-header">
+          {translate("component.utils.HeaderNavDropdown.title")}
+          <img src="/static/images/down-arrow.svg" className="dropdown-arrow" />
+        </Dropdown.Toggle>
+        <Dropdown.Menu className="header-dropdown-menu">
+          <HeaderNavItem
+            link="/accounts"
+            icon={<IconAccounts className="header-icon" />}
+            text={translate("model.accounts.title")}
```
frol, line 76: I am trying to understand the convention around "model." prefix. Could you shed some light?

frontend/src/components/dashboard/DashboardNode.tsx
```diff
@@ -7,56 +7,74 @@ import DashboardCard from "../utils/DashboardCard";
...
+                <LongCardCell
+                  title={
+                    <Term
+                      title={translate(
+                        "component.dashboard.DashboardNode.nodes_online.title"
+                      ).toString()}
+                      text={
+                        translate(
+                          "component.dashboard.DashboardNode.nodes_online.desc"
+                        ).toString() + " "
```
frol, line 33: Instead of adding these awkward `+ " "`, let's add this space to the `Term` component implementation once

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -23,285 +25,304 @@ class AccountDetails extends React.Component<Props> {
...
+                xs="12"
+                md={typeof account.storageUsage === "undefined" ? "12" : "4"}
+              >
+                <CardCell
+                  title={
+                    <Term
+                      title={translate("model.transactions.title").toString()}
+                      text={translate(
+                        "component.accounts.AccountDetails.transactions.text"
+                      ).toString()}
```
frol, line 42: Is it required to use `.toString()` explicitly? In [Search component you skipped it](https://github.com/near/near-explorer/pull/679/files?diff=split&w=1#diff-7af4e8dfed46b64428f48a72bc152ce7ea0f04b2992ec9c797ab6106765faa3dR87), and it seems to be working fine

frontend/src/components/utils/HeaderNavDropdown.jsx
think-in-universe: The original purpose is to put the commonly used fields regarding each `model` (account / block / transaction, etc.) under `model`. Renamed to `common` in the latest version to make it easy to understand.

frontend/src/components/dashboard/DashboardNode.tsx
think-in-universe: Yes. already fixed.

frontend/src/components/accounts/AccountDetails.tsx
think-in-universe: It's not required to use `toString()` for `text` property of `<Term />`, but it's required to convert to string for `title` as defined in <Term /> component. 

```ts
interface Props {
  title: string;
  text: string | React.ReactNode;
  href?: string;
}
```

We can update <Term /> componnet or create another <TermI18N /> component to reduce those redundant i18n related processing.

frontend/src/components/dashboard/DashboardBlock.tsx
think-in-universe: I remember we have tried put the html string into JSON file, but failed to render correctly. It's probably caused by how `<Term />` component is implemented, and I can improve this today.

think-in-universe: After testing, we figure out the previous failure was caused by one bug in the fallback mechanism of `react-localize-redux`. (The fallback will not work for html strings.) It's working now if we add html translation correctly for both the default language and active language

think-in-universe:
@frol we're ready with most pages except nodes and stats. 

the major blocker now is the unit test, and I need to take a bit more time tomorrow to figure out how to configure the context in `react-test-renderer` to add the translations.

think-in-universe:
@frol I have added i18n support in unit tests, but there're still some mismatch that may need your help to review. 

I didn't update any snapshots, in case there're any uncaught break.

frontend/src/components/utils/LangSwitcher.tsx
```diff
@@ -0,0 +1,39 @@
...
+  code: string;
+  name: string;
+}
+
+const LanguageToggle = ({
+  languages,
+  activeLanguage,
+  setActiveLanguage,
+}: any) => {
+  const selectLangauge = (code: string) => {
```
frol, line 15: There is a typo in `selectLangauge`

frontend/src/components/accounts/AccountDetails.tsx
frol: I see. I am fine with this explicitness, just wanted to double-check. I am fine to lift the type requirements from just `string` to whatever is needed. I don't think we need `TermI18N` component, though

frontend/src/components/transactions/__tests__/__snapshots__/ActionMessage.test.tsx.snap
```diff
@@ -223,7 +223,6 @@ Array [
      
     Ⓝ
   </span>,
-  " ",
```
frol, line 226: Without this space the phrase will be jammed together

frontend/src/components/utils/LangSwitcher.tsx
think-in-universe: fixed

think-in-universe:
@frol updated the <Term /> component so the `title` also accepts property of type `React.ReactNode`

frontend/src/components/transactions/__tests__/__snapshots__/ActionMessage.test.tsx.snap
think-in-universe: fixed now

think-in-universe:
@frol Hey Frol, I think we're ready to move into production (testnet and mainnet) if there's no other blockers. 

Almost all the pages / components are translated, except the Nodes page, which I'll add the support tomorrow.

think-in-universe:
> Also, the "E" letter is partially visible behind the translation dropdown on my Arch Linux, Firefox 89.0.2 (while it works fine in Google Chrome):
> 
> image

Just noticed that Wallet has the same issue in Firefobx. We need to fix both :)

think-in-universe:
> Also, the "E" letter is partially visible behind the translation dropdown on my Arch Linux, Firefox 89.0.2 (while it works fine in Google Chrome):
> 
> image

Just noticed that Wallet has the same issue in Firefobx. We need to fix both :)

frontend/src/pages/accounts/[id].jsx
```diff
@@ -45,11 +47,16 @@ class AccountDetail extends React.Component {
           <title>NEAR Explorer | Account</title>
         </Head>
         <Content
-          title={<h1>{`Account: @${account.accountId}`}</h1>}
+          title={
+            <h1>
+              <Translate id="common.accounts.account" />
+              {`: @${account.accountId}`}
```
Comment (frol) [inline frontend/src/pages/accounts/[id].jsx:53]: Should we put the accountId into a parameter of the translation?

frontend/src/pages/blocks/[hash].jsx
```diff
@@ -55,46 +57,52 @@ class BlockDetail extends React.Component {
...
+            </Head>
             <Content
-              size="medium"
-              icon={<TransactionIcon style={{ width: "22px" }} />}
-              title={<h2>Transactions</h2>}
+              title={
+                <h1>{`${translate("page.blocks.title").toString()} ${
+                  block.height
+                    ? `#${block.height}`
+                    : `${block.hash.substring(0, 7)}...`
```
Comment (frol) [inline frontend/src/pages/blocks/[hash].jsx:71]: It should be parametrized translation component, I guess.

frontend/src/pages/blocks/[hash].jsx
```diff
@@ -55,46 +57,52 @@ class BlockDetail extends React.Component {
...
+                <BlockDetails block={block} />
+              )}
             </Content>
+            {!this.props.err ? (
+              <>
+                <Content
+                  size="medium"
+                  icon={<TransactionIcon style={{ width: "22px" }} />}
+                  title={
+                    <h2>{translate("common.transactions.transactions")}</h2>
```
Comment (frol) [inline frontend/src/pages/blocks/[hash].jsx:88]: Translation component? (I see a few other places in pages/ folder where it could be used)

frol (approved):

frol:
@think-in-universe Thank you for this attention to detail and pulling out this huge chunk of work in almost no time! :tada:

PR #684, opened 2021-07-09
feat(frontend): Chinese i18n support for accounts/blocks/transactions pages

IsQiao:


think-in-universe:
thanks @IsQiao 

This PR is working together with another one for i18n support on here: https://github.com/near/near-explorer/pull/679

PR #687, opened 2021-07-17
Fix i18n error in SSR

think-in-universe:
With https://github.com/near/near-explorer/pull/679, now we support i18n in explorer. 

However, server side rendering is in a mess.

image

This fix will render i18n in SSR with either default language or user selected langauge.


think-in-universe:
@frol I also added getting langauges from cookies and accept-langauge. 

Please have a try whether it works or not. We can merge if there's no breaks from the preview version.

Issue #667, opened 2021-06-16
Upgrade to Next.js 11


frol:
The release further improves the performance and has eslint integration (consider enabling that for guarding us against silly typos)

https://nextjs.org/blog/next-11

PR #695, opened 2021-07-29
refactor(frontend): Upgrade to Next.js 11

shelegdmitriy:
Resolves: #667


frontend/next.config.js
```diff
@@ -53,4 +53,5 @@ module.exports = {
 
     return config;
   },
+  webpack5: false,
```
shelegdmitriy, line 56: This is a temporary solution to run a project. It must be solved in next iterations

frontend/next-env.d.ts
```diff
@@ -1,10 +1,3 @@
 /// <reference types="next" />
-
-declare module "*.svg" {
-  const content: React.FC;
-  export default content;
-}
-
-declare module "react-text-collapse";
-
-declare module "*"; // for datamaps npm
```
frol, line 10: I think these might still be necessary (unit tests seem to fail due to these)

frol:
FYI, unit tests should be fixed.

frontend/src/libraries/explorer-wamp/index.ts
```diff
@@ -71,7 +71,7 @@ export class ExplorerApi {
   }
 
   // Establish and handle concurrent requests to establish WAMP connection.
-  static getWampSession(): Promise<autobahn.Session> {
+  static getWampSession(): Promise<autobahn.Session | any> {
```
shelegdmitriy, line 74: `| any` - is a temporary solution. As we must to use a new version of `typescript` there is an error: 
Снимок экрана 2021-08-24 в 11 07 12

Thats why I've added a `| any` as a temporary solution. I'll remove it later

frol (approved):
Cool! I checked it manually on my laptop and it seems to operate just fine.

I will merge it later today.

Issue #699, opened 2021-08-06
Number of transactions does not load for some accounts


bowenwang1996:
For example, https://explorer.near.org/accounts/canvas-war.feiyu.near does not show the number of transactions
<img width="1122" alt="Screen Shot 2021-08-06 at 10 05 54 AM" src="https://user-images.githubusercontent.com/13601726/128546903-87a772a1-b707-4ac4-8367-aff2ad41d0d5.png">

frol:
This particular behavior is a hotfix that I deployed to make it possible to open the account page in the first place since counting the number of transactions just runs out of resources and does not complete in any reasonable time.

The path forward is:

* decouple the transactions counting from the initial page load and introduces the loader bahaviour. This won't help the traffic-heavy accounts to display the number of transactions, but will work fine for the low-traffic accounts (@shelegdmitriy is working on this)
* optimize the way we get the number of transactions by pre-aggregating the numbers daily on the Indexer for Explorer side (@telezhnaya is working on this)

PR #703, opened 2021-08-12
Fix loading transactions on heavy accounts

shelegdmitriy:
Resolves: #699

### DIFF
--- a/frontend/src/components/accounts/AccountDetails.tsx
+++ b/frontend/src/components/accounts/AccountDetails.tsx
@@ -4,10 +4,7 @@ import React from "react";
 
 import { Row, Col, Spinner } from "react-bootstrap";
 
-import AccountsApi, {
-  Account,
-  AccountStats,
-} from "../../libraries/explorer-wamp/accounts";
+import AccountsApi, { Account } from "../../libraries/explorer-wamp/accounts";
 import { NearNetwork } from "../../libraries/config";
 
 import CardCell from "../utils/CardCell";
@@ -25,24 +22,32 @@ export interface Props {
 }
 
 interface State {
-  transactionCount?: AccountStats;
+  outTransactionsCount?: number;
+  inTransactionsCount?: number;
 }
 
 class AccountDetails extends React.Component<Props> {
-  state: State = {
-    transactionCount: {},
-  };
+  state: State = {};
 
   collectTransactionCount = async () => {
     new AccountsApi()
-      .queryTransactionCount(this.props.account.accountId)
-      .then((transactionCount) => {
-        if (transactionCount) {
-          this.setState({ transactionCount });
+      .queryAccountStats(this.props.account.accountId)
+      .then((accountStats) => {
+        if (accountStats) {
+          this.setState({
+            outTransactionsCount: accountStats.outTransactionsCount,
+            inTransactionsCount: accountStats.inTransactionsCount,
+          });
         }
         return;
       })
-      .catch((error) => console.error(error));
+      .catch((error) => {
+        console.warn(
+          "Account information retrieval failed for ",
+          this.props.account.accountId,
+          error
+        );
+      });
   };
 
   componentDidMount() {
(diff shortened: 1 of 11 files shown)


frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -131,11 +131,12 @@ export default class AccountsApi extends ExplorerApi {
...
     };
 
-    const [accountInfo, accountBasic, accountStats] = await Promise.all([
+    const [accountInfo, accountBasic] = await Promise.all([
       this.queryAccount(accountId),
       queryBasicInfo(),
-      queryTransactionCount(),
     ]);
+    // accountStats is loaded separately to prevent infinite loader on heavy accounts
+    const accountStats = await queryTransactionCount();
```
frol, line 139: This still blocks this `getAccountInfo` function until we get the count, so it does not solve the problem. Move this `queryTransactionCount`, so it is called separately.

frol (changes requested):
The change is fine, but it won't solve the original problem for the heavy accounts

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -50,19 +90,19 @@ class AccountDetails extends React.Component<Props> {
                     <>
                       <span>
                         &uarr;
-                        {typeof account.outTransactionsCount === undefined ? (
-                          <Spinner animation="border" variant="secondary" />
+                        {transactionCount?.outTransactionsCount ? (
```
frol, line 93: bug: If the value will be 0, it will cast it to false and display spinner instead of "0"

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -10,9 +10,9 @@ export interface AccountBasicInfo {
   deletedAtBlockTimestamp: number | null;
 }
 
-interface AccountStats {
-  inTransactionsCount: number;
-  outTransactionsCount: number;
+export interface AccountStats {
+  inTransactionsCount?: number;
+  outTransactionsCount?: number;
```
frol, line 15: 1. I feel `?` is not necessary since you either have the TransactionsStats or not have them at all
2. `Account` type (defined below) should not have `AccountStats`

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -17,13 +20,50 @@ import WalletLink from "../utils/WalletLink";
...
   currentNearNetwork: NearNetwork;
 }
 
+interface State {
+  transactionCount?: AccountStats;
+}
+
 class AccountDetails extends React.Component<Props> {
+  state: State = {
+    transactionCount: {},
```
frol, line 33: I would remove it completely or define as `undefined` to match the `State` interface.

```suggestion
```

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -17,13 +20,50 @@ import WalletLink from "../utils/WalletLink";
...
+  };
+
+  componentDidMount() {
+    this.collectTransactionCount();
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    if (prevProps.account.accountId !== this.props.account.accountId) {
+      this.setState({ transactionCount: {} });
+      this.collectTransactionCount();
```
frol, line 55: 1. set the `transactionCount` to undefined
2. use the `setState` callback to avoid a situation when the `transactionCount` get reset after `collectTransactionCount` set state to the fresh values. It should not happen in this scenario since setting a state is faster than a query to the network, but it is better to use the proper setup always to avoid unexpected bugs due to ordering of certain events.

```suggestion
      this.setState({ transactionCount: undefined }, this.collectTransactionCount);
```

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -88,7 +88,18 @@ export default class AccountsApi extends ExplorerApi {
...
+      this.queryAccount(accountId),
+      queryBasicInfo(),
+    ]);
+    return {
+      ...accountInfo,
+      ...accountBasic,
+    };
+  }
+
+  async queryTransactionCount(accountId: string): Promise<AccountStats> {
```
frol, line 101: I suggest we rename the function to reflect the relation to the account (e.g. `queryAccountStats`). This way the name of the function and the return value make sense.

P.S. update the `state` field name in the AccountDetails component accordingly.

frol (changes requested):

frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -49,11 +102,23 @@ class AccountDetails extends React.Component<Props> {
                   text={
                     <>
                       <span>
-                        &uarr;{account.outTransactionsCount.toLocaleString()}
+                        &uarr;
+                        {outTransactionsCount &&
+                        typeof outTransactionsCount !== undefined ? (
```
frol, line 107: 1. `(0 && ...)` is `false`, so the check is incorrect
2. using `typeof` you will get a string `"undefined"`

Thus, here is the proper check:

```suggestion
                        {outTransactionsCount !== undefined ? (
```

frol:
Add a test with `outTransactionsCount` and `inTransactionsCount` equal to 0. It should render 0.

frol (approved):

PR #700, opened 2021-08-09
feat(frontend): NEAR Explorer translation for Vietnamese Users

that83:
Contribution to the bounty: https://github.com/near/bounties/issues/62

PR #706, opened 2021-08-15
fix(frontend): the missing i18n support in the ValidatorRow component

IsQiao:
Fix: the missing i18n support in the ValidatorRow component
Related PR: #700 
Because I don’t understand Vietnamese, it kept the default in Vietnamese mode


think-in-universe (changes requested):
suggestion about Chinese translation

think-in-universe (approved):

think-in-universe:
great job! thanks @IsQiao

frol (approved):

shelegdmitriy (approved):

PR #718, opened 2021-08-24
feat(fossa): run fossa only on master to unbreak external contributors

chefsale:
Running FOSSA only on master is sufficient enough. We can easily revert or fix the issues as they occur.


frol (approved):

frol:
@chefsale Thanks!

PR #722, opened 2021-08-24
refactor: Upgraded Node.js in Docker images to 16.4.2

### LINKED ISSUE #719: Unable to build explorer backend inside Docker
Backend uses 12 version of JS
It does not support `??` operator https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator
@shelegdmitriy used it recently in `backend/src/index.js`

I tried to bump up JS version: it's not so easy, it requires installing python and gives other weird issues.

We can easily change `??` to `||`, looks like it should work in our case.

@frol @khorolets

frol:
*Resolves #719*

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,11 +1,11 @@
-FROM mhart/alpine-node:12
+FROM mhart/alpine-node:16
 
 USER root
 
 WORKDIR /near-explorer/backend
 ENV HOME=/tmp
 COPY ./package.json ./package-lock.json ./
-RUN apk add --no-cache --virtual=.build-dependencies git && \
+RUN apk add --no-cache --virtual=.build-dependencies git python2 make g++ musl-dev && \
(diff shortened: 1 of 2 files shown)


backend/Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM mhart/alpine-node:12
+FROM mhart/alpine-node:16
 
 USER root
 
 WORKDIR /near-explorer/backend
 ENV HOME=/tmp
 COPY ./package.json ./package-lock.json ./
-RUN apk add --no-cache --virtual=.build-dependencies git && \
+RUN apk add --no-cache --virtual=.build-dependencies git python2 make g++ musl-dev && \
```
telezhnaya, line 8: It works great, but I'm curious - why it started to ask for python and C++? 😄

telezhnaya (approved):

PR #728, opened 2021-09-06
fix(backend): Fix caching of the validator stake

shelegdmitriy:
There was a problem that some 'stake' of the validating node wasn't change since several epochs. Also there are some changes in sorting of the validators. So now if some validator has the same 'stake' like some others but also has 'proposedStake' - that validator will be higher than others


frol (approved):
Overall, looks good to me.

Please, address the comment about sorting and let me merge it later this week with some other PR, so we don't reset our stats

frontend/src/components/nodes/ValidatorsList.tsx
```diff
@@ -21,6 +22,24 @@ class ValidatorsList extends Component<Props> {
...
+
+        // sort validators by 'proposedStake' if 'currentStake' of those validators are same
+        // that's why we need to round 'currentStake' to 'toFixed(0)'
+        if (
+          (a.proposedStake || b.proposedStake) &&
+          new BN(bCurrentStake).eq(new BN(aCurrentStake))
+        ) {
+          return new BN(b.proposedStake || 0).sub(new BN(a.proposedStake || 0));
+        }
+
```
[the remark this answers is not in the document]
shelegdmitriy: 1. Yeah, correct. There is no need to convert to BN

2. I've added that condition because we compare not actual validators value but rounded. That's why when I remove condition of proposedStake the validators ordering will be wrong:
Снимок экрана 2021-09-07 в 10 15 01
So let's leave it as is.

backend/src/index.js
```diff
@@ -433,24 +438,32 @@ async function main() {
...
 
-            const currentStake =
-              currentNodeStake ??
-              (await nearRpc.callViewMethod(
+            // query and update 'currentStake' for 'on-hold', 'newcomer' and 'idle' nodes
+            // because other nodes receive 'currentStake' from 'queryEpochStats()'
+            if (
+              !stakingStatus ||
+              nonValidatingNodeStatuses.indexOf(stakingStatus) >= 0
+            ) {
```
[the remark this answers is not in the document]
shelegdmitriy: We need it because when validator's stake decreased or increased we must to check is this validator is still has "idle" status? Maybe his status is "on-hold" now? And vice versa. Earlier we have checked this only at the first time

shahnawazalam0996:
Server miner for Android background fix

shelegdmitriy:
> Server miner for Android background fix

What do you mean?

backend/src/index.js
```diff
@@ -433,42 +444,60 @@ async function main() {
...
+                .callViewMethod(account_id, "get_total_staked_balance", {})
+                .catch((_error) => {
+                  // for some accounts on 'testnet' we can't get 'currentStake'
+                  // because they looks like pool accounts but they are not so
+                  // that's why we catch this error to avoid unnecessary errors in console
+                  return null;
+                });
+            }
 
+            // 'code_hash' === 11111111111111111111111111111111 refers to testnet only
```
frol, line 477:
```suggestion
            // 'code_hash' === 11111111111111111111111111111111 is when the validator does not have a staking-pool contract on it (common on testnet)
```

frontend/src/components/nodes/ValidatorsList.tsx
```diff
@@ -21,11 +21,36 @@ class ValidatorsList extends Component<Props> {
...
+        ) {
+          return new BN(b.proposedStake || 0).cmp(new BN(a.proposedStake || 0));
+        } else if (
+          a.stakingStatus &&
+          validatingGroup.indexOf(a.stakingStatus) < 0 &&
+          b.stakingStatus &&
+          validatingGroup.indexOf(b.stakingStatus) < 0
+        ) {
+          return new BN(b.currentStake).cmp(new BN(a.currentStake));
+        }
```
frol, line 49: How do you feel about this sorting? (I am not sure if `currentStake` is always defined, so I added `|| 0`)

```suggestion
        const aInValidatingGroup = validatingGroup.indexOf(a.stakingStatus) >= 0;
        const bInValidatingGroup = validatingGroup.indexOf(b.stakingStatus) >= 0;
        if (aInValidatingGroup && bInValidatingGroup) {
          return new BN(b.currentStake).cmp(new BN(a.currentStake));
        } else {
          const aStake = BN.max(new BN(b.currentStake || 0), new BN(b.proposedStake || 0));
          const bStake = BN.max(new BN(a.currentStake || 0), new BN(a.proposedStake || 0));
          return aStake.cmp(bStake);
        }
```

frol (changes requested):

frontend/src/components/nodes/__test__/__snapshots__/ValidatorsList.test.tsx.snap
```diff
@@ -130,7 +120,7 @@ Array [
           className="jsx-1785419880 total-value"
           style={
             Object {
-              "width": "-250.93%",
+              "width": "-52.17%",
```
frol, line 123: Why do we ever have negative width?

frontend/src/components/nodes/ValidatorsList.tsx
```diff
@@ -21,11 +21,37 @@ class ValidatorsList extends Component<Props> {
...
+            new BN(a.proposedStake || 0),
+            new BN(a.currentStake || 0)
+          );
+
+          if (aInValidatingGroup || bInValidatingGroup) {
+            return -1;
+          }
+
+          return aStake.cmp(bStake);
+        }
```
frol, line 50: As per our offline discussion:

* If "a" is a validator, it should be higher on the list
* If "b" is a validator, it should be higher on the list
* otherwise, we compare the stakes (notice that I flipped the cmp order in the last return statement

```suggestion
        if (aInValidatingGroup && bInValidatingGroup) {
          return new BN(b.currentStake).cmp(new BN(a.currentStake));
        } else if (aInValidatingGroup) {
          return -1;
        } else if (bInValidatingGroup) {
          return 1;
        } else {
          const aStake = BN.max(
            new BN(b.proposedStake || 0),
            new BN(b.currentStake || 0)
          );
          const bStake = BN.max(
            new BN(a.proposedStake || 0),
            new BN(a.currentStake || 0)
          );
          return bStake.cmp(aStake);
        }
```

frol (changes requested):

frol:
@shelegdmitriy The current implementation is completely off, see the Render link: https://near-explorer-frontend-with-indexer-pr-bolm.onrender.com/nodes/validators

frol (approved):

PR #729, opened 2021-09-09
refactor(frontend): Display correct error message on pages without any data

shelegdmitriy:
### Story
User tries to view transaction (already created and redirected from wallet). But the page shows default values that doesn't make any sense and get user a bit confused. This can happen by a lot of receipts which must proсessed by validators. That's why that kind of transactions will be available in rpc in some time.

Solution: show to the user correct message that transaction which will be available soon.

Also we had such a problem with accounts when we try to query to a non-existent account. For now we will show the message that account does not exist

### DIFF
--- a/frontend/src/libraries/explorer-wamp/accounts.ts
+++ b/frontend/src/libraries/explorer-wamp/accounts.ts
@@ -69,9 +69,10 @@ export default class AccountsApi extends ExplorerApi {
       }
 
       if (accountsBasicInfo.length === 0) {
-        return {
-          accountId,
-        } as AccountBasicInfo;
+        throw {
+          account_id: accountId,
+          status: 404,
+        };
       }
       const accountBasicInfo = accountsBasicInfo[0];
       return {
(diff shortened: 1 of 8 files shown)


frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -69,9 +69,10 @@ export default class AccountsApi extends ExplorerApi {
       }
 
       if (accountsBasicInfo.length === 0) {
-        return {
-          accountId,
-        } as AccountBasicInfo;
+        throw {
+          account_id: accountId,
+          status: 404,
+        };
```
frol, line 75: Meh, this looks when one subsystem (libraries) tries to "directly" communicate to error display message on another subsystem (page rendering). We should consider building some boundary here and don't expect that the message from here will directly go into the props of the page.

There is no action item right now, but I want us to be mindful of these concerns.

frol (approved):

PR #732, opened 2021-09-10
feat: API indexer-status endpoint

khorolets:


telezhnaya:
Thank you! 👍

frontend/src/pages/api/status-indexer.js
```diff
@@ -0,0 +1,27 @@
+import { getNearNetwork } from "../../libraries/config";
+import { ExplorerApi } from "../../libraries/explorer-wamp";
+import BlocksApi from "../../libraries/explorer-wamp/blocks";
+
+export default async function (req, res) {
+  // This endpoint provides the status of Indexer for Explorer
+  // It checks the latest block height in indexer database
+  // and the latest block height from JSON RPC
+  // If the difference in between 30 blocks it is considered as fine
+  // otherwise not fine
```
frol, line 10: I don't see this reflected in the code. Do I miss something?

frol (changes requested):
We should return some HTTP error status code when the diff is more than 30 blocks, so we can just plug it into the services that use HTTP status codes to reason about the service health.

frontend/src/pages/api/status-indexer.js
```diff
@@ -0,0 +1,27 @@
...
+
+  try {
+    const rpcFinalBlock = await new ExplorerApi(req).call(
+      "nearcore-final-block"
+    );
+    const indexerFinalBlock = await new BlocksApi(req).getBlocks(1, null);
+
+    const statusResponse = {
+      rpc_latest_block_height: rpcFinalBlock.header.height,
+      indexer_latest_block_height: indexerFinalBlock[0].height,
```
telezhnaya, line 20: Maybe it's better to pass timestamp? (could be in addition to height)
It will allow showing the time lag (smth like "we are late for 25 minutes")

telezhnaya (approved):

khorolets:
> We should return some HTTP error status code when the diff is more than 30 blocks, so we can just plug it into the services that use HTTP status codes to reason about the service health.

I like this idea. Reference https://en.wikipedia.org/wiki/List_of_HTTP_status_codes 

I think of 5xx as it's obviously a server error, not the client. So I can suggest either 500 Internal Server Error (which is easy to connect with third-party service) or 503 as we can consider that Service is unavailable.

So I can propose to return the data I'm returning currently + status 200 if the diff is between 30 blocks and 500/503 if diff is more than 30.  WDYT @frol @telezhnaya ?

frontend/src/pages/api/status-indexer.js
khorolets: It's not a problem yet not sure it's useful. @frol what do you think?

frol:
HTTP 503 sounds the most appropriate here

frontend/src/pages/api/status-indexer.js
frol: At least for now, I don't see a use for it, so I would not spend cycles here. If we observe a need, it will be trivial to add

frol (approved):

frol:
I will merge it later today batched with some other PRs.

PR #734, opened 2021-09-14
fix(frontend): Fix condition of TelemetryRow appearance

shelegdmitriy:
Resolve issue when "0" is shown 
image


frontend/src/components/nodes/ValidatorTelemetryRow.tsx
```diff
@@ -29,8 +29,7 @@ class ValidatorTelemetryRow extends PureComponent<Props> {
     } = this.props;
 
     const isTelemetryAvailable =
-      Boolean(producedBlocks) ||
-      Boolean(expectedBlocks) ||
+      (Boolean(producedBlocks) && Boolean(expectedBlocks)) ||
```
frol, line 32: Why does this fix work? I feel that we should display the telemetry line just fine if the node produced and/or expected to produce 0 blocks. With this check `0` will be casted to false.

frol (approved):

PR #735, opened 2021-09-15
feat(backend): start using analytics DB for stats page

telezhnaya:
We need to add new creds when publish it


telezhnaya:
For quick search:
awx job template for Mainnet
https://awx.near.org/#/templates/job_template/81?template_search=page_size:20;order_by:name;type:workflow_job_template,job_template
schedule for Mainnet (daily at 00:10 UTC)
https://awx.near.org/#/templates/job_template/81/schedules/22?template_search=page_size:20;order_by:name;type:workflow_job_template,job_template&schedule_search=page_size:20;order_by:name

telezhnaya:
All historical data is collected, bot for testnet and mainnet.

I added all needed creds to password manager.
Created readonly replicas and readonly users.

Both Mainnet and Testnet works fine, both with production and readonly credentials.

AWX report this night: https://awx.near.org/#/jobs/playbook/87285?job_search=page_size:20;order_by:-finished;not__launch_type:sync
Current 4 metrics spend 2-3 seconds each, sounds perfect.

I created separate job templates for Mainnet and Testnet, they are configured fine.
I also created cron jobs in AWX, I guess I've done it wrong, wait for @mhalambek check.

telezhnaya:
I've found the reason why do the queries fail. I guess I can fix that and we will be able to compute stats on public read-only replica.

Will start my Monday from
https://www.postgresql.org/docs/11/hot-standby.html
It should fix us other stats that are also ready but they are too heavy to recompute right now.

frontend/src/libraries/explorer-wamp/stats.ts
```diff
@@ -6,9 +6,9 @@ export interface TransactionsByDate {
   transactionsCount: number;
 }
 
-export interface TeragasUsedByDate {
+export interface GasUsedByDate {
   date: string;
-  teragasUsed: number;
+  gasUsed: number;
```
frol, line 11: Is it a number or a string?

frol (approved):

frontend/src/libraries/explorer-wamp/stats.ts
telezhnaya: Thank you for the catch! Much better to use string here, this number does not fit into JS int

PR #736, opened 2021-09-16
refactor(backend): Remove legacy code

shelegdmitriy:
As far as we have the legacy code in another branch we don't need this code here anymore. Please take a look, maybe I've missed something. It's draft for how


frontend/package.json
```diff
@@ -7,33 +7,27 @@
     "test": "npm run typecheck && jest",
     "dev": "next",
     "dev:local-wamp": "env WAMP_NEAR_EXPLORER_URL=ws://localhost:8080/ws npm run dev",
-    "dev:local-wamp-with-legacy-sync": "env NEAR_EXPLORER_DATA_SOURCE=LEGACY_SYNC_BACKEND npm run dev:local-wamp",
     "dev:local-wamp-with-indexer": "env NEAR_EXPLORER_DATA_SOURCE=INDEXER_BACKEND npm run dev:local-wamp",
```
frol, line 10: We don't need `NEAR_EXPLORER_DATA_SOURCE` anymore, so we can merge and get rid of `-with-indexer` variants to simplify things. Indexer backend is the only source of data (analytics database is not an alternative source, it is just a second dependency)

frontend/package.json
```diff
@@ -7,33 +7,27 @@
...
     "dev:local-wamp-betanet": "env 'NEAR_NETWORKS=[{\"name\": \"betanet\", \"explorerLink\": \"http://localhost:3000/\", \"aliases\": [\"localhost:3000\", \"127.0.0.1:3000\"], \"nearWalletProfilePrefix\": \"https://wallet.betanet.near.org/profile\"}]' npm run dev:local-wamp",
```
frol, line 11: Let's drop betanet for now to avoid confusion. There is no indexer for explorer for betanet yet.

frontend/src/components/transactions/Transactions.tsx
```diff
@@ -25,7 +25,7 @@ class TransactionsWrapper extends Component<OuterProps> {
     count: number,
     paginationIndexer?: T.TxPagination
   ) => {
-    return await new TransactionsApi().getTransactions({
+    return await new TransactionsApi().getTransactionsFromIndexer({
```
frol, line 28: I suggest we rename `getTransactionsFromIndexer` to `getTransactions` to avoid confusion in the future (someone might read the name and wonder "where else can I get the list of transactions if not from the indexer?")

frontend/package.json
shelegdmitriy: Do we have to do like so in `./backend/package.json` file?

frol (approved):

PR #737, opened 2021-09-16
Get rid of Teragas, use Analytics DB for daily gas stats

telezhnaya:
I'm not sure whether we really want to get rid of TGas fully
Maybe it's better to change it to `X * 10^Y` notation? Or `XeY`, C++ fans will be happy
I don't remember "Tera" or "Exa" prefixes, but the power of 10 always works fine.

Review here only the last commit.


frol:
FYI, I have created a separate branch to deploy it on mainnet/testnet to quickly populate the empty charts: https://github.com/near/near-explorer/tree/hotfix/analytics-database-backend

PR #747, opened 2021-09-28
fix(backend): fix new unique contracts count statistics

telezhnaya:


telezhnaya:
Now it looks this way

Screenshot 2021-09-28 at 18 13 13
Screenshot 2021-09-28 at 18 13 27

backend/src/stats.js
```diff
@@ -253,26 +253,18 @@ aggregateActiveContractsCountByDate = retriable(
...
+  const contractsCountByDate = await queryUniqueDeployedContractsCountAggregatedByDate();
+  let cumulativeContractsCountByDate = [];
+  let cumulative_sum = 0;
+
+  for (let i = 0; i < contractsCountByDate.length; i++) {
+    cumulative_sum += contractsCountByDate[i].contracts_count_by_date;
+    cumulativeContractsCountByDate.push({
+      date: formatDate(new Date(contractsCountByDate[i].date)),
+      contractsCount: cumulative_sum,
+    });
```
frol, line 265: Actually, we usually build the cumulative data on the front-end side: https://github.com/near/near-explorer/blob/master/frontend/src/components/stats/NewContractsByDate.tsx#L25

That way we always return non-cumulative metrics from the backend, and be able to shape it into whatever thing we need on the frontend.

I won't block us on resolving it, but please, consider sending another PR fixing this situation.

frol (approved):

backend/src/stats.js
telezhnaya: Oh, it's much easier
Thank you, will fix that

PR #751, opened 2021-09-29
fix(frontend): Fix values on circulating supply chart

### LINKED ISSUE #750: Circulating and total supply is 1000 times more than it should be
https://explorer.near.org/stats

Total supply at the beginning was 1000M
But on the chart, it's 1000 000 M (1 000 000 000 000)

Checked the data in the DB, it's fine. So it should be an error on Explorer side.

BTW, it anyway has too many zeros, let's show it in millions
@shelegdmitriy @frol what do you think?

Found another issue: we should not show this chart on Testnet, please remove it there

shelegdmitriy:
Resolves #750 

We will show this chart only for `mainnet`:

Снимок экрана 2021-09-29 в 13 52 46

### DIFF
--- a/frontend/src/components/stats/CirculatingSupplyStats.tsx
+++ b/frontend/src/components/stats/CirculatingSupplyStats.tsx
@@ -3,6 +3,7 @@ import React, { useEffect, useState } from "react";
 import { Translate } from "react-localize-redux";
 import ReactEcharts from "echarts-for-react";
 import echarts from "echarts";
+import { utils } from "near-api-js";
 
 import StatsApi from "../../libraries/explorer-wamp/stats";
 import { Props } from "./TransactionsByDate";
@@ -22,14 +23,10 @@ const CirculatingSupplyStats = ({ chartStyle }: Props) => {
           ({ date, circulatingTokensSupply, totalTokensSupply }) => ({
             date,
             circulatingTokensSupply: new BN(circulatingTokensSupply)
-              .divn(10 ** 6)
-              .divn(10 ** 6)
-              .divn(10 ** 3)
+              .div(utils.format.NEAR_NOMINATION)
               .toNumber(),
             totalTokensSupply: new BN(totalTokensSupply)
-              .divn(10 ** 6)
-              .divn(10 ** 6)
-              .divn(10 ** 3)
+              .div(utils.format.NEAR_NOMINATION)
               .toNumber(),
           })
         );
--- a/frontend/src/components/stats/CirculatingSupplyStats.tsx
+++ b/frontend/src/components/stats/CirculatingSupplyStats.tsx
@@ -24,10 +24,12 @@ const CirculatingSupplyStats = ({ chartStyle }: Props) => {
             circulatingTokensSupply: new BN(circulatingTokensSupply)
               .divn(10 ** 6)
               .divn(10 ** 6)
+              .divn(10 ** 3)
               .toNumber(),
             totalTokensSupply: new BN(totalTokensSupply)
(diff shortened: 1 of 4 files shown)


frontend/src/components/stats/CirculatingSupplyStats.tsx
```diff
@@ -24,10 +24,12 @@ const CirculatingSupplyStats = ({ chartStyle }: Props) => {
             circulatingTokensSupply: new BN(circulatingTokensSupply)
               .divn(10 ** 6)
               .divn(10 ** 6)
+              .divn(10 ** 3)
```
telezhnaya, line 27: Why do we have `10^15` here?
It should be `10^24` to convert from yocto to NEAR

shelegdmitriy: Because of this https://github.com/near/near-explorer/blob/master/backend/src/db-utils.js#L494-L506
```
SELECT
  DATE_TRUNC('day', TO_TIMESTAMP(DIV(computed_at_block_timestamp, 1000*1000*1000))) AS date,
  DIV(circulating_tokens_supply, 1000*1000*1000) AS circulating_tokens_supply,
  DIV(total_tokens_supply, 1000*1000*1000) as total_tokens_supply
FROM aggregated__circulating_supply
ORDER BY date
```

telezhnaya: Why did we decide to divide tokens with `10^9` in the select statement?

shelegdmitriy: I decide to do so because we do such convertations in other similar requests so from the component side we'll receive the same lenght of values

telezhnaya: I guess you could be confused with division for timestamps, we usually convert nanos to seconds.
Let's choose one place where we convert yocto to NEAR. I saw such conversions on JS side, so it's better to remove division for tokens in SQL statement

shelegdmitriy: Take a look ones again please 😀

frontend/src/components/stats/CirculatingSupplyStats.tsx
```diff
@@ -22,12 +22,12 @@ const CirculatingSupplyStats = ({ chartStyle }: Props) => {
           ({ date, circulatingTokensSupply, totalTokensSupply }) => ({
             date,
             circulatingTokensSupply: new BN(circulatingTokensSupply)
-              .divn(10 ** 6)
-              .divn(10 ** 6)
+              .div(new BN((10 ** 20).toString()))
+              .div(new BN((10 ** 4).toString()))
```
frol, line 26: Use `NEAR_NOMINATION` (it is BN number of 10**24): https://github.com/near/near-explorer/blob/master/frontend/src/components/nodes/NodesCard.tsx#L25

frol (changes requested):

telezhnaya (approved):
I still want to show numbers in millions (divided by `1 000 000`), it will look simpler, but it could be done in another PR

Everything looks good here for me 👍

shelegdmitriy:
> I still want to show numbers in millions (divided by `1 000 000`), it will look simpler, but it could be done in another PR
> 
> 
> 
> Everything looks good here for me 👍 

Great! Would you mind to create an issue about it and assign to me?

telezhnaya:
I mentioned it here https://github.com/near/near-explorer/issues/750
Just don't close it after merging this PR

PR #753, opened 2021-09-30
refactor(frontend): Move telemetry to database

shelegdmitriy:
Basicly our telemetry data was stored on a disk and now we want to move those data to gcp database


backend/src/db-utils.js
```diff
@@ -101,16 +104,18 @@ const queryNodeValidators = async () => {
...
-        `,
-  ]);
+  return await queryRows(
+    [
+      `SELECT ip_address AS ipAddress, account_id AS accountId, node_id AS nodeId,
+        last_seen AS lastSeen, last_height AS lastHeight,status,
+        agent_name AS agentName, agent_version AS agentVersion, agent_build AS agentBuild,
+        latitude, longitude, city
+      FROM nodes
+      WHERE last_seen > (CAST(EXTRACT(EPOCH FROM DATE_TRUNC('seconds', NOW() - INTERVAL '60 seconds')) AS bigint) * 1000)
```
shelegdmitriy, line 114: @frol  I'm not sure about this condition

frol (changes requested):
Overall, looks good to me. Please, address the comments

backend/src/db-utils.js
shelegdmitriy: I've fixed this. It seems there is no need to update `saveNodeIntoDatabase `

backend/src/db-utils.js
```diff
@@ -21,8 +22,10 @@ function getSequelize(dataSource) {
       return models.sequelizeIndexerBackendReadOnly;
     case DS_ANALYTICS_BACKEND:
       return models.sequelizeAnalyticsBackendReadOnly;
+    case DS_TELEMETRY_BACKEND:
+      return models.sequelizeTelemetryBackendReadOnly;
     default:
-      return models.sequelizeLegacySyncBackendReadOnly;
+      return models.sequelizeTelemetryBackendReadOnly;
```
frol, line 28: Let's throw an exception here

frol (approved):

PR #754, opened 2021-10-04
Rewrite error logic on accounts page

shelegdmitriy:
Previously we've talking about error handling and we decide to move error logic from component so component will only receive an error text (in our case the string will be a key from `./translations/*.json`) and immediately display it. I'm not sure about solution as we need to support the error translations so I'll leave this PR as draft.


shelegdmitriy:
We already have a better solution here https://github.com/near/near-explorer/pull/774 so I'm closing this PR

Issue #748, opened 2021-09-28
Add search by receipt id


frol:
## Story

As a developer who debugs some transaction or explores the Indexer for Explorer database, I want to quickly navigate to the transaction with the given receipt by its id so I can debug things faster.

## Acceptance Criteria

- [ ] Search by receipt id navigates a user directly to the transaction details page and scrolls down to the specific receipt (there are anchors on the page already)

PR #755, opened 2021-10-05
feat(frontend): Add search by receipt_id

shelegdmitriy:
Resolves: #748


frontend/src/libraries/explorer-wamp/receipts.ts
```diff
@@ -119,4 +119,22 @@ export default class ReceiptsApi extends ExplorerApi {
...
     }
   }
+
+  async isReceiptInTransactionIndexed(receiptId: string) {
+    try {
+      return await this.call<any>("select:INDEXER_BACKEND", [
+        `SELECT
+          receipt_id, originated_from_transaction_hash
+         FROM receipts
+         WHERE receipt_id = :receiptId`,
```
frol, line 129: It is common to use `LIMIT 1` to help DB avoid unnecessary scans after the first instance is found:

```suggestion
         WHERE receipt_id = :receiptId
         LIMIT 1`,
```

frontend/src/libraries/explorer-wamp/receipts.ts
```diff
@@ -119,4 +119,22 @@ export default class ReceiptsApi extends ExplorerApi {
       throw error;
     }
   }
+
+  async isReceiptInTransactionIndexed(receiptId: string) {
```
frol, line 123: 1. The type of the return value should be always clearly defined to check out expectations and let readers better understanding of the code without a need to read the implementation
2. The name of the function currently does not match the implementation `is*` usually imply that it will return a boolean, but it is not the case here, and we want to return the transaction hash here, so we should rename the function
3. Use the methods in explorer-wamp package to convert from snake case to camel case (`return { originatedFromTransactionHash: originated_from_transaction_hash }), so the rest of the code base can use JS code-style variable names

Issue #391, opened 2020-08-19
Implement telemetry validation


frol:
## Story

Currently, Explorer accepts telemetry reports without validation, so anyone can submit fake information about fake nodes. Nodes that do blocks/chunks production (validation) sign their telemetry data, so we should start with validating those. The telemetry from nodes that do not validate blocks/chunks have no key that is recorded on the chain, so we cannot check if they are legit nodes or fake ones; to mitigate this we may create a contract and ask users to make a function call to register the node key. Thus, we will have:

1. telemetry that is signed with the validator_key and we require the validation nodes to have their telemetry signed with that
2. telemetry that is signed with the node_key registered via in some contract
3. telemetry that is not signed, which we may still include on our list, but we may decide to filter those out if we see any abuse

shelegdmitriy:
@frol I far as I remember this issue is blocked by serialization (or something else) so do we have any mention about it?

frol:
This is blocked on someone thinking through how to address the issue that nearcore signs telemetry JSON text and adds the signature inside that JSON, and thus when Explorer parses JSON, extracts signature, and checks the rest of the payload with that signature, verification fails due to the serialization mismatch, e.g. the immediate problem was that nearcore serializes float numbers as `0.0` while JS serializes them as `0` and thus signature verification fails, but in the future, we might potentially have other problems when the fields are mixed in their order or formatting will be different.

It is not ideal, but it is not burning yet.

PR #759, opened 2021-10-11
refactor(backend): Refactored telemetry handler

shelegdmitriy:
We tried to address #391 here but failed to recreate the same telemetry payload that is being signed by the validation node due to some float values being serialized differently (e.g. `cpu_usage` gets serialized to `0.0` in Rust and `0` in JS).


frol (approved):
I have commented out the telemetry validation code since I want to land the rest of the PR

Issue #773, opened 2021-11-01
Optimize the computation of total transaction count for each account


telezhnaya:
We can reuse https://github.com/near/near-analytics to collect ingoing and outgoing transactions for each user faster.

Current solution: 
https://github.com/near/near-explorer/blob/master/frontend/src/libraries/explorer-wamp/accounts.ts#L87-L102
BTW, don't we want to move sql statement to backend? That was non-trivial to find this place.
Current solution takes ~250s for heavy account id (I took 'near').
Our plan is to combine the data from Analytics DB (for historical data) and from Indexer DB (for today's rows).

## Outgoing transactions
We have outgoing transactions collected in [Analytics](https://github.com/near/near-analytics). We can query for it, it will give us the number for all the period except the last day.
Then, we should query for transactions count for the last day (using Indexer DB) and sum up the numbers. 
It takes 2s + 2s.

## Ingoing transactions
It's a bit harder for ingoing transactions: we should join with `receipts` table there. We can't just copy-paste SQL from current solution and add `where timestamp > start_of_today`, it will still take 70s, unacceptable.

After massive playing with this query, I came up with:
```
SELECT COUNT(DISTINCT transactions.transaction_hash)
FROM transactions
LEFT JOIN receipts ON receipts.originated_from_transaction_hash = transactions.transaction_hash
    AND transactions.block_timestamp > start_of_today
WHERE receipts.included_in_block_timestamp > start_of_today
    AND transactions.signer_account_id != requested_account_id
    AND receipts.receiver_account_id = requested_account_id
``` 
This one works 1s, cool.

On the Analytics side, the query should look smth like that
```
SELECT receipts.receiver_account_id, COUNT(DISTINCT transactions.transaction_hash)
FROM transactions
LEFT JOIN receipts ON receipts.originated_from_transaction_hash = transactions.transaction_hash
    AND transactions.block_timestamp >= start_of_day1
    AND transactions.block_timestamp < start_of_day2
WHERE receipts.included_in_block_timestamp >= start_of_day1
    AND receipts.included_in_block_timestamp < (start_of_day2 + 10 minutes)
    AND transactions.signer_account_id != receipts.receiver_account_id 
GROUP BY receipts.receiver_account_id
```
It's important to add 10 minutes (the constant is configurable) to receipt border, we should pack the transaction with all their receipts together, or the numbers will not be accurate. Other receipts from the next day will be naturally ignored.
It takes 2s.

[...]

PR #774, opened 2021-11-04
refactor: Move queries to backend

shelegdmitriy:
_Resolves_ #773 
We had a mess with queries as some of them was on backend but huge amount of them on frontend

### DIFF
--- a/backend/src/chunks.js
+++ b/backend/src/chunks.js
@@ -6,12 +6,7 @@ async function getGasUsedInChunks(blockHash) {
   if (!gasUsedInChunks) {
     return undefined;
   }
-  return gasUsedInChunks
-    .reduce((currentGas, chunk) => {
-      currentGas.iadd(new BN(chunk.gas_used));
-      return currentGas;
-    }, new BN(0))
-    .toString();
+  return gasUsedInChunks;
 }
 
 exports.getGasUsedInChunks = getGasUsedInChunks;
--- a/backend/src/chunks.js
+++ b/backend/src/chunks.js
@@ -1,0 +1,17 @@
+const { queryGasUsedInChunks } = require("./db-utils");
+const BN = require("bn.js");
+
+async function getGasUsedInChunks(blockHash) {
+  const gasUsedInChunks = await queryGasUsedInChunks(blockHash);
+  if (!gasUsedInChunks) {
+    return undefined;
+  }
+  return gasUsedInChunks
+    .reduce((currentGas, chunk) => {
+      currentGas.iadd(new BN(chunk.gas_used));
+      return currentGas;
+    }, new BN(0))
+    .toString();
+}
+
+exports.getGasUsedInChunks = getGasUsedInChunks;
--- a/backend/src/db-utils.js
+++ b/backend/src/db-utils.js
@@ -203,9 +203,9 @@ const queryDashboardBlocksStats = async (options) => {
         `SELECT
           COUNT(*) AS blocks_count_60_seconds_before
         FROM blocks
-        WHERE block_timestamp > (CAST(:latestBlockTimestamp - 60 AS bigint) * 1000 * 1000 * 1000)`,
+        WHERE block_timestamp > (CAST(:latest_block_timestamp - 60 AS bigint) * 1000 * 1000 * 1000)`,
         {
-          latestBlockTimestamp:
+          latest_block_timestamp:
             dataSource == DS_INDEXER_BACKEND
               ? latestBlockEpochTimeBN.toNumber()
               : latestBlockEpochTimeBN.muln(1000).toNumber(),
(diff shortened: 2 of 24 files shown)


shelegdmitriy:
This PR is still in progress

shelegdmitriy:
@frol some tests have failed. That because backend isn't deployable for tests. I've tested it manually and it worked fine

frontend/src/pages/accounts/[id].jsx
```diff
@@ -55,13 +55,11 @@ class AccountDetail extends Component {
...
         >
-          {accountFetchingError?.status === 404 ? (
+          {accountFetchingError ? (
             <Translate
               id="page.accounts.error.account_not_found"
-              data={{ account_id: accountFetchingError.account_id }}
+              data={{ account_id: account.accountId }}
             />
-          ) : accountFetchingError ? (
-            <Translate id="page.accounts.error.account_fetching" />
```
Comment (frol) [inline frontend/src/pages/accounts/[id].jsx:64]: not-found and failure to receive a response from the server should be distinguishable as otherwise, Explorer will misreport existing users as non-existing when in fact it was just the broken backend response.

frontend/src/libraries/explorer-wamp/blocks.ts
```diff
@@ -1,147 +1,56 @@
...
-    } catch (error) {
-      console.error("Blocks.getBlocks failed to fetch data due to:");
-      console.error(error);
-      throw error;
-    }
+  ): Promise<BlocksListInfo[]> {
+    return await this.call<BlocksListInfo[]>("blocks-list", [
+      limit,
+      paginationIndexer,
+    ]);
```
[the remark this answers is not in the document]
shelegdmitriy: Does I need to move logic from `/backend` to `/explorer-wamp` and convert to proper types? Or I just need to pass variables in snake_case there?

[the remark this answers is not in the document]
shelegdmitriy: I mean in `/backend/blocks.js` for example I map through the values and convert values to proper types (mostly to number only) is it correct or I should do it in `explorer-wamp/blocks.tsx` file?

backend/src/db-utils.js
```diff
@@ -587,6 +587,115 @@ const queryAccountTransactionsCount = async (accountId) => {
...
+  ) {
+    const query = await querySingleRow(
+      [
+        `SELECT COUNT(DISTINCT transactions.transaction_hash) AS in_transactions_count
+         FROM transactions
+         LEFT JOIN receipts ON receipts.originated_from_transaction_hash = transactions.transaction_hash
+            AND transactions.block_timestamp > :last_day_collected_timestamp
+         WHERE receipts.included_in_block_timestamp > :last_day_collected_timestamp
+            AND transactions.signer_account_id != :requested_account_id
+            AND receipts.receiver_account_id = :requested_account_id`,
```
shelegdmitriy, line 676: I had a discussion about this query with @telezhnaya. She build a great strategy to collect the data in such way:
1. query for the last collected date from analytics;
2. convert date to timestamp;
3. put timestamp to this select.

This steps are needed to avoid missing data at the midnight for whole prev day.
It works, and runs super fast! (~1s for `near` account). But when I compare this query with this one
```
SELECT COUNT(DISTINCT transactions.transaction_hash)
FROM transactions
LEFT JOIN receipts ON receipts.originated_from_transaction_hash = transactions.transaction_hash
    AND transactions.block_timestamp > (CAST(EXTRACT(EPOCH FROM DATE_TRUNC('day', NOW() - INTERVAL '1 day')) AS bigint) * 1000 * 1000 * 1000)
WHERE receipts.included_in_block_timestamp > (CAST(EXTRACT(EPOCH FROM DATE_TRUNC('day', NOW() - INTERVAL '1 day')) AS bigint) * 1000 * 1000 * 1000)
    AND transactions.signer_account_id != :requested_account_id
    AND receipts.receiver_account_id = :requested_account_id
``` 
then the resulting value will differ by about 50 transactions (tested on `near` account). @frol what do you thinks about it?

backend/src/db-utils.js
```diff
@@ -361,19 +518,206 @@ const queryActiveAccountsCountAggregatedByWeek = async () => {
...
+  }
+  async function queryTransactionsCountFromIndexerForLastDay(accountId) {
+    const query = await querySingleRow(
+      [
+        `SELECT
+         COUNT(transactions.transaction_hash) AS out_transactions_count
+         FROM transactions
+         WHERE signer_account_id = :account_id
+         AND transactions.block_timestamp >= (CAST(EXTRACT(EPOCH FROM DATE_TRUNC('day', NOW() - INTERVAL '1 day')) AS bigint) * 1000 * 1000 * 1000)
+         LIMIT 1`,
```
frol: 1. Same comment about `LIMIT 1` + `COUNT` / `SUM` as above.
2. We discussed it offline, that we need to know the last computed date from analytics database in order to always compute the value correctly (otherwise, on the day boundary or when analytics database don't get updated for over a day we will have incorrect values). To query the latest date and the `SUM` in a single query you can use `MAX(collected_for_day)` next to `SUM`

backend/src/db-utils.js
```diff
@@ -550,6 +894,177 @@ const queryFirstProducedBlockTimestamp = async () => {
...
+       FROM receipts
+       WHERE receipts.included_in_block_hash = :blockHash
+       AND receipts.receipt_kind = 'ACTION'`,
+      { blockHash },
+    ],
+    { dataSource: DS_INDEXER_BACKEND }
+  );
+};
+
+const queryTransactionHashByReceiptId = async (receiptId) => {
```
frol, line 1006: I would adjust the implementation to match the name of the function (according to the name, this function should return a transaction hash string)

backend/src/db-utils.js
```diff
@@ -550,6 +894,177 @@ const queryFirstProducedBlockTimestamp = async () => {
...
+       FROM receipts
+       WHERE receipt_id = :receiptId
+       LIMIT 1`,
+      { receiptId },
+    ],
+    { dataSource: DS_INDEXER_BACKEND }
+  );
+};
+
+const queryIsTransactionIndexed = async (transactionHash) => {
```
frol, line 1020: Function name and the implementation do not match

backend/src/chunks.js
```diff
@@ -0,0 +1,17 @@
...
+  const gasUsedInChunks = await queryGasUsedInChunks(blockHash);
+  if (!gasUsedInChunks) {
+    return undefined;
+  }
+  return gasUsedInChunks
+    .reduce((currentGas, chunk) => {
+      currentGas.iadd(new BN(chunk.gas_used));
+      return currentGas;
+    }, new BN(0))
+    .toString();
```
frol, line 14: Why don't we just `SUM` on the database level?

frol (approved):

Issue #780, opened 2021-11-21
Deleted accounts should be viewable on Explorer


frol:
## Problem

Here is the account that existed, but was deleted: https://explorer.testnet.near.org/accounts/qq.frol10.testnet

https://explorer.testnet.near.org/transactions/4DAMdL6azqV39R7tN5ii9uoX8nS88kfBSV2PLZsQK1Rh

## What are we planning to do about it?

Let's fix the condition that checks for account existence, so the page shows the account regardless if it was deleted

PR #782, opened 2021-11-22
fix(backend): Fix viewing of deleted account

shelegdmitriy:
_Resolves_: #780 

Also, I find a bug with number of transactions on AccountDetails page. If account was just created, there was no any transactions at the top of page, but you could see all of them in the **Transactions** section.


frol (approved):
CI is not happy about formatting

PR #787, opened 2021-11-26
refactor(backend): Refactor queries for income and outcome tx count

### LINKED ISSUE #786: Speed up querying the outgoing tx for the account
https://github.com/near/near-explorer/blob/master/backend/src/db-utils.js#L570-L573

If we change the query to 
```
SELECT SUM(outgoing_transactions_count) AS out_transactions_count,
                MAX(collected_for_day) AS last_collected_timestamp
         FROM daily_outgoing_transactions_per_account_count
         WHERE account_id = :account_id
```

It works 0.1s instead of 3s. Let's move the work with the timestamp to the JS side?

shelegdmitriy:
Resolves #786 

From now we convert date to timestamp via js only. I've checked income and outcome transactions count on `near` account and values are same but loads faster

### DIFF
--- a/backend/src/db-utils.js
+++ b/backend/src/db-utils.js
@@ -578,10 +578,10 @@ const queryAccountOutcomeTransactionsCount = async (accountId) => {
       { dataSource: DS_ANALYTICS_BACKEND }
     );
     const lastDayCollectedTimestamp = query?.last_day_collected
-      ? (new Date(query.last_day_collected).getTime() +
-          ONE_DAY_TIMESTAMP_MILISEC) *
-        1000 *
-        1000
+      ? new BN(new Date(query.last_day_collected).getTime())
+          .add(new BN(ONE_DAY_TIMESTAMP_MILISEC))
+          .muln(10 ** 6)
+          .toString()
       : undefined;
     return {
       out_transactions_count: query?.out_transactions_count
@@ -599,7 +599,10 @@ const queryAccountOutcomeTransactionsCount = async (accountId) => {
     // we must put 'lastDayCollectedTimestamp' as below to dislay correct value
     const timestamp =
       lastDayCollectedTimestamp ||
-      (new Date().getTime() - ONE_DAY_TIMESTAMP_MILISEC) * 1000 * 1000;
+      new BN(new Date().getTime())
+        .sub(new BN(ONE_DAY_TIMESTAMP_MILISEC))
+        .muln(10 ** 6)
+        .toString();
     const query = await querySingleRow(
       [
         `SELECT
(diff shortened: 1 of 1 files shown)


backend/src/db-utils.js
```diff
@@ -568,20 +570,24 @@ const queryAccountOutcomeTransactionsCount = async (accountId) => {
...
-                MAX(CAST(EXTRACT(EPOCH FROM DATE_TRUNC('day', collected_for_day + INTERVAL '1 day')) AS bigint) * 1000 * 1000 * 1000) AS last_day_collected_timestamp
+                MAX(collected_for_day) AS last_day_collected
          FROM daily_outgoing_transactions_per_account_count
          WHERE account_id = :account_id`,
         { account_id: accountId },
       ],
       { dataSource: DS_ANALYTICS_BACKEND }
     );
+    const lastDayCollectedTimestamp = query?.last_day_collected
+      ? (new Date(query.last_day_collected).getTime() +
```
frol, line 581: What is the point of converting it to Date to only call `getTime` on it?

```
> new Date(1234).getTime()
1234
```

backend/src/db-utils.js
```diff
@@ -568,20 +570,24 @@ const queryAccountOutcomeTransactionsCount = async (accountId) => {
...
+          ONE_DAY_TIMESTAMP_MILISEC) *
+        1000 *
+        1000
```
frol, line 584: This overflows JS number type since the timestamp in nanoseconds is ~2^60, but JS can only keep the precision for the integers up to 2^53:

```
> Math.pow(2, 53) + 1
9007199254740992
> Math.pow(2, 53)
9007199254740992
```

Use BN.js

backend/src/db-utils.js
```diff
@@ -593,15 +599,18 @@ const queryAccountOutcomeTransactionsCount = async (accountId) => {
     // we must put 'lastDayCollectedTimestamp' as below to dislay correct value
     const timestamp =
       lastDayCollectedTimestamp ||
-      `(CAST(EXTRACT(EPOCH FROM DATE_TRUNC('day', NOW() - INTERVAL '1 day')) AS bigint) * 1000 * 1000 * 1000)`;
+      (new Date().getTime() - ONE_DAY_TIMESTAMP_MILISEC) * 1000 * 1000;
```
frol, line 602: I don't have hard preference at the moment, but I want to bring it to your attention that when you get the current time on JS side you start relying on the correct time being set on the server where node JS service is running while previously we relied on the time being properly set on the database side. This change should not bring any performance improvements since the timestamp value would be evaluated once on the database (compared with the case where you used it inside the `SELECT MAX(...)`)

frol (changes requested):

frol (approved):

PR #794, opened 2021-12-04
hotfix(frontend): fixed Validators page where unknown countryCode would crash the whole page

frol:
@shelegdmitriy I have hotfixed the validators page getting crashed when staking pool has unknown `countryCode` ([`BYAA`](https://en.wikipedia.org/wiki/ISO_3166-3#BYAA) is a correct country code identifier, but the JS packages we use do not recognize it, so neither country name nor flag is pulled in)

image

This is suboptimal, and we should have a better fix here, so this branch is here only for reference (I deployed this fix to production as part of the `hotfix/revert-telemetry` branch).

We should consider that staking pool metadata might even have malicious data, so we should have some strategy on how to handle it properly (e.g. check if countries-data has the information about the country and if it does not have it, we should just fall back to "unknown")


shelegdmitriy:
@frol you are right. The thing with this example that **BYAA** is in _ISO 3166-3_ format but if we pass **BY** (_Former codes_ column) to our js library - we'll see correct country flag. It's just FYI

frontend/src/components/utils/CountryFlag.tsx
```diff
@@ -20,22 +25,53 @@ const DefaultCountryFlag = () => (
...
+    typeof props.country !== undefined &&
+    countries[props.country.toUpperCase()]
+      ? props.country.toUpperCase()
+      : undefined;
+  const countryCode: string | undefined =
+    props.countryCode && typeof props.countryCode !== undefined
+      ? props.countryCode.toUpperCase()
+      : mayBeCountryCode
+      ? mayBeCountryCode
+      : undefined;
```
frol, line 41: I find this mix of conditions unreadable

```suggestion
  let countryCode: string | undefined = props.countryCode?.toUpperCase();
  if (!(countryCode in countries)) {
    const countryAsCountryCode = props.country?.toUpperCase();
    if (countryAsCountryCode in countries) {
      countryCode = countryAsCountryCode;
    }
  }
```

frontend/src/components/utils/CountryFlag.tsx
```diff
@@ -20,22 +25,53 @@ const DefaultCountryFlag = () => (
...
+  const countryLabel: string | undefined = countryCode
+    ? countries[countryCode].name
+    : props.country
+    ? props.country
+    : undefined;
```
frol, line 46:
```suggestion
  const countryLabel: string | undefined = countries[countryCode]?.name ?? props.country;
```

frontend/src/components/nodes/ValidatorMainRow.tsx
```diff
@@ -99,17 +98,11 @@ class ValidatorMainRow extends PureComponent<Props> {
...
-                      {countryCode && typeof countryCode !== undefined
-                        ? countries[countryCode.toUpperCase()].name
-                        : country}
-                    </Tooltip>
-                  }
-                >
-                  <CountryFlag countryCode={countryCode} />
-                </OverlayTrigger>
+                <CountryFlag
+                  id={`${countryCode || "country"}_${index}`}
```
frol, line 102: Can we drop this `id`?

shelegdmitriy: No we can't because of warning

shelegdmitriy (approved):

frontend/src/components/nodes/ValidatorMainRow.tsx
```diff
@@ -99,17 +98,11 @@ class ValidatorMainRow extends PureComponent<Props> {
...
-                      {countryCode && typeof countryCode !== undefined
-                        ? countries[countryCode.toUpperCase()].name
-                        : country}
-                    </Tooltip>
-                  }
-                >
-                  <CountryFlag countryCode={countryCode} />
-                </OverlayTrigger>
+                <CountryFlag
+                  id={`country_flag_${account_id}`}
```
frol, line 102:
```suggestion
                  id={`country_flag_${accountId}`}
```

PR #798, opened 2021-12-08
refactor(frontend&backend): Display `console.error` whenever it needed

shelegdmitriy:
After we moved our queries from frontend to backend we got a lot of `console.error` that was displayed on every failed request even when we control it. I've changed the structure of our errors messages for better understanding of what goes wrong.
Also I've find a cause why pagination on Blocks page was broken and fixed it as well

### DIFF
--- a/backend/src/db-utils.js
+++ b/backend/src/db-utils.js
@@ -962,7 +962,9 @@ const queryBlocksList = async (limit = 15, paginationIndexer) => {
       ORDER BY blocks.block_timestamp DESC`,
       {
         limit,
-        pagination_indexer: paginationIndexer,
+        pagination_indexer: paginationIndexer
+          ? new BN(paginationIndexer).muln(10 ** 6).toString()
+          : undefined,
       },
     ],
     { dataSource: DS_INDEXER_BACKEND }
--- a/backend/src/transactions.js
+++ b/backend/src/transactions.js
@@ -52,7 +52,7 @@ async function getTransactionsList(limit, paginationIndexer) {
     paginationIndexer
   );
   if (transactionsList.length === 0) {
-    return undefined;
+    return [];
   }
   return await createTransactionsList(transactionsList);
 }
@@ -64,7 +64,7 @@ async function getAccountTransactionsList(accountId, limit, paginationIndexer) {
     paginationIndexer
   );
   if (accountTxList.length === 0) {
-    return undefined;
+    return [];
   }
   return await createTransactionsList(accountTxList);
 }
@@ -76,7 +76,7 @@ async function getTransactionsListInBlock(blockHash, limit, paginationIndexer) {
     paginationIndexer
   );
   if (txListInBlock.length === 0) {
-    return undefined;
+    return [];
   }
   return await createTransactionsList(txListInBlock);
 }
--- a/backend/src/transactions.js
+++ b/backend/src/transactions.js
@@ -52,6 +52,8 @@ async function getTransactionsList(limit, paginationIndexer) {
     paginationIndexer
   );
   if (transactionsList.length === 0) {
+    // we should return an empty array instead of undefined
+    // to allow our ListHandler to work properly
     return [];
   }
   return await createTransactionsList(transactionsList);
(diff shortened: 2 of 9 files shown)


backend/src/transactions.js
```diff
@@ -52,7 +52,7 @@ async function getTransactionsList(limit, paginationIndexer) {
     paginationIndexer
   );
   if (transactionsList.length === 0) {
-    return undefined;
+    return [];
```
shelegdmitriy, line 55: We have to return an empty array to allow our ListHandler to work properly

backend/src/transactions.js
```diff
@@ -64,7 +64,7 @@ async function getAccountTransactionsList(accountId, limit, paginationIndexer) {
     paginationIndexer
   );
   if (accountTxList.length === 0) {
-    return undefined;
+    return [];
```
shelegdmitriy, line 67: same as here https://github.com/near/near-explorer/pull/798/files#r764806723

backend/src/transactions.js
```diff
@@ -76,7 +76,7 @@ async function getTransactionsListInBlock(blockHash, limit, paginationIndexer) {
     paginationIndexer
   );
   if (txListInBlock.length === 0) {
-    return undefined;
+    return [];
```
shelegdmitriy, line 79: the same as on line 55

backend/src/transactions.js
frol: It might have been a comment in the code, actually

frontend/src/libraries/explorer-wamp/accounts.ts
```diff
@@ -67,39 +67,44 @@ export default class AccountsApi extends ExplorerApi {
...
     } catch (error) {
       console.error(
         "AccountsApi.isAccountIndexed failed to fetch data due to:"
       );
       console.error(error);
       throw error;
     }
+    if (!isAccountExist) {
+      throw new Error(`
+        AccountsApi.isAccountIndexed: account '${accountId}' not found`);
```
frol, line 84: :heart_eyes:

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -222,109 +222,112 @@ export default class TransactionsApi extends ExplorerApi {
...
       );
       console.error(error);
       throw error;
     }
+
+    if (!transactionInfo) {
+      throw new Error(
+        `TransactionsApi.getTransactionInfo: transaction '${transactionHash}' not found`
+      );
+    } else {
```
frol, line 242: We don't need `else` since, which means that we can save 1 indentation level here :smile:

frol (approved):

frontend/src/libraries/explorer-wamp/blocks.ts
```diff
@@ -22,51 +22,54 @@ export default class BlocksApi extends ExplorerApi {
...
       throw error;
     }
+
+    if (!block) {
+      throw new Error(`BlocksApi.getBlockInfo: block '${blockId}' not found`);
+    }
+    const [gasUsed, receiptsCount] = await Promise.all([
+      this.getGasUsedInBlock(block.hash),
+      this.call<number>("receipts-count-in-block", [block.hash]),
+    ]);
```
frol, line 40: Well, we should wrap it into `try/catch` or `.catch()` function

frontend/src/libraries/explorer-wamp/transactions.ts
```diff
@@ -222,109 +222,112 @@ export default class TransactionsApi extends ExplorerApi {
...
+      const transactionExtraInfo = await this.call<any>("nearcore-tx", [
+        transactionInfo.hash,
+        transactionInfo.signerId,
+      ]);
```
frol, line 246: This might raise exceptions either due to WAMP being unavailable or RPC node behind `nearcore-tx` being unavailable, etc. We should handle those cases gracefully.

backend/src/db-utils.js
```diff
@@ -962,7 +962,9 @@ const queryBlocksList = async (limit = 15, paginationIndexer) => {
       ORDER BY blocks.block_timestamp DESC`,
       {
         limit,
-        pagination_indexer: paginationIndexer,
+        pagination_indexer: paginationIndexer
+          ? new BN(paginationIndexer).muln(10 ** 6).toString()
```
frol, line 966: Ah, I see it now, we return `DIV(blocks.block_timestamp, 1000*1000)`, and then we messed up to multiply it back. This logic is not intuitive at all, but let's merge it as is and address it some time in the future.

frol (approved):

PR #804, opened 2021-12-10
refactor(cli): removed CLI since it is outdated by now

frol:


luixo:
Probably changed should be also made to:
- `README.md` (question "How to query internal Explorer database from CLI?")
- `CODEOWNERS`

luixo (approved):

frol:
@luixo Good catch! Thanks

PR #805, opened 2021-12-10
refactor(backend): disambiguated the configuration options

frol:
@shelegdmitriy @luixo I don't expect you do extensive review here, but I still want to mention you in the PR, so you are aware of the change

Main objectives:

* I wanted to add a quick fix for telemetry database connection configuration, so it can use up to 15 connections (otherwise we cannot store 50+ telemetry requests per second of constant load)
* Along the way, I wanted to decouple read vs write credentials to telemetry database, so we can actually use the read replica for the read queries
* I decided to make the other configuration options more explicit as to what type of access they are going to be used (received a feedback from people that initially tried to use non-archival RPC server with Explorer and failed to fetch the details about transactions that were included more than 5 epoch [~2.5 days] ago)

### DIFF
--- a/backend/config/database.js
+++ b/backend/config/database.js
@@ -1,23 +1,67 @@
 module.exports = {
-  telemetryDatabase: {
+  readOnlyIndexerDatabase: {
     dialect: "postgres",
-    host: process.env.NEAR_TELEMETRY_DATABASE_HOST,
-    database: process.env.NEAR_TELEMETRY_DATABASE_NAME,
-    username: process.env.NEAR_TELEMETRY_DATABASE_USERNAME,
-    password: process.env.NEAR_TELEMETRY_DATABASE_PASSWORD,
+    host:
+      process.env.NEAR_READ_ONLY_INDEXER_DATABASE_HOST ||
+      process.env.NEAR_INDEXER_DATABASE_HOST,
+    database:
+      process.env.NEAR_READ_ONLY_INDEXER_DATABASE_NAME ||
+      process.env.NEAR_INDEXER_DATABASE_NAME,
+    username:
+      process.env.NEAR_READ_ONLY_INDEXER_DATABASE_USERNAME ||
+      process.env.NEAR_INDEXER_DATABASE_USERNAME,
+    password:
+      process.env.NEAR_READ_ONLY_INDEXER_DATABASE_PASSWORD ||
+      process.env.NEAR_INDEXER_DATABASE_PASSWORD,
+    logging: false,
   },
-  indexerDatabase: {
+  readOnlyAnalyticsDatabase: {
     dialect: "postgres",
-    host: process.env.NEAR_INDEXER_DATABASE_HOST,
-    database: process.env.NEAR_INDEXER_DATABASE_NAME,
-    username: process.env.NEAR_INDEXER_DATABASE_USERNAME,
-    password: process.env.NEAR_INDEXER_DATABASE_PASSWORD,
+    host:
+      process.env.NEAR_READ_ONLY_ANALYTICS_DATABASE_HOST ||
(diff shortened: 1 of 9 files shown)


backend/config/database.js
```diff
@@ -1,23 +1,67 @@
 module.exports = {
-  telemetryDatabase: {
+  readOnlyIndexerDatabase: {
     dialect: "postgres",
-    host: process.env.NEAR_TELEMETRY_DATABASE_HOST,
-    database: process.env.NEAR_TELEMETRY_DATABASE_NAME,
-    username: process.env.NEAR_TELEMETRY_DATABASE_USERNAME,
-    password: process.env.NEAR_TELEMETRY_DATABASE_PASSWORD,
+    host:
```
luixo, line 4: It's still pretty ambiguous for me - why the property is called readonly, while the env can be either readonly or not-readonly?
Should we call variable `NEAR_INDEXER_DATABASE_HOST` something like `NEAR_READ_WRITE_INDEXER_DATABASE_HOST`?

backend/models/index.js
```diff
@@ -5,81 +5,69 @@ const path = require("path");
...
-  dbConfig.telemetryDatabase.username,
-  dbConfig.telemetryDatabase.password,
-  {
-    ...dbConfig.telemetryDatabase,
-    logging: false,
-  }
-);
+let _exports = {};
+
+const sequelizeTelemetryBackend = dbConfig.writeOnlyTelemetryDatabase.host
```
luixo, line 11: Why the variable below called `sequelizeTelemetryBackendReadOnly`, but this one is not `sequelizeTelemetryBackendWriteOnly`?

backend/src/index.js
```diff
@@ -140,13 +140,8 @@ function startStatsAggregation() {
 async function main() {
   console.log("Starting Explorer backend...");
 
-  try {
+  if (models.sequelizeTelemetryBackend) {
```
luixo, line 143: Why won't this be throwing errors anymore?

shelegdmitriy: If I'm not mistaken that's because now it will switch to public credentials automatically which prevents throwing annoying errors to console

shelegdmitriy:
Well, I think all this refactor also caused by our attempt to configure backend & frontend preview deploy
@frol am I right?

backend/config/database.js
frol: Those are fallback arguments to keep the backwards compatibility

frol: > Should we call variable NEAR_INDEXER_DATABASE_HOST something like NEAR_READ_WRITE_INDEXER_DATABASE_HOST?

We don't write to it from Explorer side, so it is read-only.

For telemetry, we have a master database instance where we write the data when Explorer receives the telemetry requests, and then we also read the data from a read replica.

backend/models/index.js
frol: I did not want to refactor the whole world. My main concern was that we actually used the same telemetry instance for writes and reads due to re-using the same configuration options, and I actually wanted to fix that.

backend/src/index.js
frol: We only needed to catch an error here since we used the same credentials for reads and writes to telemetry database and this code attempted to create tables with read-only credentials (unless you configure it with private credentials with write access), and we simply ignored that failure since the assumption was that if we cannot create the telemetry tables, we won't be able to store the telemetry data, but that is not critical since on dev server you won't get telemetry requests anyway.

Now, in the contrast, we can simply skip even trying to initialize telemetry database if the write credentials are not set. Thus, we don't throw any scary errors at developers.

frol:
I have added objectives to the description of the PR, so it we are on the same page.

shelegdmitriy (approved):

PR #806, opened 2021-12-10
feat(backend): Update `seat_price` algorithm

### LINKED ISSUE #797: Seat price after switching to the new validator seat allocation algorithm
Protocol version 49 will introduce the new validator seat allocation algorithm, which will change the way the seat price is calculated.
Please consider updating the "Seat Price" field on the Nodes pages.

Details on how to calculate Seat Price can be found here: https://github.com/near/near-cli/issues/895

shelegdmitriy:
Resolves #797

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -943,9 +943,9 @@
       "integrity": "sha512-8ZtvEnA2c5aYCZYd1cvgdnU6cqwixRoYg70xPLWUws5ORTa/lnw+u4amixRS/Ac5U5mQVgp9pnlSUnbNWFaWZQ=="
     },
     "node_modules/near-api-js": {
-      "version": "0.44.0",
-      "resolved": "https://registry.npmjs.org/near-api-js/-/near-api-js-0.44.0.tgz",
-      "integrity": "sha512-aYJhIOf/c9zd9GSTtKHIbfu9cDdLZYPNoHl0k0i+/Kk1AFUmvtEaNcohC278yxiJfS16kOYCuqHl5u6GKh9G/g==",
+      "version": "0.44.1",
+      "resolved": "https://registry.npmjs.org/near-api-js/-/near-api-js-0.44.1.tgz",
+      "integrity": "sha512-c3kE5tBdlQJTKk9gC9CnToMnO2fSAuM/PRuAhMme2PUV+1KcRtmgczG9Q4kZi2KyeSYQFnlkyYyolz66iNhpxw==",
       "dependencies": {
         "bn.js": "5.2.0",
         "borsh": "^0.6.0",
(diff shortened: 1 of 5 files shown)


backend/src/near.js
```diff
@@ -25,12 +25,16 @@ const queryFinalBlock = async () => {
...
+  );
   const networkProtocolConfig = await nearRpc.sendJsonRpc(
     "EXPERIMENTAL_protocol_config",
     { finality: "final" }
   );
   const epochStatus = await nearRpc.sendJsonRpc("validators", [null]);
-  const numSeats =
+  const maxNumberOfSeats =
     networkProtocolConfig.num_block_producer_seats +
     networkProtocolConfig.avg_hidden_validator_seats_per_shard.reduce(
```
shelegdmitriy, line 39: @frol I just want to clarify that params `num_block_producer_seats` and  `avg_hidden_validator_seats_per_shard` in `near-cli` were received from [EXPERIMENTAL_genesis_config](https://github.com/near/near-cli/blob/master/utils/validators-info.js#L15) and not from `EXPERIMENTAL_protocol_config` as we have. I compared the calculation of `maxNumberOfSeats` from  `EXPERIMENTAL_genesis_config` and `EXPERIMENTAL_protocol_config` and both values are the same. But is there any difference or we can leave it as is?

backend/src/near.js
```diff
@@ -25,12 +25,16 @@ const queryFinalBlock = async () => {
 };
 
 const queryEpochStats = async () => {
+  const networkGenesisConfig = await nearRpc.sendJsonRpc(
+    "EXPERIMENTAL_genesis_config",
+    {}
+  );
```
frol, line 31: Let's move this query (and the dependent data below) inside the `if (currentEpochStartHeight !== epochStartHeight)` to avoid unnecessary queries unless we actually need to update the seat price (once an epoch)

backend/package-lock.json
```diff
@@ -943,9 +943,9 @@
       "integrity": "sha512-8ZtvEnA2c5aYCZYd1cvgdnU6cqwixRoYg70xPLWUws5ORTa/lnw+u4amixRS/Ac5U5mQVgp9pnlSUnbNWFaWZQ=="
     },
     "node_modules/near-api-js": {
-      "version": "0.44.0",
-      "resolved": "https://registry.npmjs.org/near-api-js/-/near-api-js-0.44.0.tgz",
-      "integrity": "sha512-aYJhIOf/c9zd9GSTtKHIbfu9cDdLZYPNoHl0k0i+/Kk1AFUmvtEaNcohC278yxiJfS16kOYCuqHl5u6GKh9G/g==",
+      "version": "0.44.1",
```
frol, line 946: We should also explicitly depend on 0.44.1 version in `package.json`

frol (approved):
Once the comments are addressed, it is good to be merged

shelegdmitriy:
Is seatPrice correct on `testnet` now?
Снимок экрана 2021-12-10 в 23 44 18

frol:
@shelegdmitriy Yes, that seems to be aligned with my expectation since on testnet there is room for 200 validators and currently we only have 94, so 106 seats are still free and the minimum stake is the sum of all the current stake * 1/6250, which is around 17k NEAR now.

PR #811, opened 2021-12-13
refactor: Upgrade jest

luixo:


frol:
> "TypeError: Cannot read properties of undefined (reading 'wampNearExplorerUrl')"

This is exactly what we observed the last time we tried to upgrade. This is a result of next config mocks not getting picked up (https://github.com/near/near-explorer/blob/32920d9b833559aaadc0ae5378e3555ecd052f7e/frontend/src/__mocks__/next/config.js)

frol:
@luixo You might find some useful tips here: https://nextjs.org/docs/testing#setting-up-jest-with-the-rust-compiler

(jest was configured quite some time ago, so you might want to revisit if we should reconfigure it from scratch)

luixo:
We'll reconfigure it from scratch, but I prefer to keep PRs atomic.
Tests failed because jest is [now](https://jestjs.io/blog/2021/05/25/jest-27#flipping-defaults) running in `node` environment instead of `jsdom` (browser) one by default.
This is considered being much faster, but it can break things if code is conditioned on browser-specific variables (e.g. `window`).
First we should switch to `jsdom` and we'll be able to use rust-compiled config pre-generated by `next.js` team.
Later we should revisit our test environment (what are we actually testing, browser? SSR?) and time spent on running test to evaluate if we should switch back to `node`.

frol (approved):

PR #814, opened 2021-12-13
CSS in Jest tests

luixo:
Before we move to [rust compiler in jest](https://nextjs.org/docs/testing#setting-up-jest-with-the-rust-compiler), we need to turn on proper `style` tag snapshotting, which is not actually relevant at the moment (we snap class name hash instead of actual CSS content).
The option in `.babelrc` will be removed as soon as we'll enable next-generated `jest` config.

### DIFF
--- a/frontend/.babelrc
+++ b/frontend/.babelrc
@@ -1,4 +1,18 @@
 {
-  "plugins": ["inline-react-svg"],
-  "presets": ["next/babel"]
+    "plugins": ["inline-react-svg"],
+    "presets": ["next/babel"],
+    "env": {
+        "test": {
+            "presets": [
+                [
+                    "next/babel",
+                    {
+                        "styled-jsx": {
+                            "babel-test": true
+                        }
+                    }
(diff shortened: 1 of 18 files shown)


luixo:
Before we move to [rust compiler in jest](https://nextjs.org/docs/testing#setting-up-jest-with-the-rust-compiler), we need to turn on proper `style` tag snapshotting, which is not actually relevant at the moment (we snap class name hash instead of actual CSS content).
The option in `.babelrc` will be removed as soon as we'll enable next-generated `jest` config.

luixo:
Actually, after moving to nextjs generated config, we'll get more noise:
- attributes like `jsx` and `global` would not be deleted from snapshots (because there will be no `babel` transformation on that)
- svg would be just a stub line instead of actual svg lines (same reason - no babel)

Rust is fast but is making DX worse at the moment :(

frol (approved):
This is a great quality-of-life improvement!

PR #815, opened 2021-12-14
fix(backend): Display validators 'stakingStatus' properly

### LINKED ISSUE #800: Fix the number of currently validating nodes
After #795 we under-report the number of validation nodes in the current epoch. It should be 61 on mainnet (the "kicked out" are still expected to produce blocks in the current epoch)

shelegdmitriy:
Resolves #800 
We have an issue when the status of the validator is not correct. This applies to validators with statuses **Kickout** and **Proposal**


frol (approved):
I feel this whole logic we ended up with cries for some major re-write. However, I would avoid spending too much time on it now, thus I am approving it.

I will merge it now, so I can see if the numbers get fixed before one of the call I have later today, where this discrepancy can be brought up.

Issue #819, opened 2021-12-14
Typescript-only project


luixo:
## Problem

I strongly believe that typescript helps maintaining any significant-sized app in a better way than javascript.
I'll keep this issue as an umbrella for all the PRs to complete migration to typescript.
At the moment project is 61.3% typescript and 38.4% javascript. Let's make it 99%+!

## PRs

- [ ] Typed components: https://github.com/near/near-explorer/pull/820
- [ ] Typed pages: https://github.com/near/near-explorer/pull/822
- [ ] Typed jest: https://github.com/near/near-explorer/pull/823
- [ ] Typed Next.js config: https://github.com/near/near-explorer/pull/824
- [ ] Typed cypress: https://github.com/near/near-explorer/pull/825
- [ ] Typed Next.js API: https://github.com/near/near-explorer/pull/827

PR #820, opened 2021-12-14
Refactor (frontend): typed-components

luixo:
Step 1 of [getting complete typescript coverage](https://github.com/near/near-explorer/issues/819)


frontend/src/components/transactions/TransactionsHeader.jsx
```diff
@@ -1,30 +0,0 @@
-import { Row, Col } from "react-bootstrap";
```
luixo, line 1: Those files were not used in the project.
Should we create an issue to have a PR check for unused modules?

frontend/src/components/utils/Balance.tsx
```diff
@@ -1,10 +1,19 @@
 /// Copied from near-wallet project:
```
luixo, line 1: Usage of `BN / string` in this file is kinda messy. We should address it later.

frontend/src/components/utils/HeaderNetworkDropdown.tsx
```diff
@@ -18,13 +24,13 @@ const HeaderDropdownItem = ({ link, title }) => {
...
 
-const HeaderNetworkDropdown = () => (
+const HeaderNetworkDropdown: FC = () => (
   <NetworkConsumer>
     {(context) => (
       <Dropdown>
         <Dropdown.Toggle className="header-network" variant="secondary">
           <div className="network-icon"></div>
-          {context.currentNearNetwork.name}
+          {context.currentNearNetwork?.name}
```
luixo, line 33: It seems like this file shouldn't be rendered if context is empty.
Though context should either provide sane fallbacks or props should be passed from the outer component.
Another issue to address later.

frontend/src/components/utils/MobileHeaderNavDropdown.tsx
```diff
@@ -59,13 +65,15 @@ const MobileNavItem = ({ link, icon, text }) => {
 
 class MobileNavDropdown extends Component {
   state = { showMenu: false };
+  dropdownWrapper: HTMLDivElement | null = null;
+  dropdownMenu: HTMLDivElement | null = null;
 
-  showMenu = (event) => {
+  showMenu = (event: MouseEvent | React.MouseEvent<HTMLDivElement>) => {
```
luixo, line 71: We have mixed input type (react / non-react events).

frontend/src/components/utils/Timer.tsx
```diff
@@ -1,8 +1,19 @@
...
 import Moment from "../../libraries/moment";
 
-class Timer extends PureComponent {
-  constructor(props) {
+interface Props {
+  time?: number;
+}
+
+interface State {
+  time: Date | number;
```
luixo, line 9: Mixed usage as well, we should make it more strict

frontend/src/components/utils/HeaderNetworkDropdown.tsx
frol: Good catch

frontend/src/components/utils/Search.tsx
```diff
@@ -79,7 +87,7 @@ class Search extends Component {
   render() {
     return (
       <form
-        onSubmit={this.handleSearch}
+        onSubmit={(e) => this.handleSearch(e)}
```
frol, line 90: Hmm, this way we will have a new function-handler on every render. Why do we need to wrap the handler?

frol (approved):
I agree with all the risen concerns

frontend/src/components/utils/Search.tsx
luixo: Yep, probably I tried to figure out things and forgot to clean up :)

shelegdmitriy (approved):

shelegdmitriy:
@luixo LGTM!

PR #822, opened 2021-12-16
Refactor (frontend): typed pages

luixo:
Step 2 of [getting complete typescript coverage](https://github.com/near/near-explorer/issues/819)


frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -173,74 +173,75 @@ class AccountDetails extends Component<Props> {
...
-                          currentNearNetwork.nearWalletProfilePrefix
-                        }
-                      />
-                    }
-                  />
-                </Col>
-              </Row>
-            )}
+            {typeof account.nonStakedBalance !== "undefined" &&
+              typeof account.stakedBalance !== "undefined" && (
```
luixo, line 177: Apparently they only come together, but we need to comply with types

frontend/src/components/utils/Content.tsx
```diff
@@ -10,8 +10,7 @@ export interface Props {
   fluid?: boolean;
   contentFluid?: boolean;
   icon?: React.ReactNode;
-  className: string;
-  children: React.ReactNode;
```
luixo, line 14: Or should I keep it to mark that children are mandatory?

frontend/src/context/DatabaseProvider.tsx
```diff
@@ -25,11 +31,7 @@ export interface DbContext {
 
 const DatabaseContext = createContext<DbContext>({});
 
-export interface Props {
-  children: React.Component;
```
luixo, line 29: The same as previous comment

frontend/src/context/NodeProvider.tsx
```diff
@@ -12,11 +12,7 @@ export interface INodeContext {
 
 const NodeContext = createContext<INodeContext>({});
 
-export interface Props {
-  children: React.Component;
```
luixo, line 16: And one more mandatory `children`

frontend/src/libraries/language.ts
```diff
@@ -27,27 +42,20 @@ function normalizeLocales(arr) {
...
+  let languageList: string[] = [];
 
   if (typeof window !== "undefined") {
     const { navigator } = window;
 
     if (navigator.languages) {
       languageList = languageList.concat(navigator.languages);
     }
+    // https://caniuse.com/mdn-api_navigator_language
+    // 97.31% users have this property as of 2021-12-15
```
luixo, line 55: Is that good enough for us? Do we have a profile of our users?

frontend/src/libraries/language.ts
```diff
@@ -88,24 +96,30 @@ export function getBrowserLocale(appLanguages) {
...
-export function getAcceptedLocale(appLanguages, acceptedLanguages) {
-  acceptedLanguages =
+export function getAcceptedLocale(
+  appLanguages: string[],
+  acceptedLanguages?: string
+): string | undefined {
+  const parsedAcceptedLanguages =
     acceptedLanguages && parseAcceptLanguage(acceptedLanguages);
 
   debugLog("Languages from Accepted Languages header:", acceptedLanguages);
```
luixo, line 106: Woops, should put `parsedAcceptedLanguages` here as well

frontend/src/libraries/language.ts
```diff
@@ -138,13 +152,14 @@ function findBestSupportedLocale(appLocales, browserLocales) {
         const existingMatch = matchedLocales[matchedPartialLocale];
 
         // Deduct a thousandth for being non-exact match.
+        const newMatchScore = 0.999 - index / walletLocales.length;
```
luixo, line 155: That's a bug that has been discovered with types.
It still exists in a wallet, we should probably report it after merging that.

frontend/src/libraries/language.ts
```diff
@@ -242,11 +266,10 @@ export function getI18nConfigForProvider({ cookies, acceptedLanguages }) {
   }
 }
 
-export function setI18N(props) {
+export function setI18N(props: LocalizeContextProps): void {
   const config = getI18nConfig();
   const { languages } = config;
-  const { cookies } = props;
```
luixo, line 248: There are actually no `cookies` in props, I need a sanity check here

frontend/src/pages/_app.tsx
```diff
@@ -35,15 +41,18 @@ class _App extends App {
...
+        currentNearNetwork = getNearNetwork(req.headers.host);
+        cookies = req.headers.cookie;
+        acceptedLanguages = req.headers["accept-language"];
+      }
     } else {
       currentNearNetwork = getNearNetwork(window.location.host);
     }
     return {
       currentNearNetwork,
-      ...(await App.getInitialProps({ ...appContext, currentNearNetwork })),
```
luixo, line 46: Why did [we put](https://github.com/near/near-explorer/commit/f6e22c0b41ec1852b04b5e8165d501a2873049f2#diff-c74b166bc48fd32a6900a92b516a36c250f268be81d57463ce342172c3148ecbL25-R39) `currentNearNetwork` here? It seems like `App.getInitialProps` doesn't do anything with it and nothing seems broken after I removed it.

frontend/src/pages/accounts/[id].tsx
```diff
@@ -14,10 +15,27 @@ import Content from "../../components/utils/Content";
...
+  accountError?: unknown;
+}
+
+class AccountDetail extends Component<Props> {
+  static async getInitialProps({
+    req,
+    query: { id: rawId },
+    res,
+  }: NextPageContext) {
+    // Change to string[] if pagename changes from `[id]` to `[...id]`
```
Comment (luixo) [inline frontend/src/pages/accounts/[id].tsx:36]: It this comment a little too obvious?

frontend/src/pages/stats/index.tsx
```diff
@@ -51,7 +51,7 @@ class Stats extends PureComponent {
           </div>
           <NetworkConsumer>
             {({ currentNearNetwork }) =>
-              currentNearNetwork.name === "mainnet" ? (
+              currentNearNetwork?.name === "mainnet" ? (
```
luixo, line 54: Those `?`s seem a little excess. We probably should make a hook which encapsulate network existence or make a default network

frontend/src/libraries/language.ts
shelegdmitriy: @luixo you can create an issue on [wallet](https://github.com/near/near-wallet)

frontend/src/pages/_app.tsx
frol: I don’t see the reason either. I must have refactored something along the way and it was a leftover from some intermediate steps. Good catch!

frontend/src/components/utils/Content.tsx
shelegdmitriy: From my point of view `children` is required everywhere exept of  `frontend/src/pages/contracts/index.jsx` but I can't find any place where we use it. @luixo can we set `FC` to `children`? With `className?: string` - agree

frontend/src/pages/accounts/[id].tsx
Comment (frol) [reply frontend/src/pages/accounts/[id].tsx:36]: Well, it is a bit hard to reason about this comment at first glance, but after a second thought it is obvious and next I thought of “why did you tell me this?”. We either need to rephrase it or just drop it (there is no meaningful use-case for […id])

frontend/src/pages/stats/index.tsx
frol: My only concern would be that we may potentially hide some bugs lurking due to the fallback always hiding non-set value, but it seems it should not be the case with the current network

frol (approved):
👍

shelegdmitriy (approved):

frontend/src/pages/accounts/[id].tsx
Comment (luixo) [reply frontend/src/pages/accounts/[id].tsx:36]: Let's remove it, actually

frontend/src/pages/stats/index.tsx
luixo: I'll make encapsulate it in a hook in a PR soon, it'll throw error if there is no current network.

frontend/src/components/utils/Content.tsx
luixo: @shelegdmitriy
Sorry, I didn't get you
I removed three excess `children: React.ReactNode` in `Context.tsx`, `DatabaseProvider.tsx` and `NodeProvider.tsx`, how is contracts page connected to that?
All three cases seem to render completely fine without children

shelegdmitriy: @luixo when I was searching for places where we use `Content` I realized that we always pass children to this component exept for [this one](https://github.com/near/near-explorer/blob/5e9807af2093c2aa85ba8ee596f860d87e30115c/frontend/src/pages/contracts/index.tsx#L12). But the thing is thet we doesn't use this page anywhere. And so we can remove this file as well. 
But nevermind, no action here

luixo: Щл

frontend/Dockerfile
```diff
@@ -20,7 +20,6 @@ ENV HOME=/tmp \
 COPY --from=build /near-explorer/production_node_modules ./node_modules
 COPY --from=build /near-explorer/frontend/public ./public
 COPY --from=build /near-explorer/frontend/next.config.js ./
-COPY --from=build /near-explorer/frontend/src/libraries/consts.js ./src/libraries/consts.js
```
luixo, line 23: @frol 
I removed `consts.js` from here, it seems like it's pretty much unneeded now

frontend/next.config.ts
```diff
@@ -10,6 +9,10 @@ export interface NearNetwork {
   nearWalletProfilePrefix: string;
 }
 
+enum DataSource {
+  LegacySyncBackend = "LEGACY_SYNC_BACKEND",
```
luixo, line 13: It's the only data source still used in client code

PR #823, opened 2021-12-16
Refactor (frontend): typed jest

luixo:
Step 3 of [getting complete typescript coverage](https://github.com/near/near-explorer/issues/819)


frontend/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "NEAR blockchain explorer",
   "scripts": {
     "typecheck": "tsc --noEmit",
-    "test": "npm run typecheck && jest",
+    "test": "npm run typecheck && TZ=UTC jest",
```
luixo, line 7: We need that for proper time faking. Do you have any ideas where I should put a comment about that?

frontend/src/components/accounts/__tests__/AccountDetails.test.tsx
```diff
@@ -1,47 +1,10 @@
...
-    );
-
-    const outTransactionsCount = 19;
-    const inTransactionsCount = 11;
-
-    wrapper.instance().state.outTransactionsCount = outTransactionsCount;
-    wrapper.instance().state.inTransactionsCount = inTransactionsCount;
-    wrapper.update();
-
-    expect(wrapper.instance().state).toBeDefined();
```
luixo, line 42: This test didn't make much sense, as it tested that `state` exist, but I don't see no way how this could be false.
We should probably implement tests with WAMP mocks

frontend/src/components/accounts/__tests__/AccountRow.test.js
```diff
@@ -1,53 +0,0 @@
...
-    );
-
-    const amount = new BN("100000000000000000000000000").add(
-      new BN("1000000000000000000000000")
-    );
-
-    wrapper.instance().state.amount = amount.toString();
-    wrapper.update();
-
-    expect(wrapper.instance().state.amount).toEqual(
```
luixo, line 27: This test also didn't make much sense - in an essence it tested proper work of BN package

frontend/package.json
frol: Yeah JSON is not great at giving ways to leave comments for humans 😞 

I don’t have anything to offer here except ugly hack of creating .test-comment key next to it and write the comment inside or have echo command before setting TZ.

Also, instead of relying on shell to support setting env variables, use env:

```suggestion
    "test": "npm run typecheck && env TZ=UTC jest",
```

frontend/src/components/accounts/__tests__/AccountDetails.test.tsx
shelegdmitriy: Agree, I believe this test is not relevant anymore after we started to collect those data from *near-analytics*

frol (approved):

shelegdmitriy (approved):

frontend/package.json
luixo: I hope a person who'll be curious why `TZ=UTC` was added will make his way to this PR for explanation.

> Also, instead of relying on shell to support setting env variables, use env

Ok, I didn't have much experience with cross-system env variables

frontend/src/components/accounts/__tests__/AccountDetails.test.tsx
luixo: Just in case - I pointed out that test didn't make sense because it didn't check anything, not because the data was fetched from a different source

PR #824, opened 2021-12-16
Refactor (frontend): typed Next.js config

luixo:
Step 4 of [getting complete typescript coverage](https://github.com/near/near-explorer/issues/819)


frontend/next.config.ts
```diff
@@ -0,0 +1,77 @@
+import { NextConfig } from "next";
```
luixo, line 1: We don't have that much of a `next.js` config to **really** need to typecheck it, but I believe it's a nice way to keep types that we use via `next/config` coherent.

frol: Makes total sense!

frol (approved):
🚀

shelegdmitriy (approved):

frontend/next.config.js
```diff
@@ -1,57 +1,19 @@
...
-      nearWalletProfilePrefix: "https://wallet.near.org/profile",
-    },
-  ];
-}
-const nearNetworkAliases = {};
-for (const nearNetwork of nearNetworks) {
-  for (const alias of nearNetwork.aliases) {
-    nearNetworkAliases[alias] = nearNetwork;
-  }
+// Jest registers ts-node for us, double run of ts-node/register will break code
```
luixo, line 1: @frol @shelegdmitriy 
JFYI that's a very tricky point with double registration of `ts-node`, if you ever stumble upon `TS Error: Emit skipped` or types mismatch in properly typed file - let me know.

PR #827, opened 2021-12-17
Refactor (frontend): Typed Next.js API

luixo:
Step 6 of [getting complete typescript coverage](https://github.com/near/near-explorer/issues/819)


frontend/src/pages/api/nodes.ts
```diff
@@ -1,12 +1,13 @@
...
 
-export default async function (req, res) {
+const handler: NextApiHandler = async (req, res) => {
   try {
     let ip_address =
       req.headers["x-forwarded-for"] ||
       req.connection.remoteAddress ||
       req.socket.remoteAddress ||
-      req.connection.socket.remoteAddress;
+      req.connection.remoteAddress;
```
shelegdmitriy, line 10: Was it just a typo?

shelegdmitriy (approved):

frontend/src/pages/api/nodes.ts
frol, line 10: There is already `req.connection.remoteAddress` two lines above. I don’t recall where I got this `req.connection.socket.remoteAddress` from, but if it does not exist, just drop it

frol (approved):

frontend/src/pages/api/nodes.ts
luixo: Ok, I'll drop it

PR #831, opened 2021-12-20
refactor(frontend): Switch to getServerSideProps

luixo:
[getInitialProps](https://nextjs.org/docs/api-reference/data-fetching/getInitialProps) are considered kinda deprecated, in favor for `getServerSideProps` and `getStaticProps`.

### DIFF
--- a/frontend/cypress/integration/account-details-navigation.spec.ts
+++ b/frontend/cypress/integration/account-details-navigation.spec.ts
@@ -17,7 +17,7 @@ context("Account Details navigation", () => {
       });
   });
 
-  it("Direct navigation to /accounts/:accountId page with account ID in upper/mixed case should return HTTP 301 (Permanent Redirect)", () => {
+  it("Direct navigation to /accounts/:accountId page with account ID in upper/mixed case should return HTTP 308 (Permanent Redirect)", () => {
     cy.fixture("mixedcase-account-ids.json")
       .as("mixedcaseAccountIds")
       .then((mixedcaseAccountId) => {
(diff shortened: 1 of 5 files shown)


frontend/cypress/integration/account-details-navigation.spec.ts
```diff
@@ -31,7 +31,7 @@ context("Account Details navigation", () => {
               if (mixedcaseAccountId === mixedcaseAccountId.toLowerCase()) {
                 expect(resp.status).to.eq(200);
               } else {
-                expect(resp.status).to.eq(301);
+                expect(resp.status).to.eq(308);
```
luixo, line 34: `Next.js` seems to be using 308 instead of 301, and the only difference is that 301 allows method to be changed to `GET` and 308 does not.
Doesn't seems relatable to this case.

frontend/src/pages/_app.tsx
```diff
@@ -157,17 +157,6 @@ const App: AppType = ({
 };
 
 App.getInitialProps = async (appContext) => {
-  // WARNING: Do not remove this getInitialProps implementation as it
```
luixo, line 160: Do we actually need that comment here? It doesn't seem so

frol: Well, it used to be a thin proxy-function implemented here, but later i18n brought some use for this method to exist, and indeed, this comment can be dropped now.

frol (approved):

shelegdmitriy (approved):

PR #835, opened 2021-12-22
Refactor (frontend): Network context

luixo:
I made working with (NEAR) network context a little more straightforward.


frontend/src/components/accounts/__tests__/AccountDetails.test.tsx
```diff
@@ -20,12 +20,6 @@ describe("<AccountDetails />", () => {
             storageUsage: "876",
             lockupAccountId: "gjturigjgkjnfidsjffjsa.lockup.near",
           }}
-          currentNearNetwork={{
```
luixo, line 23: We now have network context in every test - via `renderI18nElement` (which should be probably renamed to `renderTestElement`)

frontend/src/context/NetworkContext.tsx
```diff
@@ -0,0 +1,11 @@
+import { createContext } from "react";
+import { NearNetwork } from "next.config";
+
+export interface NetworkContext {
+  currentNetwork: NearNetwork;
+  networks: NearNetwork[];
+}
+
+export const NetworkContext = createContext<NetworkContext | undefined>(
```
luixo, line 9: We don't need an explicit provider, simple context would be fine (and minimalistic)

frontend/src/hooks/use-network-context.ts
```diff
@@ -0,0 +1,10 @@
+import { useContext } from "react";
+import { NetworkContext } from "../context/NetworkContext";
+
+export const useNetworkContext = (): NetworkContext => {
```
luixo, line 4: It could be that name `useSureNetworkContext` would be semantically better (or some other prefix explicitly stating that context should always be here)

frontend/src/libraries/config.ts
```diff
@@ -1,14 +1,29 @@
...
-export function getNearNetwork(hostname?: string): NearNetwork {
+function getHostname(req?: IncomingMessage): string | undefined {
+  if (typeof window !== "undefined") {
+    return window.location.host;
+  } else if (req) {
+    return req.headers.host;
+  }
+}
+
+export function getNearNetwork(req?: IncomingMessage): NearNetwork {
```
luixo, line 17: Actually, we have only two options on this function usage:
- we use it server-side hence we have `req`
- we use it client-side hence we have `window.location.host`

frontend/src/libraries/explorer-wamp/index.ts
```diff
@@ -27,7 +28,7 @@ export class ExplorerApi {
   dataSource: string;
   nearNetwork: NearNetwork;
 
-  constructor(apiPrefixSource?: any) {
+  constructor(req?: IncomingMessage) {
```
luixo, line 31: What was even `apiPrefixSource`? It's definitely either `req` or nothing.

frontend/src/pages/accounts/[id].tsx
```diff
@@ -64,10 +62,7 @@ const AccountDetail: NextPage<Props> = ({
             data={{ account_id: account.accountId }}
           />
         ) : (
-          <AccountDetails
-            account={account}
-            currentNearNetwork={currentNetwork}
```
Comment (luixo) [inline frontend/src/pages/accounts/[id].tsx:69]: That was weird, why not use `currentNearNetwork` in the element that needs it?

frontend/src/libraries/tester.tsx
```diff
@@ -4,6 +4,17 @@ import renderer, {
...
 import { LocalizeProvider } from "react-localize-redux";
 import LocalizeWrapper from "../components/utils/LocalizeWrapper";
+import { NetworkContext } from "../context/NetworkContext";
+
+const networkContext = {
+  currentNetwork: {
+    name: "testing",
+    explorerLink: "http://explorer/",
+    aliases: ["alias1"],
+    nearWalletProfilePrefix: "http://wallet/profile",
```
shelegdmitriy, line 14:
```suggestion
    explorerLink: "http://explorer.near.org/",
    aliases: ["alias1"],
    nearWalletProfilePrefix: "http://wallet.near.org/profile",
```

shelegdmitriy (changes requested):

frontend/src/hooks/use-network-context.ts
frol: I feel the current name + strict type is already quite explicit. I feel I would be more confused trying to parse `useSureNetworkContext`

frontend/src/libraries/config.ts
frol: I am not sold on having "get near network by incoming message". The previous implementation with "get near network by hostname" makes more sense to me.

frontend/src/libraries/tester.tsx
luixo: Apparently that doesn't affect anything as it's for test purposes only.
We'll fix that whenever we need actual addresses.

shelegdmitriy (approved):

frontend/src/libraries/config.ts
luixo: It does make sense, but also makes us duplicate extracting this hostname (not only from `req`, but also from `window.location`).
We can keep function dependent on `hostname`, but then we need another one to extract `hostname` from `req` / `window`.

What's your suggest?

PR #838, opened 2021-12-22
Refactor (frontend): mixpanel as hook

luixo:
A little bit of hooks :)

### DIFF
--- a/frontend/src/hooks/analytics/use-analytics-track-on-mount.ts
+++ b/frontend/src/hooks/analytics/use-analytics-track-on-mount.ts
@@ -1,0 +1,8 @@
+import { Dict } from "mixpanel-browser";
+import { useEffect } from "react";
+import { useAnalyticsTrack } from "./use-analytics-track";
+
+export const useAnalyticsTrackOnMount = (event: string, args?: Dict) => {
+  const track = useAnalyticsTrack();
+  return useEffect(() => track(event, args), [track, event, args]);
+};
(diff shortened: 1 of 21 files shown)


frontend/src/hooks/analytics/use-analytics-track-on-mount.ts
```diff
@@ -0,0 +1,8 @@
+import { Dict } from "mixpanel-browser";
+import { useEffect } from "react";
+import { useAnalyticsTrack } from "./use-analytics-track";
+
+export const useAnalyticsTrackOnMount = (event: string, args?: Dict) => {
```
shelegdmitriy, line 5: If I'm not mistaken we have to use `useAnalyticsTrackOnMount` on pages only? So, maybe it's better to provide an explanation above?

frontend/src/hooks/analytics/use-analytics-track.ts
```diff
@@ -0,0 +1,9 @@
+import { useCallback } from "react";
+import mixpanel, { Dict } from "mixpanel-browser";
+
+export const useAnalyticsTrack = () => {
```
shelegdmitriy, line 4: `useAnalyticsTrack` we have to use in components only?

shelegdmitriy (approved):

frontend/src/hooks/analytics/use-analytics-track-on-mount.ts
luixo: Well it depends on our goals, we may want to log not the page mount, but specific component.
At the moment it is true, but one can easily find it out by looking up hook usage.

frontend/src/hooks/analytics/use-analytics-track.ts
luixo: I didn't get the question :)
We can use it everywhere where we want track something to analytics on action.

PR #839, opened 2021-12-22
Feature (frontend): Why did you render

luixo:
The `why-did-you-render` is a very helpful tool to have to debug unnecessary rerenders.


shelegdmitriy (approved):

shelegdmitriy:
@luixo that's pretty nice improvement! May I ask you to send a screenshot with the example of responce from this library if it was? Because I can't find any...

luixo:
To enable it, add `Component.whyDidYouRender = true` after component definition. It does not trigger by default if props actually changed, only when they're deep equal, but shallowly not equal.

PR #846, opened 2022-01-02
backend: fix typo in readme

webern:
The backend README says in its title that it is the frontend README.
Surely that was a typo.


luixo (approved):

frol:
@webern Thanks for catching it!

PR #848, opened 2022-01-10
Refactor (frontend): Move i18n to i18next

luixo:
As mentioned before - we move from `react-localize-redux` to something more viable, my choice was `i18next` with `react-i18next` over it. Maintainers suggest using `next-i18next`, but I wasn't able to dedupe "i18n routes" from translations themselves (there are more minor downsides of using `next-i18next`, e.g. obligatory usage of SSR boilerplate in every page file).

The essence of the PR:
- Translation are now called by invoking `t` function extracted from `useTranslation` hook (the invocation is the same);
- Translation keys are now strongly typed, type check will fail if you try to use non-existent key;
- React elements interpolation is now different, seems more native (lookup example by key `pool_info_tip` or `...DashboardBlock.block_height.text`)
- A lot of excessive `.toString()` removed.

What is left to be done later:
- Lazy-loading translations. At the moment all of them are just bundled together. I didn't do this in this PR as it required extensive research and testing;
- Add more namespaces to minimize amount of excessive data loaded by user;
- Unify casing in translations, it is pretty messy at the moment.
- Add browser caching to reduce amount of fetched data on page reloading.
- Make a thorough review of moment usage. It was considered too heavy in a lot of projects, maybe we should switch to `date-fns` or other date-formatting libraries.


frontend/src/libraries/language.ts
```diff
@@ -81,205 +30,105 @@ export const parseAcceptLanguage = (input: string): string[] => {
...
-    browserLanguages &&
-    browserLanguages.length > 0
-  ) {
-    return findBestSupportedLocale(appLanguages, browserLanguages);
-  }
-}
-
-export function getAcceptedLocale(
-  appLanguages: string[],
+export function getAcceptedLanguage(
```
luixo, line 33: We probably should move to something more error-prone, like language detectors provided by `i18next` (in [browser](https://github.com/i18next/i18next-browser-languageDetector), or in [server](https://github.com/i18next/i18next-http-middleware#language-detection) environment)

frontend/src/libraries/language.ts
```diff
@@ -5,68 +5,17 @@
  * - https://github.com/near/near-wallet/blob/master/src/utils/getUserLocale.js
```
luixo, line 5: Please review this file changed thoroughly as well.
I don't see a reason for us to have browser-based language detection, as we'll definitely will go through `App.getInitialProps` where we get the same data (`Accept-Language` instead of `navigator.language`, `req.headers.cookie` instead of `window.cookie`) and will set the cookie for the user.

frontend/src/libraries/tester.tsx
```diff
@@ -16,17 +14,15 @@ const networkContext = {
   networks: [],
 };
 
-export const renderI18nElement = (
+export const renderElement = (
```
luixo, line 17: It seems like it's just a base providers wrapper.
We should probably refactor "base providers" out as a separate component to reuse it in real code and unit testing. Or we should not :)

frontend/src/pages/_app.tsx
```diff
@@ -14,37 +14,33 @@ import DatabaseProvider from "../context/DatabaseProvider";
...
-}) => {
+type AppType = NextComponentType<
+  AppContextType<Router>,
+  InitialProps,
+  AppPropsType & InitialProps
+>;
+
+const App: AppType = ({ Component, currentNearNetwork, language }) => {
+  if (typeof window !== "undefined" && language) {
+    initializeI18n(language);
```
luixo, line 42: This code will be called only client-side in case of first render - we'll have language in this case.
In other cases we already have i18n initialized:
- Server-side we initialize it in `getInitialProps`;
- Client-side on subsequent renders we already have everything initialized.

frontend/src/translations/en/common.json
```diff
@@ -164,8 +164,7 @@
           "title": "Latest Telemetry Update"
         },
         "node_agent_name": {
-          "text": "NEAR Protocol could have multiple implementations, so agent is the name of that implementation, where 'near-rs' is ${agent_name_url}.",
-          "url_text": "the official implementation",
+          "text": "NEAR Protocol could have multiple implementations, so agent is the name of that implementation, where 'near-rs' is <nearCoreLink>the official implementation</nearCoreLink>.",
```
luixo, line 167: Example of using custom components in translations

frontend/src/translations/vi/common.json
```diff
@@ -556,7 +552,14 @@
     },
     "ListHandler": {
       "last_block": "Chiều cao khối mới nhất là "
-    }
+    },
```
luixo, line 555: We don't have this data in vietnamese.
@frol Can we get it before merging this PR?

frontend/src/translations/zh-hans/common.json
```diff
@@ -556,7 +552,14 @@
     },
     "ListHandler": {
       "last_block": "最后的区块高度是"
-    }
+    },
```
luixo, line 555: @frol The same for chinese

frontend/src/components/transactions/__tests__/__snapshots__/ActionRow.test.tsx.snap
```diff
@@ -29,7 +29,7 @@ Array [
             <div
               className="action-row-title col"
             >
-              Called method: 'on_staking_pool_withdraw' in contract: 
+              component.transactions.ActionMessage.FunctionCall.called_method
```
luixo, line 32: I believe that using "cimode" of `i18next` (which make all translation requests evaluate to just keys) helps in unit testing, but is bad for e2e testing.
@frol @shelegdmitriy Need your opinion :)

frontend/src/libraries/language.ts
```diff
@@ -81,205 +30,106 @@ export const parseAcceptLanguage = (input: string): string[] => {
...
   }
 }
 
-function findBestSupportedLocale(
-  appLocales: string[],
-  browserLocales: string[]
-): string | undefined {
-  const matchedLocales: Record<string, Language> = {};
+function findBestSupportedLanguage(
+  availableLanguaes: readonly Language[],
```
frol, line 50: Typo:

```suggestion
  availableLanguages: readonly Language[],
```

frontend/src/libraries/language.ts
frol: > I don't see a reason for us to have browser-based language detection

Well,

1. Probably that was inherited from Wallet, which is an SPA application - we received i18n contribution from folks who also worked on i18n for Wallet
2. If we even consider to have a fallback from SSR to SPA, we would need to have browser-based language detection (yet, we may address it at that point, so we don't need to make any decisions right now)

Overall, it looks good to me so far

frontend/src/translations/vi/common.json
frol: What data do you refer to? Do you mean that some translations are lacking? I suggest we don't do anything right now since we will likely need more translation refactoring after we do UX revamp

frol (approved):
Wow, this is massive! It looks good to me!

frontend/src/translations/vi/common.json
luixo: I refer to a couple of keys I added to the json here. Ok, I'll keep text in english then.

frontend/src/components/transactions/__tests__/__snapshots__/ActionRow.test.tsx.snap
luixo: > Arguably, I can see benefit of reviewing how all the translations affect the test and whether all of them actually render proper result, so we could even consider having 1 + N snapshots (1 with raw translation keys, and N snapshots for each supported language)

I like that idea. I'll implement N more copies tests later then.

frontend/src/components/transactions/__tests__/__snapshots__/TransactionDetails.test.tsx.snap
```diff
@@ -2106,7 +2106,7 @@ exports[`<TransactionDetails /> renders with one small deposit 1`] = `
             <div
               className="ml-auto card-cell-text align-self-center col-md-12 col-auto"
             >
-              February 01, 2019 at 12:00:00am
+              co00on.5amt5_ti05.5amt5_ti05_for0amt
```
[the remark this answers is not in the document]
luixo: That's a good catch. Do you have any idea how should we still properly test this (in cimode)?

frontend/src/components/transactions/__tests__/__snapshots__/ActionRow.test.tsx.snap
shelegdmitriy: Well, for me it's fine that snapshot store the proper translation code. Even from contributor's side it's much easier to follow by than to see the translation in english (default one). LGTM

frontend/src/components/transactions/__tests__/__snapshots__/TransactionDetails.test.tsx.snap
```diff
@@ -894,7 +894,7 @@ exports[`<TransactionDetails /> renders no deposit 1`] = `
...
```
shelegdmitriy, line 897: Yeah this looks strange...

shelegdmitriy (approved):

frontend/src/components/transactions/__tests__/__snapshots__/TransactionDetails.test.tsx.snap
luixo: It's date format `common.date_time.date_time_format` instead of `MMMM DD, YYYY [at] h:mm:ssa` applied to february the first.
`co<minutes>on.<day-of-week><am-pm>t<day-of-week-locale>_ti<minute><day-of-week-locale>.<day-of-week><am-pm>t<day-of-week-locale>_ti<minute><day-of-week-locale>_for<minute><am-pm>t

PR #857, opened 2022-01-18
Fix: error pages serialization

luixo:
Apparently, we broke serialization of errors in multiple pages leading to errors (at least in dev environment).

### DIFF
--- a/frontend/src/pages/accounts/[id].tsx
+++ b/frontend/src/pages/accounts/[id].tsx
@@ -119,7 +119,7 @@ export const getServerSideProps: GetServerSideProps<
     return {
       props: {
         account: { accountId: id },
-        accountError,
+        accountError: String(accountError),
       },
     };
   }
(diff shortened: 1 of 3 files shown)


frontend/src/pages/accounts/[id].tsx
```diff
@@ -119,7 +119,7 @@ export const getServerSideProps: GetServerSideProps<
     return {
       props: {
         account: { accountId: id },
-        accountError,
+        accountError: String(accountError),
```
Comment (shelegdmitriy) [inline frontend/src/pages/accounts/[id].tsx:122]: [There](https://github.com/near/near-explorer/blob/2fb503266e5ea5f7b60b416a1d046fbaa6a01484/frontend/src/pages/accounts/%5Bid%5D.tsx#L113) must be the same

shelegdmitriy (changes requested):

frontend/src/pages/accounts/[id].tsx
Comment (luixo) [reply frontend/src/pages/accounts/[id].tsx:122]: Yup, you're right

frol (approved):

frontend/src/pages/accounts/[id].tsx
Comment (luixo) [reply frontend/src/pages/accounts/[id].tsx:122]: @shelegdmitriy Need your approve now :)

shelegdmitriy (approved):

PR #858, opened 2022-01-18
refactor(frontend): Fix forced loading of blocks

shelegdmitriy:
We had some strange behavior with forced loading blocks on Blocks page. When you click on "Refresh..." you'll see this:
Снимок экрана 2022-01-18 в 12 14 16
Снимок экрана 2022-01-18 в 12 14 20
Снимок экрана 2022-01-18 в 12 14 26

Also we had some issue with block (hash: `CYug4KVfwXqt2cNhzzPDoN4Tz4EWEsrvj1hnna8tKrUk`). This block exist but we doun't display any information about it because `getGasUsedInBlock` request return null.

### DIFF
--- a/frontend/src/components/utils/ListHandler.tsx
+++ b/frontend/src/components/utils/ListHandler.tsx
@@ -70,17 +70,13 @@ const Wrapper = <T, I>(config: StaticConfig<T, I>): FC<Props<T, I>> => {
             {!props.detailPage ? (
               <DatabaseConsumer>
                 {(context) => (
-                  <>
-                    {context.latestBlockHeight ? (
-                      <div onClick={fetch}>
-                        {config.category === "Block" ? (
-                          <Update>{`${translate(
-                            "utils.ListHandler.last_block"
-                          ).toString()}#${context.latestBlockHeight}.`}</Update>
-                        ) : null}
-                      </div>
+                  <div onClick={fetch}>
+                    {config.category === "Block" ? (
+                      <Update>{`${translate(
+                        "utils.ListHandler.last_block"
+                      ).toString()}#${context.latestBlockHeight}.`}</Update>
                     ) : null}
-                  </>
+                  </div>
                 )}
               </DatabaseConsumer>
             ) : null}
--- a/frontend/src/components/utils/ListHandler.tsx
+++ b/frontend/src/components/utils/ListHandler.tsx
@@ -28,25 +28,19 @@ const Wrapper = <T, I>(config: StaticConfig<T, I>): FC<Props<T, I>> => {
     const [hasMore, setHasMore] = useState(true);
     const [loading, setLoading] = useState(false);
 
-    const fetch = useCallback(
-      (count) => {
-        setLoading(true);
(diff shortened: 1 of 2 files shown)


frontend/src/libraries/explorer-wamp/blocks.ts
```diff
@@ -34,10 +34,14 @@ export default class BlocksApi extends ExplorerApi {
     if (!block) {
       throw new Error(`BlocksApi.getBlockInfo: block '${blockId}' not found`);
     }
-    const [gasUsed, receiptsCount] = await Promise.all([
-      this.getGasUsedInBlock(block.hash),
-      this.call<number>("receipts-count-in-block", [block.hash]),
+
+    const gasUsed = await this.getGasUsedInBlock(block.hash).catch(
```
luixo, line 38: We don't need to decouple `Promise.all` here, just add catch callback

frontend/src/components/utils/ListHandler.tsx
```diff
@@ -76,13 +70,17 @@ const Wrapper = <T, I>(config: StaticConfig<T, I>): FC<Props<T, I>> => {
             {!props.detailPage ? (
               <DatabaseConsumer>
                 {(context) => (
-                  <div onClick={fetch}>
-                    {config.category === "Block" ? (
-                      <Update>{`${translate(
-                        "utils.ListHandler.last_block"
-                      ).toString()}#${context.latestBlockHeight}.`}</Update>
+                  <>
```
luixo, line 73: Is it a visual issue? Can you please show screenshots of how does that look without adding conditional?

shelegdmitriy: Sometimes instead of number of block `undefined` is shown

shelegdmitriy: You're right, It's just a visual issue

luixo: Should we fix it separately?
I'm all in atomic changes :)

shelegdmitriy: Do you want to remove it from here and fix it by you own?

luixo: It's ok if you fix it, but it would be really nice to have it in a different PR (which should be called something like "undefined in list handler"), it would be easier to blame it later that way.

shelegdmitriy: Ok, I'll move it to another PR

frontend/src/libraries/explorer-wamp/blocks.ts
```diff
@@ -34,10 +34,12 @@ export default class BlocksApi extends ExplorerApi {
     if (!block) {
       throw new Error(`BlocksApi.getBlockInfo: block '${blockId}' not found`);
     }
+
```
luixo, line 37: The spaces are excessive

luixo (approved):

frol (approved):

PR #862, opened 2022-01-18
Stitches: proof of concept

luixo:
I consider any `css-in-js` library significantly more readable (and configurable as we'll probably have themin) than `styled-jsx`.
Beside all the basic features of `css-in-js` libs (like theming, `styled` function or SSR support), it has very thought-out [breakpoints](https://stitches.dev/docs/breakpoints) and [utils](https://stitches.dev/docs/utils) functionality.
Anyway, there is a good comparison I found on the web, [check it out](https://github.com/andreipfeiffer/css-in-js#stitches)!


frontend/src/components/accounts/AccountDetails.tsx
```diff
@@ -337,29 +353,11 @@ const AccountDetails: FC<Props> = ({ account }) => {
...
               background-color: #f8f8f8;
             }
-
-            .transaction-link-icon {
-              width: 15px;
-              margin: 0 0 12px 12px;
-            }
           `}</style>
-        </div>
+        </AccountInfoContainer>
```
frol, line 360: I don't have strong opinions about CSS libraries out there (which is why I went with the default solution in the first place), so I rely on your judgement here. Yet, worth to mention that we don't want to mix up different approaches in the same code-base, so I wonder why this `.account-card-back` was not refactored?

frontend/patches/@stitches+react+1.2.6.patch
```diff
@@ -0,0 +1,41 @@
+diff --git a/node_modules/@stitches/react/dist/index.cjs b/node_modules/@stitches/react/dist/index.cjs
```
frol: Why do we need to patch it? 🤔

frontend/cypress/integration/accounts-list-page.spec.ts
```diff
@@ -51,8 +51,8 @@ context("Accounts List page", () => {
       .should("exist")
       .first()
       .click();
-    cy.get(".account-info-container", { timeout: 5000 }).should("exist");
-    cy.get(".account-info-container .card-cell .card-body")
+    cy.get(".c-dOsIRQ", { timeout: 5000 }).should("exist");
+    cy.get(".c-dOsIRQ .card-cell .card-body")
```
frol, line 55: Are CSS identifiers stable across builds? Do changes in the CSS rules change the identifier? I can see how that could be bugging experience to update many tests after you edited CSS rules.

frontend/src/components/accounts/AccountDetails.tsx
luixo: I don't want to mix solutions as well, but I had to make a couple of changes to check out that basic features work.
This specific style was not refactored because it is used in another files as well (and I'd have to change them too), so I decided to leave it to migration step.

frontend/patches/@stitches+react+1.2.6.patch
luixo: Oh yeah, forgot to mention this one: https://github.com/modulz/stitches/issues/908
I'll open a PR in their repo later.

frontend/cypress/integration/accounts-list-page.spec.ts
luixo: Identifiers are stable across builds, as they are based on hashing styles themselves. Changing the style will change the hash.
We'll probably have more readable names [soon](https://github.com/modulz/stitches/pull/916), and we'll be able to rely not on full name (which is now `c-<hash>`), but on what name starts with (`c-My-Element-<hash>`).

luixo:
@frol Do we have any open questions on this?

luixo:
My idea is that session was closed and was not reopened later. I struggle to reproduce it now.
I believe that if it will not reproduce in a few days - merge it and warn engineers on call to revert this change in case they got this specific error.

shelegdmitriy:
I like the idea of using this library, but it seems some styles are broken. Here is how it should be:
Снимок экрана 2022-01-21 в 12 34 19
and how it looks on this branch:
Снимок экрана 2022-01-21 в 12 34 24

frol:
@luixo Thanks for addressing all my questions, I am fine with getting forward using stitches.

P.S. make sure you address the styling issues @shelegdmitriy pointed out 🙏

luixo:
> I like the idea of using this library, but it seems some styles are broken

Apparently `stitches` styles come before `bootstrap`s ones. I changed the way bootstrap css is imported, now it's as intended.

luixo:
Please, push "approve" button or the merge will be kept blocked

frol (approved):

Issue #864, opened 2022-01-19
Search does not reload the transactions list


frol:
Having Account Details page open, e.g. https://explorer.near.org/accounts/foundation.near, and trying to search for some other account (e.g. `frol.near`), the list of transactions does not get updated.

luixo:
Closed via https://github.com/near/near-explorer/pull/866

PR #866, opened 2022-01-20
Fix: transaction list does not update on block / account change

luixo:
Fixing https://github.com/near/near-explorer/issues/864

### DIFF
--- a/frontend/src/components/transactions/Transactions.tsx
+++ b/frontend/src/components/transactions/Transactions.tsx
@@ -45,6 +45,7 @@ const TransactionsWrapper: FC<OuterProps> = ({
   );
   return (
     <TransactionsList
+      key={accountId || blockHash}
       count={count}
       fetchDataFn={fetchDataFn}
       detailPage={accountId || blockHash ? true : false}
--- a/frontend/src/components/utils/ListHandler.tsx
+++ b/frontend/src/components/utils/ListHandler.tsx
@@ -40,7 +40,14 @@ const Wrapper = <T, I>(config: StaticConfig<T, I>): FC<Props<T, I>> => {
           setLoading(false);
           setShouldShow(true);
         });
-    }, [setLoading, props.count]);
+    }, [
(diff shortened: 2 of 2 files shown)


frol (approved):

frontend/src/components/transactions/Transactions.tsx
```diff
@@ -45,6 +45,7 @@ const TransactionsWrapper: FC<OuterProps> = ({
   );
   return (
     <TransactionsList
+      key={accountId || blockHash}
```
frol, line 48: I am a bit surprised to see `key` used in this context. Are you sure it is needed here?

shelegdmitriy: I just want to add that there are three transactions lists:

1. Transactions on 'Transactions' page (which is independent from any props except of `paginationIndexer` at least);
2. Transactions on Blocks page (which depends on `blockHash`);
3. Transactions on Accounts page (which depends on `accountId`);

So from my point of view even if you want to add `key` than `accountId` and `blockHash` is not enough. Why do we need it here?

shelegdmitriy (approved):

frontend/src/components/transactions/Transactions.tsx
luixo: > Transactions on 'Transactions' page (which is independent from any props except of paginationIndexer at least);

It never changes its source (`fetch` function, technically speaking), so it's not the case.

> Transactions on Blocks page (which depends on blockHash)
Transactions on Accounts page (which depends on accountId)

Both 'Blocks' and 'Accounts' page transaction lists use this exact component.

> I am a bit surprised to see key used in this context. Are you sure it is needed here?

Sure, we basically remount the whole component if we need to change transaction list source.
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key

PR #868, opened 2022-01-24
Display all included and executed receipts in particular block

### LINKED ISSUE #851: Improve clarity of when receipt was created vs executed
Receipts on the transaction details page refer to the block they were executed, but in most of the cases, users won't see the receipt itself on the block details page since the receipt is usually created in a previous block.

Let's have two sections: "Receipts Included in this Block" and "Receipts Executed in this Block"

Just for the reference:

https://explorer.near.org/transactions/G9hkwqjiuJ2RG5AM3JM4z4fJ6reJTCDpqZ48C54jYhhV#5JVXU64ZBTfGabkykJaUtiFG9LFkRcL1QZKMkiBdRaAm

shelegdmitriy:
I displayed two sections as we discussed before. But also I display one section with `executionStatus` of receipt. This is just for demo, and if it looks strange I'll remove it.

If preview link is not working here is some links for local representation:
- http://localhost:3000/blocks/9N9AsHK8BA2HtKNRn5erLhrH3wwMN6XsD6D2L5d7Hda4
- http://localhost:3000/blocks/85i6J26eCKo4Mv8SGhmUM4vBGvy99b5qZ9ArmyxycERH


backend/src/receipts.js
```diff
@@ -17,6 +18,41 @@ const INDEXER_COMPATIBILITY_RECEIPT_ACTION_KINDS = new Map([
...
+      receipts.push(receipt);
+    }
+    actions.push({
+      args: receiptAction.args,
+      kind: indexerCompatibilityTransactionActionKinds.get(receiptAction.kind),
+    });
+  }
+  return receipts;
+};
+
```
shelegdmitriy, line 55: This part is just for demo and if we'll leave this implementation I'll refactor it

backend/src/receipts.js
```diff
@@ -57,10 +93,99 @@ async function getReceiptsList(blockHash) {
...
+          receiptAction.originated_from_transaction_hash,
+        tokensBurnt: receiptAction.tokens_burnt,
+      };
+      receipts.push(receipt);
+    }
+    actions.push({
+      args: receiptAction.args,
+      kind: indexerCompatibilityTransactionActionKinds.get(receiptAction.kind),
+    });
+  }
```
shelegdmitriy, line 134: This part is just for demo and if we'll leave this implementation I'll refactor it

backend/src/receipts.js
```diff
@@ -57,10 +93,99 @@ async function getReceiptsList(blockHash) {
...
+  const receiptsIncludedInBlock = await generateProperReceiptsList(
+    await queryReceiptsList(blockHash)
+  );
+  const receiptsExecutedInBlock = await generateProperReceiptsList(
+    await queryExecutedReceiptsList(blockHash)
+  );
+  const receiptsIdsSet = new Set([
+    ...receiptsIncludedInBlock.map(({ receiptId }) => receiptId),
+    ...receiptsExecutedInBlock.map(({ receiptId }) => receiptId),
+  ]);
```
shelegdmitriy, line 152: On this stage I want to collect all included and executed receipt ids

backend/src/receipts.js
```diff
@@ -57,10 +93,99 @@ async function getReceiptsList(blockHash) {
...
+      });
+    } else if (includedReceipt && !executedReceipt) {
+      receiptsList.push({
+        ...includedReceipt,
+        executionStatus: "delayed",
+      });
+    } else {
+      receiptsList.push({
+        ...includedReceipt,
+        executionStatus: "success",
```
shelegdmitriy, line 182: My idea is to set execution statuses and display this receipts in one section. It you like the idea we could think about proper status names. For example if receipt included and executed in this block we can show nothing and just show tooltip with explaining.

backend/src/wamp.js
```diff
@@ -441,6 +441,12 @@ wampHandlers["transaction-hash-by-receipt-id"] = async ([receiptId]) => {
 wampHandlers["receipts-list-by-block-hash"] = async ([blockHash]) => {
   return await receipts.getReceiptsList(blockHash);
 };
+wampHandlers["executed-receipts-list-by-block-hash"] = async ([blockHash]) => {
+  return await receipts.getExucutedReceiptsList(blockHash);
+};
+wampHandlers["all-receipts-list-by-block-hash"] = async ([blockHash]) => {
```
shelegdmitriy, line 447: This name is not the final one

frontend/src/pages/blocks/[hash].tsx
```diff
@@ -87,9 +87,25 @@ const BlockDetail: NextPage<Props> = (props) => {
           <Content
             size="medium"
             icon={<TransactionIcon style={{ width: "22px" }} />}
-            title={<h2>{t("common.receipts.receipts")}</h2>}
+            title={<h2>Receipts included in this block</h2>}
```
Comment (luixo) [inline frontend/src/pages/blocks/[hash].tsx:90]: What's our i18n goal? Should we just drop translations?

Comment (shelegdmitriy) [reply frontend/src/pages/blocks/[hash].tsx:90]: No. For this moment it's just a demo. I'll add translations a bit later to this PR

shelegdmitriy:
As I mentioned preview links is not working properly, so here is some screenshots:
Снимок экрана 2022-01-25 в 11 04 05
<hr />
Снимок экрана 2022-01-25 в 11 04 13
<hr />
Снимок экрана 2022-01-25 в 11 04 48

Ones again: it's just my view of this representation as in my opinion it's much better to view one section with receipts and see their execution statuses then to compare two (almost identical) sections to understand which receipt was delayed or just executed in block

frol:
@shelegdmitriy I will take a look tomorrow morning. Thanks for putting all the details in the description and comments, it is much easier to review!

frol:
I like the idea of having a single list, but the semantics is hard. Here is the receipt status naming that I am thinking about:
* Executed after being delayed (too long)
* Executed
* Delayed (it might be misleading since if you don't see "executed after being delayed" on the same page, you might think that this was an execution of a delayed receipt)

The problem I see is that if we really want to solve this representation issue, we also should have a way to navigate to the block that the receipt was included (in case of "Executed after being delayed") and to the block that the receipt was eventually executed (in case of "Delayed").

I still don't have a good solution here

shelegdmitriy:
> we also should have a way to navigate to the block that the receipt was included

We can add this information to receipt here before (or instead) **Executed in Block** section:

Снимок экрана 2022-01-27 в 11 42 26

We could do the next:
- if receipt included and executed in the same block then we display "Included & Executed in Block" instead of "Executed in Block";
- if receipt included or executed then we'll display two fields "Included in Block" and "Executed in Block".

So on the block page you just click on receipt and view all the details on the **Transaction Execution Plan** section.

@frol WDYT?

shelegdmitriy:
I've created a new [PR](https://github.com/near/near-explorer/pull/893) with two receipt sections. This one I want to leave as is (or we can just close it) before we decide how to display receipts it much better.

frol:
Let's close this PR for now (#893 got the data exposed in a way that is already useful for the target audience [chain developers]), and get back to it as a reference in the future.

PR #874, opened 2022-01-27
preview: Added NearAppsUsers to the Stats page

frol:
Special PR for nf-labs folks preview

~We decided to push this chart to production to feature the fact that NearApps brings a lot of user activity which is not immediately visible since they provide permissioned access (cut the costs of gas fees for the end-users).~

@sparkyrider will update the hard-coded stats weekly by submitting new PRs


frontend/src/components/stats/ActiveNearAppsUsersByDate.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+import { Props } from "./TransactionsByDate";
+
+const ActiveNearAppsUsersByDate = ({ chartStyle }: Props) => {
+  const { t } = useTranslation();
+  const [activeAccountsByWeek, setWeekAccounts] = useState(Array());
+  const [week, setWeek] = useState(Array());
+
+  useEffect(() => {
+    // TODO(sparkyrider): Put a proper implementation in near-analytics after User Activity NEP lands and get implemented on indexer-for-explorer side.
```
luixo, line 16: Should we open an issue to keep in mind that we need to make it proper way?

frontend/src/components/stats/ActiveNearAppsUsersByDate.tsx
```diff
@@ -0,0 +1,171 @@
...
+    // Meanwhile, @sparkyrider will manually submit the new values weekly in PRs to NEAR Explorer.
+    // XXX: Hardcoded values were generated by this query:
+    // select to_timestamp(receipt_included_in_block_timestamp / 1000000000)::date as date
+    //  ,count(*) as accountscount
+    //  from action_receipt_actions
+    // where receipt_receiver_account_id = 'v1.nearapps.near'
+    //   and to_timestamp(receipt_included_in_block_timestamp / 1000000000) > NOW() - interval '60 days'
+    // group by to_timestamp(receipt_included_in_block_timestamp / 1000000000)::Date;
+    let accounts = [
```
luixo, line 25: Should it be `const`?
Let's put it out of the component to prevent creating the object on every render.

frontend/src/components/stats/ActiveNearAppsUsersByDate.tsx
```diff
@@ -0,0 +1,171 @@
...
+      {
+        date: "2022-01-16 23:59:59",
+        accountsCount: 393725,
+      },
+      {
+        date: "2022-01-23 23:59:59",
+        accountsCount: 30946,
+      },
+    ];
+    if (accounts) {
```
luixo, line 59: This condition is excessive at the moment

frontend/src/components/stats/ActiveNearAppsUsersByDate.tsx
```diff
@@ -0,0 +1,171 @@
...
+      const newAccounts = accounts.map((account: AccountsByDate) =>
```
luixo, line 60: Do we need typing?

frontend/src/pages/stats/index.tsx
```diff
@@ -61,10 +62,14 @@ const Stats: NextPage = () => {
           <NewContractsByDate chartStyle={chartStyle} />
         </div>
         <hr />
-        <div id="activeAccountsBydate">
+        <div id="activeAccountsByDate">
```
luixo, line 65: @frol 
I strongly believe that one of the most disciplining practices in coding is treating PRs as [atomic entities](https://fagnerbrack.com/one-pull-request-one-concern-e84a27dfe9f1).
Can I hope to enforce that practice in explorer repo?

luixo (changes requested):

frontend/src/pages/stats/index.tsx
frol: @luixo Sure, I will extract this change into a separate PR.

frontend/src/components/stats/ActiveNearAppsUsersByDate.tsx
```diff
@@ -0,0 +1,171 @@
...
+  useEffect(() => {
+    // TODO(sparkyrider): Put a proper implementation in near-analytics after User Activity NEP lands and get implemented on indexer-for-explorer side.
+    // Meanwhile, @sparkyrider will manually submit the new values weekly in PRs to NEAR Explorer.
+    // XXX: Hardcoded values were generated by this query:
+    // select to_timestamp(receipt_included_in_block_timestamp / 1000000000)::date as date
+    //  ,count(*) as accountscount
+    //  from action_receipt_actions
+    // where receipt_receiver_account_id = 'v1.nearapps.near'
+    //   and to_timestamp(receipt_included_in_block_timestamp / 1000000000) > NOW() - interval '60 days'
+    // group by to_timestamp(receipt_included_in_block_timestamp / 1000000000)::Date;
```
frol, line 24: @sparkyrider As I mentioned in our group chat, this query is incorrect, and thus we will need to update the figures (and the query) before we merge this into master.

This query as it is now counts every individual function call as individual user. Do you want to say that each user has only a single interaction with the chain? There should be deduplication be in place

frontend/src/components/stats/ActiveNearAppsUsersByDate.tsx
frol: Good point, I will do it before transitioning this issue to "ready for review" stage (so far there is an issue with the data itself, so we are not ready to merge it anyway)

frontend/src/components/stats/ActiveNearAppsUsersByDate.tsx
frol: Thanks for catching it! The code was based on a copy-pasted version of another component and I did not clean it up properly since I expected we would only have it in the staging preview, but things got changed and the new plan is to merge it, so indeed it needs some clean up now.

frontend/src/components/stats/ActiveNearAppsUsersByDate.tsx
sparkyrider: Fixing query & integrating some automation. A few issues to work through, but should have the new information available in the coming days.

Issue #877, opened 2022-02-01
NaN% online when the epoch just starts and there are 0 blocks expected to be produced


frol:
![image](https://user-images.githubusercontent.com/304265/151949565-48164cff-13df-499b-b39f-228848cb6742.png)

We need to come up with some decision as to what to say in such situations. The simplest thing is to say "n/a", but I feel we can do better. Probably, n/a is already better than NaN%, so we can have a quick fix there before we come up with a better solution.

PR #882, opened 2022-02-01
Fix: NaN in uptime of validator

luixo:
As mentioned in: https://github.com/near/near-explorer/issues/877


frontend/src/components/nodes/ValidatorTelemetryRow.tsx
```diff
@@ -56,7 +56,11 @@ const ValidatorTelemetryRow: FC<Props> = ({
             <Col className="validator-nodes-text uptime">
               {producedBlocks !== undefined && expectedBlocks !== undefined ? (
                 <>
-                  {((producedBlocks / expectedBlocks) * 100).toFixed(3)}% &nbsp;
+                  {expectedBlocks === 0
+                    ? "n/a"
```
shelegdmitriy, line 60: Let's display N/A as we already do in several places here.

```suggestion
                    ? t("common.state.not_available")
```

Or can just add a condition to 57 line `|| expectedBlocks === 0`.

@frol @luixo WDYT?

shelegdmitriy (approved):

frontend/src/components/nodes/ValidatorTelemetryRow.tsx
luixo: Totally makes sense
I'm not used to i18n yet

shelegdmitriy (approved):

PR #887, opened 2022-02-04
fix(backend): Fix query to execute contract info

### LINKED ISSUE #878: near account does not say that there is actually a contract deployed to it
https://explorer.near.org/accounts/near

There should be a contract section on that page, but it is missing.

image

Notice that there is contract code hash

shelegdmitriy:
Resolves #878


backend/src/db-utils.js
```diff
@@ -1111,20 +1111,99 @@ const queryExecutedReceiptsList = async (blockHash) => {
...
+    );
+  };
+
+  const {
+    latest_updated_timestamp: latestUpdatedTimestamp,
+  } = await queryLatestContractUpdate();
+
+  // this query is needed to find the fresh contract update
+  // if it return 'undefined' then there was no update since deployed_at_block_timestamp
+  const queryContractInfoFromIndexer = async () => {
```
shelegdmitriy, line 1168: The thing is this query is running for quite long time (for `near` account as example it takes up to 30s) and you have to wait. We can solve this by returning value from analytics db (it runs very fast) but how about latest contract update if it was?

shelegdmitriy: We can split it into two queries:
 - the first query from analytics to find and return contract info as fast as we can;
 - the second one to query from indexer db
All this queries will be runned one by one from the component. If there is some fresh data the section with contract will be updated.
But I'm not sure. @frol WDYT?

backend/src/db-utils.js
```diff
@@ -1111,20 +1111,99 @@ const queryExecutedReceiptsList = async (blockHash) => {
...
-       WHERE action_receipt_actions.action_kind = 'DEPLOY_CONTRACT' AND action_receipt_actions.receipt_receiver_account_id = :account_id
-       ORDER BY action_receipt_actions.receipt_included_in_block_timestamp DESC
-       LIMIT 1`,
-      { account_id: accountId },
-    ],
-    { dataSource: DS_INDEXER_BACKEND }
-  );
+  let contractInfo;
+  // query to analytics db to find latest historical record
+  const queryLatestHistoricalInfo = async () => {
```
frol, line 1116: 1. Fix the order of requests as per offline discussion
2. Avoid unnecessary functions

backend/src/db-utils.js
```diff
@@ -1111,20 +1111,99 @@ const queryExecutedReceiptsList = async (blockHash) => {
...
+  };
+
+  const indexerContractInfo = await queryContractInfoFromIndexer();
+
+  if (!indexerContractInfo) {
+    const { hash } = await queryDeployedContractTransactionHash();
+    contractInfo = {
+      block_timestamp: deployedAtBlockTimestamp
+        ? Number(deployedAtBlockTimestamp) / (1000 * 1000)
+        : undefined,
```
frol, line 1197: Move div to the query (database)

frol (changes requested):

luixo:
I still [don't see](https://near-explorer-experimental-frontend-with-i3mp.onrender.com/accounts/near) contract info on `near` account. Should I?

shelegdmitriy:
> don't see

No, because we don't deploy backend on preview. You can only see it on local machine

frol:
@luixo Just for the reference, I checked it on localhost:

image

PR #893, opened 2022-02-09
feat(frontend): Expose all included and executed receipts in particular block

shelegdmitriy:
Partially resolves: #851 

This PR is a part of disscussion of [another PR](https://github.com/near/near-explorer/pull/868)

### DIFF
--- a/backend/src/accounts.js
+++ b/backend/src/accounts.js
@@ -63,7 +63,7 @@ async function getAccountActivity(accountId) {
   if (!accountActivity) {
     return undefined;
   }
-  const indexerCompatibilityTransactionActionKinds = await getIndexerCompatibilityTransactionActionKinds();
+  const indexerCompatibilityTransactionActionKinds = getIndexerCompatibilityTransactionActionKinds();
   return accountActivity.map((activity) => ({
     timestamp: activity.timestamp,
     updateReason: activity.update_reason,
--- a/backend/src/receipts.js
+++ b/backend/src/receipts.js
@@ -2,6 +2,7 @@ const {
   queryReceiptsCountInBlock,
   queryReceiptInTransaction,
   queryReceiptsList,
+  queryExecutedReceiptsList,
 } = require("./db-utils");
 
 const BN = require("bn.js");
@@ -21,8 +22,7 @@ async function getIndexerCompatibilityReceiptActionKinds() {
   return INDEXER_COMPATIBILITY_RECEIPT_ACTION_KINDS;
 }
 
-async function getReceiptsList(blockHash) {
-  const receiptActions = await queryReceiptsList(blockHash);
+async function generateProperReceiptsList(receiptsList) {
   // The receipt actions are ordered in such a way that the actions for a single receipt go
   // one after another in a correct order, so we can collect them linearly using a moving
   // window based on the `previousReceiptId`.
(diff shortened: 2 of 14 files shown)


backend/src/receipts.js
```diff
@@ -21,46 +22,57 @@ async function getIndexerCompatibilityReceiptActionKinds() {
...
-async function getReceiptsList(blockHash) {
-  const receiptActions = await queryReceiptsList(blockHash);
+async function generateProperReceiptsList(receiptsList) {
   // The receipt actions are ordered in such a way that the actions for a single receipt go
   // one after another in a correct order, so we can collect them linearly using a moving
   // window based on the `previousReceiptId`.
   let receipts = [];
   let actions;
   let previousReceiptId = "";
   const indexerCompatibilityTransactionActionKinds = await getIndexerCompatibilityTransactionActionKinds();
```
frol, line 32: BTW, it does not need to be async

backend/src/receipts.js
```diff
@@ -21,46 +22,57 @@ async function getIndexerCompatibilityReceiptActionKinds() {
   return INDEXER_COMPATIBILITY_RECEIPT_ACTION_KINDS;
 }
 
-async function getReceiptsList(blockHash) {
-  const receiptActions = await queryReceiptsList(blockHash);
+async function generateProperReceiptsList(receiptsList) {
```
frol, line 25: "proper" is subjective and not descriptive. I suggest `groupReceiptActionsIntoReceipts` and the input argument should be `receiptActions` (while the return value will be a list of receipts)

frontend/src/components/receipts/ReceiptsIncludedInBlock.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { useWampSimpleQuery } from "../../hooks/wamp";
+import ReceiptsList from "./ReceiptsList";
+
+interface Props {
+  blockHash: string;
+}
+
+const ReceiptsIncludedInBlock: FC<Props> = ({ blockHash }) => {
+  const receiptsList = useWampSimpleQuery("receipts-list-by-block-hash", [
```
frol, line 11: Let's rename the wamp endpoint to `included-receipts-list-by-block-hash` to avoid confusion in the future

frontend/src/translations/ru/common.json
```diff
@@ -311,12 +311,14 @@
...
-      }
-    },
-    "blocks": {
-      "ReceiptsInBlock": {
+      },
+      "ReceiptsList": {
         "no_receipts": "В этом блоке нет квитанций."
       },
+      "ReceiptsIncludedInBlock": "Квитанции добавлены в блок",
+      "ReceiptsExecutedInBlock": "Квитанции выполнены в блоке"
```
frol, line 319:
```suggestion
      "ReceiptsIncludedInBlock": "Квитанции включённые в блок",
      "ReceiptsExecutedInBlock": "Квитанции выполненные в блоке"
```

frol (changes requested):

frol (approved):

frontend/src/components/receipts/ReceiptsList.tsx
```diff
@@ -5,18 +5,14 @@ import Placeholder from "../utils/Placeholder";
 import PaginationSpinner from "../utils/PaginationSpinner";
 
 import { useTranslation } from "react-i18next";
-import { useWampSimpleQuery } from "../../hooks/wamp";
+import { Receipt } from "../../libraries/wamp/types";
 
 interface Props {
-  blockHash: string;
+  receiptsList?: Receipt[];
```
luixo, line 11: Why is that optional?

frol:
@luixo This feature is a valuable addition to the Block Details page, which helps nearcore chain team to be informed about the pieces that were included and executed in a given block. It does not need to be beautiful, so I believe it is ready to be merged (unless there are concerns about the implementation)

frontend/src/components/receipts/ReceiptsList.tsx
shelegdmitriy: Because on `ReceiptsList` component we show loader if `receiptsList: undefined` (on the page load) otherwise an array. Also `useWampSimpleQuery` returns `T | undefined` so in my opinion it's ok. Maybe it's better to return an empty array as an initial value but then the placeholder message will keep popping up and only after some time will be replaced by data

backend/src/accounts.js
```diff
@@ -63,7 +63,7 @@ async function getAccountActivity(accountId) {
   if (!accountActivity) {
     return undefined;
   }
-  const indexerCompatibilityTransactionActionKinds = await getIndexerCompatibilityTransactionActionKinds();
+  const indexerCompatibilityTransactionActionKinds = getIndexerCompatibilityTransactionActionKinds();
```
luixo, line 66: I believe this kind of fixes should be made outside of main PR (and they can easily be merged without waiting for owner-approve)

backend/src/receipts.js
```diff
@@ -61,6 +61,18 @@ async function getReceiptsList(blockHash) {
   return receipts;
 }
 
+// As a temporary solution we split receipts list into two lists:
```
luixo, line 64: What would be a permanent solution?

luixo (approved):

backend/src/receipts.js
shelegdmitriy: One of possible solutions is here https://github.com/near/near-explorer/pull/868. I'll add a link to this PR to this comment

PR #897, opened 2022-02-12
Fix (frontend): outdated dashboard link

tooBusyNow:
image

Dashboard [link to docs](https://docs.near.org/docs/validator/staking#run-the-node) seems to be outdated, so changed it to follow [here](https://docs.near.org/docs/develop/node/intro/what-is-a-node)


luixo (approved):
Thanks for catching that!

PR #899, opened 2022-02-15
Fix: Infinite loop on BN.js null instead of 0

luixo:
Resolving bug introduced by `bn.js`: https://github.com/indutny/bn.js/issues/186

### DIFF
--- a/frontend/src/providers/blocks.ts
+++ b/frontend/src/providers/blocks.ts
@@ -15,12 +15,12 @@ export const getBlock = async (
     return;
   }
   const [gasUsed, receiptsCount] = await Promise.all([
-    wampCall("gas-used-in-chunks", [blockInfo.hash]).catch(() => "0"),
+    wampCall("gas-used-in-chunks", [blockInfo.hash]),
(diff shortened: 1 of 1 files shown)


shelegdmitriy:
> I believe something's broken on Render integration side as I changed its settings extensively. I'll look into this.

Ok, I'll check this locally, not a big deal

shelegdmitriy:
> 49MysHDCtYNQsLXbuDksbFgYthKZHN6GANRgqJqqzAdN

Yeah, that's exactly what I vanted to check because we had some issue (which I can't find) with this. But it works fine now

shelegdmitriy (approved):

frol (approved):
I am ok with this as a quick fix, but let's have a TODO properly handling the error since with this behavior we risk to misinform users about the actual gas usage

luixo:
Before we had this changed the page blew with 500, I believe we have an improvement :)

luixo:
> Yeah, that's exactly what I wanted to check because we had some issue (which I can't find) with this. But it works fine now

It's testnet by the way

PR #910, opened 2022-03-08
No ambiguous types in Header on MouseEvent

luixo:
The types are actually pretty straight

### DIFF
--- a/frontend/src/components/utils/MobileHeaderNavDropdown.tsx
+++ b/frontend/src/components/utils/MobileHeaderNavDropdown.tsx
@@ -128,24 +128,21 @@ const MobileNavDropdown = () => {
   const dropdownWrapperRef = useRef<HTMLDivElement>(null);
   const dropdownMenuRef = useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/utils/MobileHeaderNavDropdown.tsx
```diff
@@ -160,8 +157,6 @@ const MobileNavDropdown = () => {
   useEffect(() => {
     if (isMenuShown) {
       document.addEventListener("click", closeMenu);
-    } else {
-      document.removeEventListener("click", closeMenu);
```
frol, line 164: Why do you believe this should be removed?

frol (approved):
Not all the changes are related to the topic and I wonder why that change is necessary, but otherwise I approve the change

frontend/src/components/utils/MobileHeaderNavDropdown.tsx
luixo: Listener is removed on effect destructor, so every time the effect is called (as well as on unmount) - previous listener is cleaned up.

PR #911, opened 2022-03-08
No "as" in next/link

luixo:
It doesn't seem we need this as per `Next 9.5.3`: https://nextjs.org/docs/api-reference/next/link


frol (approved):
Good point

shelegdmitriy (approved):

PR #914, opened 2022-03-08
Always use React.memo

luixo:
This PR should raise a discussion about using namespaces.
Should it be
```
import * as React from 'react';
const MyComp: React.FC = () => {...};
```
or
```
import {FC} from 'react';
const MyComp: FC = () => {...};
```
?

The chosen option should be enforced by linter.
@shelegdmitriy any ideas?

### DIFF
--- a/frontend/additional.d.ts
+++ b/frontend/additional.d.ts
@@ -1,9 +1,4 @@
 // See https://nextjs.org/docs/basic-features/typescript for more information.
 
-declare module "*.svg" {
-  const content: React.FC;
-  export default content;
-}
-
 declare module "react-text-collapse";
 declare module "json-2-prom";
--- a/frontend/src/components/accounts/AccountDetails.tsx
+++ b/frontend/src/components/accounts/AccountDetails.tsx
@@ -45,7 +45,7 @@ export interface Props {
   };
 }
 
-const AccountDetails: React.FC<Props> = ({ account }) => {
+const AccountDetails: React.FC<Props> = React.memo(({ account }) => {
   const { t } = useTranslation();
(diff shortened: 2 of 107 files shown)


luixo:
Also, an article on topic that explains why we should not skip `memo`: https://attardi.org/why-we-memo-all-the-things/

frol (approved):
I am fine with using `memo` everywhere, but the way we currently apply it over the whole function body makes it hard to read the props list, and sometimes adds another indentation level. In case of custom equality rules passed to `memo` as a second argument, it will make things even worse. Would you consider applying `memo` at the bottom of the file on export?

frol:
> This PR should raise a discussion about using namespaces.

I usually prefer explicit namespacing (`React.FC`) to ease reading, but with tree-shaking JS seems to need to import selectively (unless some heuristics were already implemented and I have outdated information).

luixo:
> Would you consider applying memo at the bottom of the file on export?

I have no problem with that except the fact we'll probably want a eslint rule for memo and tracking `memo`s on export is harder than tracking them inline.

> I usually prefer explicit namespacing (React.FC) to ease reading, but with tree-shaking JS seems to need to import selectively (unless some heuristics were already implemented and I have outdated information).

Yep, it's outdated, even nested import tree-shaking is [supported](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#nested-tree-shaking) in Webpack nowadays.

shelegdmitriy (approved):
I'm fine with this option totally:
```
import * as React from 'react';
const MyComp: React.FC = () => {...};
```

PR #915, opened 2022-03-08
Remove Legacy Backend support

luixo:
@frol 
I can't find a trace of using this legacy backend feature in code, so I removed it. Please point out if I missed something out.

Also, please point out how `NEAR_IS_INDEXER_BACKEND_ENABLED` variable is used, is it "false" on production and if so - what is the reason for that?


frol (approved):

frol:
> Also, please point out how NEAR_IS_INDEXER_BACKEND_ENABLED variable is used, is it "false" on production and if so - what is the reason for that?

It is, actually, set to `true` in production.

We originally left the switch in the backend implementation, so in case we would ever need to migrate to yet another backend (e.g. SingleStore instead of PostgreSQL), we could just add it. Yet, now I think it creates more burden than it solves problems, so I am fine with dropping this support now.

PR #939, opened 2022-04-06
Fix: validators fee as null

luixo:
I managed to reproduced the [problem](https://github.com/near/near-explorer/pull/906#issuecomment-1084742726) (kudos to @frol) and fixed it.

The bug was created by migration to typescript.

I decided to make types more semantical - and separate `stakingPoolInfo` to a nested object, removing the need to have types like `Foo | undefined | null`.

### DIFF
--- a/frontend/src/libraries/wamp/types.ts
+++ b/frontend/src/libraries/wamp/types.ts
@@ -39,6 +39,22 @@ export interface ValidationProgress {
   };
 }
 
+export type ValidationNodeInfo = {
+  account_id: string;
+  is_slashed?: boolean;
+  progress?: ValidationProgress;
+  public_key?: string;
+  proposedStake?: string;
+  cumulativeStakeAmount?: string;
+  stakingStatus?: StakingStatus;
+  networkHolder?: boolean;
+  shards?: number[];
+  nodeInfo?: NodeInfo;
+  stakingPoolInfo?: StakingPoolInfo;
+  currentStake?: string;
+  poolDetails?: PoolDetails;
+};
+
 export interface PoolDetails {
   country?: string;
   country_code?: string;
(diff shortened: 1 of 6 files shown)


frontend/src/libraries/wamp/types.ts
```diff
@@ -38,19 +38,7 @@ export interface ValidationProgress {
...
-  public_key?: string;
-  currentStake?: string;
-  proposedStake?: string;
-  cumulativeStakeAmount?: string;
-  stakingStatus?: StakingStatus;
-  networkHolder?: boolean;
-  shards?: number[];
-  nodeInfo?: NodeInfo;
-}
+export interface BaseValidationNodeInfo {}
```
shelegdmitriy, line 41: Why do we need this? Can we delete it?

luixo: That's a glitch, removed it

frol:
Well, current stake should be present regardless of the fact if there is a staking pool contract or not, so I feel stake field should be outside of `stakingPoolInfo` object.

luixo:
@frol 
Fixed that.

frol:
@luixo @shelegdmitriy Oddly, both the production code and this PR provide different stake values and somehow neither is matching the CLI (thin wrapper over the JSON RPC `validators` API):

Production:

image

This PR:

image

CLI / JSON RPC:

image

FYI, CLI next epoch output:

image

luixo:
> Oddly, both the production code and this PR provide different stake values and somehow neither is matching the CLI (thin wrapper over the JSON RPC validators API)

I fixed the difference between production and staging (as one should expect).
The problem of different value to JSON RPC would be addressed in another PR where validators would be revisited.

frol:
@luixo PR still does not match the production values

luixo:
> @luixo PR still does not match the production values

Yes, but it matches production and the idea was to get rid of nulls, not to match RPC.

I'll dig deeper into validators in another PR.

frol (approved):
Let's get this fix merged to resolve the spinners on testnet and create a follow-up issue to address the discrepancy in the reported staked amount later.

PR #941, opened 2022-04-07
Feature: multiple networks

luixo:
As we have proper staging environments now, we need to test changes on different networks.
We don't have no automation on domain creation so far, so let's have this 'forced' network option at least.


shelegdmitriy (approved):
LGTM. Everything is working as before

frol:
@luixo it works oddly in my testing. Server-side is working fine, but client-side navigation is mostly broken, e.g. if you go to https://frontend-pr-941-5pl1.onrender.com/blocks?network=testnet, and try to open any block, you will get "not found", and you need to refresh the page (go through server-side rendering) to get the page loaded correctly. Nevertheless, it is enough for the most use-cases these days.

frontend/package.json
```diff
@@ -5,17 +5,11 @@
   "scripts": {
     "typecheck": "tsc --noEmit",
     "test": "npm run typecheck && env TZ=UTC jest",
-    "dev": "next",
-    "dev:local-wamp-testnet": "dotenv -c -e ../testnet.env -- next",
-    "dev:local-wamp-mainnet": "dotenv -c -e ../mainnet.env -- next",
-    "dev:production-wamp-testnet": "dotenv -c -e ../testnet.env -e ../production-wamp.env -- next",
-    "dev:production-wamp-mainnet": "dotenv -c -e ../mainnet.env -e ../production-wamp.env -- next",
+    "dev": "dotenv -c -e ../frontend.env -- next",
```
luixo, line 8: I made scripts section easier to comprehend now - with just `dev` and `dev:production-wamp`, I hope to make it even simpler with moving away from `wamp` someday :)

frontend/src/pages/_app.tsx
```diff
@@ -78,8 +79,28 @@ declare module "next/app" {
...
 }
 
+const wrapHandler = <T extends NextRouter["replace"]>(
+  router: NextRouter,
+  originalHandler: T
+): T => {
+  return ((href, as, ...args) => {
+    const network = router.query.network;
+    if (network) {
+      href += `?network=${network}`;
```
luixo, line 89: What do you think of this hacky solution guys?
Should I parse `href` and `as` and append network properly? Or it is good enough until we have more query params?

frontend/src/pages/_app.tsx
```diff
@@ -78,8 +79,28 @@ declare module "next/app" {
...
+      as += `?network=${network}`;
+    }
+    return originalHandler(href, as, ...args);
+  }) as T;
+};
+
 const App: AppType = React.memo(
   ({ Component, currentNearNetwork, language, pageProps }) => {
+    const router = useRouter();
```
luixo, line 98: Should I extract it to `useOverwriteRouterQueries` or smth?

luixo (commented):
Actually, I hope to merge server-side networks as well so one can jumpstart server & client locally with all network at the same time, there seem to be no technical limitations on this.

frontend/README.md
```diff
@@ -43,16 +43,14 @@ Project Root
 It is useful to have a live-updating dev-server during development, and most of the time you can use the production Explorer backend:
 
 ```
-$ npm run dev:production-wamp-testnet
```
frol, line 46: Hmm, and will it be testnet or mainnet? This raises more questions in my head than it solves.

frontend/src/pages/_app.tsx
luixo: Ok

frontend/README.md
luixo: Oh, that's a good point, I changed links in `NEAR_NETWORKS` config in `frontend.env` so the network-changing menu will actually work.

luixo: I believe it is actually more intuitive to change the network via interface.
I hope we'll be able to move backend to the unified system as well.

frol: > I believe it is actually more intuitive to change the network via interface.
> I hope we'll be able to move backend to the unified system as well.

From the scalability and security point of view we should not spend our time merging the backends together as it increases complexity and won't be used in production that way.

frontend/package.json
frol: Yep, thanks!

frol (approved):

frontend/README.md
luixo: It will get development process easier and help mimicry the production (where switching between the networks is one-click procedure).
It's not urgent at all, I'll mention you in the PR if I ever get to do that.

PR #960, opened 2022-04-25
Factoring out checks

luixo:
I believe it's a easier-to-read and easier-to-tune version of regular checks on backend

The biggest problem of this PR is that it's going to be hard to blame code because most of it just moved.


luixo:
Feel free to add comments after the merge.
e2e check fails because it uses production wamp and the scheme changed in this PR

PR #968, opened 2022-04-29
Refactor proposal: simpler variables

luixo:
I'm not sure that's the way, but I would like to hear your opinions


backend/src/db-utils.ts
```diff
@@ -205,7 +205,9 @@ const queryStakingPoolAccountIds = async (): Promise<string[]> => {
       [
         `SELECT account_id as "accountId"
     FROM accounts
-    WHERE account_id LIKE '%${nearStakingPoolAccountSuffix}'`,
+    WHERE account_id LIKE '%${
+      config.accountIdSuffix.stakingPool[config.networkName]
+    }'`,
```
shelegdmitriy, line 210: Will this query works normally in case of 'localhostnet'?

shelegdmitriy (commented):
There is much better to follow up the code, I like it!

backend/src/db-utils.ts
luixo: The line will be `WHERE account_id LIKE '%undefined'` and will return zero rows.
Looks good enough to me :)

backend/src/config.ts
```diff
@@ -1,134 +1,53 @@
...
-  variableName: string,
-  defaultValue: number
-): number =>
-  getEnvWithDefault(
-    variableName,
-    (v) => {
-      const num = parseInt(v || "");
-      return isNaN(num) ? undefined : num;
+/*
+To override a config parameter you should use a specific environment variable, following rules:
```
luixo, line 6: @frol 
How do you like new naming system for env variables?

render.yaml
```diff
@@ -121,16 +126,16 @@ services:
...
           type: web
           name: wamp
           property: host
-      - key: NEAR_EXPLORER_WAMP_BACKEND_SECRET
+      - key: NEAR_EXPLORER_CONFIG__WAMP__BACKEND_SECRET
         fromService:
           type: web
           name: wamp
-          envVarKey: NEAR_EXPLORER_WAMP_BACKEND_SECRET
+          envVarKey: NEAR_EXPLORER_CONFIG__BACKEND_SECRET
```
frol, line 138: 1. Don't forget to update the secrets env names on Render
2. FYI, You will need to manually remove the unused variable names as Render does not remove those automatically

frol (approved):
It looks much cleaner! 🎉

shelegdmitriy (approved):

backend/src/config.ts
```diff
@@ -1,134 +1,53 @@
...
+
+For example, to override `config.accountIdSuffix.stakingPool.localhostnet` you should use variable
+NEAR_EXPLORER_CONFIG__ACCOUNT_ID_SUFFIX__STAKING_POOL__LOCALHOSTNET
+*/
+export const config = merge(
+  {
+    archivalRpcUrl: "http://localhost:3030",
+    networkName: "localhostnet" as NetworkName,
+    accountIdSuffix: {
+      lockup: "lockup.near",
```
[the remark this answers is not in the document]
luixo: That's a great idea

render.yaml
luixo: Sure.
I'll make another PR to migrate DB env variables to this system as well, because it requires a little more effort

luixo:
It took quite a research to make both next build and jest work with common modules though it works now.
We'll probably migrate to swc + custom plugin for stitches friendly class names instead of babel later, but it'll take a PR in next.js team (I'll open an issue later).

PR #969, opened 2022-04-29
Feature: fungible tokens on account

luixo:
There is no design here, but you can check out the data flow on page:

https://frontend-pr-969-kgmz.onrender.com/beta/accounts/v2.ref-finance.near
https://frontend-pr-969-kgmz.onrender.com/beta/accounts/vlad.near/fungible-tokens/kongztoken.near

### DIFF
--- a/frontend/src/components/beta/accounts/AccountFungibleTokenHistory.tsx
+++ b/frontend/src/components/beta/accounts/AccountFungibleTokenHistory.tsx
@@ -1,0 +1,192 @@
+import * as React from "react";
+import JSBI from "jsbi";
+import { shortenString } from "../../../libraries/formatting";
+
+import { styled } from "../../../libraries/styles";
+import {
+  AccountFungibleToken,
+  AccountFungibleTokenHistoryElement,
+} from "../../../types/common";
+import { trpc } from "../../../libraries/trpc";
+import moment from "moment";
+import { useTranslation } from "react-i18next";
+import { TokenAmount } from "../../utils/TokenAmount";
+import LinkWrapper from "../../utils/Link";
+import { buildAccountUrl } from "../../../hooks/use-account-page-options";
+
+const TableWrapper = styled("div", {
+  display: "flex",
+  flexDirection: "column",
+  justifyContent: "center",
+  fontFamily: "Manrope",
+  paddingHorizontal: 24,
+  borderRadius: 8,
+  border: "1px solid #e8e8e8",
+  marginLeft: 20,
+  flex: 2,
+
+  "@media (max-width: 768px)": {
+    marginLeft: 0,
+    marginTop: 20,
+  },
+});
+
+const TableHeader = styled("thead", {
+  textTransform: "uppercase",
+  color: "#c4c4c4",
+  borderBottom: "1px solid #e8e8e8",
+
+  fontSize: 12,
+  fontWeight: 600,
+});
+
+const TableHeaderCell = styled("th", {
+  paddingVertical: 20,
+});
+
+const TableRow = styled("tr", {
+  fontSize: 14,
+  fontWeight: 500,
+  height: 50,
+
+  "& + &": {
+    borderTop: "1px solid #e8e8e8",
+  },
+});
+
+const TableElement = styled("td", {
+  verticalAlign: "top",
+  paddingVertical: 8,
+  paddingRight: 12,
+});
+
+const Link = styled(LinkWrapper, {
+  cursor: "pointer",
+
+  variants: {
+    disabled: {
+      true: {
+        cursor: "inherit",
+      },
+    },
+  },
+});
+
+type ItemProps = {
+  item: AccountFungibleTokenHistoryElement;
+  balance: string;
+  token: AccountFungibleToken;
+};
+
+const AccountFungibleTokenHistoryElementView: React.FC<ItemProps> = React.memo(
+  ({ item, token, balance }) => {
+    const { t } = useTranslation();
+    return (
+      <TableRow>
+        <Link
+          href={
+            item.counterparty
+              ? buildAccountUrl({
+                  accountId: item.counterparty,
+                  tab: "fungible-tokens",
+                })
+              : ""
+          }
+          shallow
+          disabled={!item.counterparty}
+        >
+          <TableElement>
+            {item.counterparty
+              ? shortenString(item.counterparty)
+              : item.direction === "in"
+              ? "MINT"
+              : "BURN"}
+          </TableElement>
+        </Link>
+        <TableElement>
+          <TokenAmount
+            token={{
+              ...token,
+              balance: (item.direction === "in" ? "" : "-") + item.amount,
+            }}
+            prefix={item.direction === "in" ? "+" : undefined}
+            noSymbol
+          />
+        </TableElement>
+        <TableElement>
+          <TokenAmount token={{ ...token, balance }} noSymbol />
+        </TableElement>
+        <Link
+          href={`/transactions/${item.transactionHash}#${item.receiptId}`}
+          shallow
+        >
+          <TableElement>{shortenString(item.receiptId)}</TableElement>
+        </Link>
+        <TableElement>
+          {moment(item.timestamp).format(t("common.date_time.date_format"))}
+        </TableElement>
+      </TableRow>
+    );
+  }
+);
+
+type Props = {
+  accountId: string;
+  token: AccountFungibleToken;
+};
+
+const AccountFungibleTokenHistory: React.FC<Props> = React.memo(
+  ({ accountId, token }) => {
+    const tokenHistoryQuery = trpc.useQuery([
+      "account-fungible-token-history",
+      { accountId, tokenAuthorAccountId: token.authorAccountId },
+    ]);
+    if (tokenHistoryQuery.status !== "success") {
+      if (tokenHistoryQuery.status === "loading") {
+        return <div>loading..</div>;
+      }
+      return null;
+    }
+    const { elements, baseAmount } = tokenHistoryQuery.data;
+    const cumulativeBalance = elements
+      .reduceRight<JSBI[]>(
+        (acc, element) => [
+          ...acc,
+          element.direction === "in"
+            ? JSBI.add(acc[acc.length - 1], JSBI.BigInt(element.amount))
+            : JSBI.subtract(acc[acc.length - 1], JSBI.BigInt(element.amount)),
+        ],
+        [JSBI.BigInt(baseAmount)]
+      )
+      .slice(1)
+      .reverse();
+    return (
+      <TableWrapper>
+        <table>
+          <TableHeader>
+            <tr>
+              <TableHeaderCell>Sender / Reciever</TableHeaderCell>
+              <TableHeaderCell>Amount</TableHeaderCell>
+              <TableHeaderCell>New Balance</TableHeaderCell>
+              <TableHeaderCell>Id</TableHeaderCell>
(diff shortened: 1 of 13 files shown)


backend/src/db-utils.ts
```diff
@@ -748,6 +748,44 @@ const queryAccountsList = async (
...
+        contractId: string;
+      },
+      {
+        accountId: string;
+        limit: number;
+        offset: number;
+      }
+    >(
+      [
+        `SELECT DISTINCT emitted_by_contract_account_id as "contractId"
```
shelegdmitriy, line 768:
```suggestion
        `SELECT DISTINCT emitted_by_contract_account_id AS "contractId"
```

shelegdmitriy (changes requested):
LGTM. But we need some design here. May be we can display it as a table (like on validators page,  but with three columns)
| Token | Author                                              | Balance                            |
| Utopia | utopia.secretskelliessociety.near | 373624.61642350 UTO |
as a temporary solution and not to procrastinate?

luixo:
> LGTM. But we need some design here. May be we can display it as a table (like on validators page, but with three columns)
| Token | Author | Balance |
| Utopia | utopia.secretskelliessociety.near | 373624.61642350 UTO |
as a temporary solution and not to procrastinate?

Do we need to rush actually?
We can add USN as a promo FT for example, but for a proper release we need proper design

luixo:
This one is ready for review.
It lacks some of the features (like "new balance") at the moment.

frol:
* [ ] NEAR native coin is missing. It should use a different source, but the same representation
* [ ] Sync with @reefoh on how we want the balances displayed (e.g. should we display at least 2 decimals after the floating point ["3.00" instead of just "3"], or should we add the symbol after the value like "3 USN" or "3.00 USN" or "3 $USN")
* [ ] Icons of some of the coins are missing (USN and KONGZ don't have the icon: https://frontend-pr-969-kgmz.onrender.com/beta/accounts/vlad.near/fungible-tokens/usn)
* [ ] In order to pull the new balance, make `ft_balance_of()` view call to the contract on the block height of the last asset event you pull, and then apply the deltas of the balance change to all the newer records

P.S. making view-calls through near-api-js is still convoluted, and does not let you specify the block height ([it has hardcoded `finality: "final"`](https://github.com/near/near-api-js/blob/3206032df37a3d7fadd7c9a8325711ec9a629293/src/account.ts#L540-L568)), so I recommend we further extend [our implementation in near-explorer](https://github.com/near/near-explorer/blob/9e032fb8e182db18e962c16db4c8168edd2241cc/backend/src/utils/near.ts#L30) with the following implementation that adds `blockReference` (I tried to be as realistic as possible, but I have not tested this code):

```ts
async viewFunction(
        contractId: string,
        methodName: string,
        args: any = {},
        { parse = parseJsonFromRawResponse, stringify = bytesJsonStringify, jsContract=false, blockReference = {finality: "final"} } = {}
    ): Promise<any> {
        let encodedArgs;
        this.validateArgs(args);
    
        if(jsContract){
            encodedArgs = this.encodeJSContractArgs(contractId, methodName, Object.keys(args).length >  0 ? JSON.stringify(args): '');
        } else{
            encodedArgs =  stringify(args);
        }

        const result = await this.connection.provider.query<CodeResult>({
            request_type: 'call_function',
            account_id: jsContract ? this.connection.jsvmAccountId : contractId,
            method_name: jsContract ? 'view_js_contract'  : methodName,
            args_base64: encodedArgs.toString('base64'),
            ...blockReference,
        });

        if (result.logs) {
            this.printLogs(contractId, result.logs);
        }

        return result.result && result.result.length > 0 && parse(Buffer.from(result.result));
    }
```

And then you will be able to pass `blockReference: { block_id: 68217757 }`

frontend/src/components/beta/accounts/AccountFungibleTokenHistory.tsx
```diff
@@ -0,0 +1,169 @@
...
+                  tab: "fungible-tokens",
+                })
+              : ""
+          }
+          shallow
+          disabled={!item.counterparty}
+        >
+          <TableElement>
+            {item.counterparty
+              ? shortenString(item.counterparty)
```
frol, line 98: Why don't we use AccountLink component? It makes perfect sense to be able to click on the account to navigate to that account details page.

frontend/src/components/beta/accounts/AccountFungibleTokenHistory.tsx
```diff
@@ -0,0 +1,169 @@
...
+              : item.direction === "in"
+              ? "MINT"
+              : "BURN"}
+          </TableElement>
+        </Link>
+        <TableElement>
+          <TokenAmount
+            token={{ ...token, balance: item.amount }}
+            sign={item.direction === "out" ? -1 : 1}
```
frol, line 107: This looks odd to me to use -1 / 1 as part of the component interface, to be honest. Why don't just use `balance` sign for that?

frontend/src/components/beta/accounts/AccountFungibleTokenHistory.tsx
```diff
@@ -0,0 +1,169 @@
...
+};
+
+const AccountFungibleTokenHistory: React.FC<Props> = React.memo(
+  ({ accountId, token }) => {
+    const tokenHistoryQuery = trpc.useQuery([
+      "account-fungible-token-history",
+      { accountId, tokenAuthorAccountId: token.authorAccountId },
+    ]);
+    if (tokenHistoryQuery.status !== "success") {
+      if (tokenHistoryQuery.status === "loading") {
```
frol, line 137: Somehow I felt like it is more idiomatic to use UPPER_SNAKE_CASE for the string consts.

frontend/src/components/utils/TokenAmount.tsx
```diff
@@ -0,0 +1,54 @@
...
+      overlay={
+        <Tooltip id="fungible-token">
+          {`${props.sign === -1 ? "-" : ""}${formattedAmount.quotient}${
+            remainderNoZeroes.length === 0 ? "" : `.${remainderNoZeroes}`
+          } ${props.token.symbol}`}
+        </Tooltip>
+      }
+    >
+      <span>
+        {props.sign === -1 ? "-" : ""}
```
frol, line 47: We should display `+` when it is not `-` if we display balance *change*

image

frontend/src/pages/beta/accounts/[...slug].tsx
```diff
@@ -82,18 +90,27 @@ const AccountQueryView: React.FC<QueryProps> = React.memo((props) => {
 
 export const getServerSideProps: GetServerSideProps<
   {},
-  { id: string }
+  { slug: string[] }
```
Comment (frol) [inline frontend/src/pages/beta/accounts/[...slug].tsx:93]: Do we have to give up on Next.js routing to avoid the account header re-render when user navigates between the tabs? It seems that recently added [Nested Layouts feature in Next.js](https://nextjs.org/blog/layouts-rfc) should help here

frol (changes requested):

shelegdmitriy:
When I cheked this [preview link](https://frontend-pr-969-kgmz.onrender.com/beta/accounts/v2.ref-finance.near) I saw some strange things:

1. The first thing you see is the list of FTs on the left, and a lot of blank space on the right:Снимок экрана 2022-06-22 в 22 41 08
Maybe it's only weird for me but maybe the first item of FTs list active by default?
2. When you choose the list item with a huge history you have to wait some time (quite a lot of time) (try to select *Utopia* from [preview link](https://frontend-pr-969-kgmz.onrender.com/beta/accounts/v2.ref-finance.near)). Also the list of history is too long to scroll it down. Maybe we can use pagination to avoid all it?

Overall it looks good, and all things I mention could solve later.

luixo:
> NEAR native coin is missing. It should use a different source, but the same representation

I guess we eventually decided to keep tokens without NEAR token for now, but we probably need to discuss how can we help with expectation of NEAR token history being there.

> Sync with @reefoh on how we want the balances displayed (e.g. should we display at least 2 decimals after the floating point ["3.00" instead of just "3"], or should we add the symbol after the value like "3 USN" or "3.00 USN" or "3 $USN")

I'll discuss the decimals. About the symbols - that's what we do now - display the symbols. We will not do "$USN" as a generic decision definitely - because `usn` contract has symbol `USN` in it, not `$USN`. We can apply some custom formatting for known FTs though.

> Icons of some of the coins are missing (USN and KONGZ don't have the icon: https://frontend-pr-969-kgmz.onrender.com/beta/accounts/vlad.near/fungible-tokens/usn)

I made a base64 validation function, but further research gave me an idea that react protects us from plain XSS attacks and browser protects us from CSRF attacks. Unfortunately, we'll have to add image sources manually in `next.config.js` (like for KONGZ logo)

> In order to pull the new balance, make ft_balance_of() view call to the contract on the block height of the last asset event you pull, and then apply the deltas of the balance change to all the newer records

That's a neat idea, actually!

frontend/src/components/beta/accounts/AccountFungibleTokenHistory.tsx
luixo: That's `react-query` status constant.

luixo:
> In order to pull the new balance, make ft_balance_of() view call to the contract on the block height of the last asset event you pull, and then apply the deltas of the balance change to all the newer records

Actually, as we don't have pagination yet, we can just use zero as initial balance.

frontend/src/components/beta/accounts/AccountFungibleTokenHistory.tsx
luixo: Well, actually, the thing is that I'll have to modify it quite a lot to use:
- we probably need the whole div around the name be clickable
- I'll have to work around disabled option for mint / burn events
- we need to derive if it's beta to have a proper link

Let's unify account link in another PR.

frontend/src/components/beta/accounts/AccountFungibleTokenHistory.tsx
luixo: I think we can add a check for negative numbers in `formatToPowerOfTen`, that will work.
We still need some specific prop to show "+" there, I chose a "prefix".

luixo:
> Maybe we can use pagination to avoid all it?

I made a hard limit 200 for now, we'll implement it in the next PR.

frol:
> It would be a little tough as we don't know what's the first FT for the user at page fetch and we keep path as our source of truth.

We could use /coins as our main URL, and then redirect to the first FT token when the list is known. Having NEAR native token on the list will, however simplify it, so we might as well just skip this auto-selection implementation now in favor of using /coins/NEAR in the future.

> I made a hard limit 200 for now, we'll implement it in the next PR.

👍 

> Actually, as we don't have pagination yet, we can just use zero as initial balance.

Given the above and also given the JSON Events support was not always there, we should not use 0 as our initial balance.

frol (changes requested):
Prefix in the tooltip is off:

image

New balance sometimes is odd:

Screen Shot 2022-06-25 at 13 02 06

Overall, I feel this PR is in a good shape to be merged and iterate on the design side in the next PRs.

luixo:
> We could use /coins as our main URL, and then redirect to the first FT token when the list is known. Having NEAR native token on the list will, however simplify it, so we might as well just skip this auto-selection implementation now in favor of using /coins/NEAR in the future.

I'm not sure we should use `coins` at all as not all fungible tokens are coins.

> Given the above and also given the JSON Events support was not always there, we should not use 0 as our initial balance.

Yep, found it the hard way :)

> Prefix in the tooltip is off:

Fixed!

> New balance sometimes is odd:

Fixed!

PR #972, opened 2022-04-30
Feature: deploy info

luixo:
A small view to understand what commit is currently deployed and what backend are we fetching data from.

### DIFF
--- a/frontend/src/pages/_app.tsx
+++ b/frontend/src/pages/_app.tsx
@@ -1,14 +1,18 @@
 import "../libraries/wdyr";
-import NextApp, { AppContext, AppInitialProps } from "next/app";
+import NextApp, { AppContext, ExtraAppInitialProps } from "next/app";
 import Head from "next/head";
 import { NextRouter, useRouter } from "next/router";
 import * as React from "react";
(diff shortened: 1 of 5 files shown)


shelegdmitriy (approved):
This is a good feature. Is this a community request? But I'm not sure we need to show this for all users. @luixo @frol what do you guys thinking?

luixo:
No, that was my request as I was guessing sometimes did environment update yet or did the domain moved yet.

frontend/src/pages/_app.tsx
```diff
@@ -105,8 +113,11 @@ const AppContextWrapper: React.FC<ContextProps> = React.memo((props) => {
   );
 });
 
+let extraAppInitialPropsCache: ExtraAppInitialProps;
```
luixo, line 116: This weird trick is needed because custom app unfortunately [doesn't have](https://github.com/vercel/next.js/discussions/10874) `getServerSideProps` yet, and we have to keep initial props somewhere when navigating between pages.

shelegdmitriy (approved):
LGTM

PR #974, opened 2022-05-06
Refactor: prepare DB for migration to query builder

luixo:


backend/src/db-utils.ts
```diff
@@ -748,12 +788,12 @@ const queryAccountsList = async (
   );
 };
 
-async function queryOutcomeTransactionsCountFromAnalytics(
+export const queryOutcomeTransactionsCountFromAnalytics = async (
```
frol, line 791: I wonder why do you prefer const + anonymous function over a named function. Is it just a code-style?

P.S. I am fine with it

frol (approved):

shelegdmitriy (approved):

backend/src/db-utils.ts
luixo: Yep, it's just a code style, there's no other point for doing that, but I do make things consistent here and there before I'll start full-blown linting.

PR #975, opened 2022-05-06
Kanel: DB types generation

luixo:
Woohoo! Our DB types are going to be precise starting very soon! :)


frol (approved):
Not bad! It seems you will love Rust 😄 

I am yet to see how it handles custom selected fields since rarely we actually pull in all the columns from a single table as is.

luixo:
It's not a query builder yet, and the query builder handles types pretty good except when I have to put custom raw sql in.
You'll in the next PR.

PR #976, opened 2022-05-06
Refactor: query builder

luixo:
That's a big one.
Please re-verify the queries for mistakes (I found a couple of mistakes like accidental change of ">" to "<", those cannot be tracked by types system).
There are a couple of problems left yet in those types, I'll get back to them later.


luixo:
@frol @shelegdmitriy Ready for review

shelegdmitriy:
@luixo Transactions page is not working (at least from preview link)

Снимок экрана 2022-05-11 в 21 00 50

shelegdmitriy:
Also there is a problem on Account Details page:
Снимок экрана 2022-05-11 в 22 02 29
Снимок экрана 2022-05-11 в 22 03 26

luixo:
@shelegdmitriy An eagle eye! :)
Just found out that postgresql doesn't complain on ambiguity of fields when one of them is from original `from` table and another one is from `as` expression in select.
Fixed all transaction queries that had this problem.

PR #977, opened 2022-05-09
Feature/beta transaction prototype

shelegdmitriy:
This is just a prototype of our new designs


frol:
Currently, the execution plan can be misleading

https://frontend-pr-977-43h6.onrender.com/beta/transactions/EztXEUwan7o5vYUZsqNtvdg7tymNwCnKMcTr97q2MiMt

It seems that each subsequent transfer is initiated by the previous transfer:

image

Here is the expected output:

image

shelegdmitriy:
I guess we can sunset this PR because I've split it into two PRs:
 - #1047 
 - #1048 
 
 @frol any objections?

frol:
I have no objections if we merge those PRs quickly as otherwise we loose the capability to have a preview link

PR #990, opened 2022-05-12
Refactor: no excessive account check

luixo:


frontend/src/pages/accounts/[id].tsx
```diff
@@ -113,27 +113,18 @@ export const getServerSideProps: GetServerSideProps<
...
-            account: (await fetcher("account-info", [accountId]))!,
-          },
-        };
-      } catch (accountFetchingError) {
-        return {
-          props: {
-            ...commonProps,
-            accountFetchingError,
-          },
-        };
```
Comment (frol) [inline frontend/src/pages/accounts/[id].tsx:131]: What is going to be displayed when users opens the Account Details page while the RPC is down? This handler was here to ensure that we don't misreport the account as non-existing when in fact we just had some backend problems.

Comment (luixo) [reply frontend/src/pages/accounts/[id].tsx:131]: Actually, it will show `page.accounts.error.account_fetching` translation which is `Information is not available at the moment. Please, check if the account name is correct or try later.` at the moment.
Seems like the same what it was before.

We have distinct view on `null` account and error in query.

PR #992, opened 2022-05-12
Refactor: account details

luixo:
A little bit of refactoring as I got confused a couple of times.
Also fixes bug of not showing deleted accounts.

### DIFF
--- a/backend/src/accounts.ts
+++ b/backend/src/accounts.ts
@@ -11,7 +11,7 @@ import {
   queryOutcomeTransactionsCountFromAnalytics,
   queryOutcomeTransactionsCountFromIndexerForLastDay,
 } from "./db-utils";
-import { callViewMethod, sendJsonRpc, sendJsonRpcQuery } from "./near";
+import { callViewMethod, sendJsonRpcQuery } from "./near";
 import { BIMax } from "./utils";
 
 export const isAccountIndexed = async (accountId: string): Promise<boolean> => {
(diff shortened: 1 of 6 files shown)


backend/src/accounts.ts
```diff
@@ -117,115 +117,40 @@ function ignoreIfDoesNotExist(error: unknown): null {
...
-
   return {
-    storageUsage: storageUsage.toString(),
-    stakedBalance: stakedBalance.toString(),
-    nonStakedBalance: nonStakedBalance.toString(),
-    minimumBalance: minimumBalance.toString(),
-    availableBalance: availableBalance.toString(),
-    totalBalance: totalBalance.toString(),
-    lockupAccountId: lockupDetails?.accountId,
-    lockupTotalBalance: lockupDetails?.totalBalance,
```
luixo, line 227: @frol
Why did we use to have this much of lockup data?
Also minimum balance, available balance and total balance were not used at all.

backend/src/accounts.ts
```diff
@@ -75,10 +75,10 @@ export const getAccountInfo = async (accountId: string) => {
   return {
     accountId: accountInfo.account_id,
     createdByTransactionHash:
-      accountInfo.created_by_transaction_hash || undefined,
+      accountInfo.created_by_transaction_hash || "Genesis",
```
frol, line 78: @luixo I just want to share the idea behind `undefined` vs "Genesis"; when the variable named with a clear type in mind ("Hash"), I do not expect to see some arbitrary value there (`Genesis`), which is why I preferred using `undefined`

backend/src/accounts.ts
```diff
@@ -75,10 +75,10 @@ export const getAccountInfo = async (accountId: string) => {
...
     createdAtBlockTimestamp: accountInfo.created_at_block_timestamp
       ? parseInt(accountInfo.created_at_block_timestamp)
-      : undefined,
+      : 0,
```
frol, line 81: On a similar note as above, `0` is now have a special meaning. Similarly to "Genesis", I usually prefer not to mix legal values with some special values.

backend/src/accounts.ts
frol: @luixo Historically, we tried to compute "total account balance", but we failed miserably on that side, but considered to get back to it with a similar approach as Wallet did for the profile page: https://wallet.near.org/profile/frol.near

backend/src/accounts.ts
luixo: Totally agree with that.
Addressed it here: https://github.com/near/near-explorer/pull/1007

PR #1007, opened 2022-05-20
Refactor: created/deleted on account

luixo:
Extracting out the values helps keeping types consistent.
Also, addressing this issue: https://github.com/near/near-explorer/pull/992#discussion_r872571488

### DIFF
--- a/frontend/src/components/accounts/AccountRow.tsx
+++ b/frontend/src/components/accounts/AccountRow.tsx
@@ -1,4 +1,3 @@
-import JSBI from "jsbi";
 import moment from "moment";
 
 import * as React from "react";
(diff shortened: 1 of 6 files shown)


frontend/src/components/accounts/AccountRow.tsx
```diff
@@ -97,7 +91,9 @@ const AccountRow: React.FC<Props> = React.memo(({ accountId }) => {
                   />
                   <TransactionRowTimer>
                     {t("component.accounts.AccountRow.created_on")}{" "}
-                    {moment(accountInfo.createdAtBlockTimestamp).format("LL")}
+                    {accountInfo.created
+                      ? moment(accountInfo.created.timestamp).format("LL")
+                      : "Genesis"}
```
luixo, line 96: The only thing that was added.
I don't think we need to translate that as there is no chance anybody will scroll down to this account :)

shelegdmitriy (approved):

PR #1009, opened 2022-05-23
Update state of transactions list

### LINKED ISSUE #1000: Transaction State Not Updating When Switching Account
When I go to an account, I can see the list of transactions but when I type in a different account in the search bar, the transactions don’t update until I refresh the page.  I've attached a quick demo:

https://user-images.githubusercontent.com/57506486/169072218-f1be5173-f701-473d-9124-3f9800f3f20c.mov

shelegdmitriy:
Resolves #1000


frontend/src/pages/transactions/index.tsx
```diff
@@ -27,7 +27,7 @@ const TransactionsPage: NextPage = React.memo(() => {
         <title>NEAR Explorer | Transactions</title>
       </Head>
       <Content title={<h1>{t("common.transactions.transactions")}</h1>}>
-        <Transactions fetch={fetch} />
+        <Transactions fetch={fetch} queryKey={"total-list"} />
```
shelegdmitriy, line 30: @luixo is it OK to pass `queryKey` like this?

frontend/src/components/accounts/Accounts.tsx
```diff
@@ -23,7 +23,7 @@ const Accounts: React.FC<Props> = React.memo(({ items }) => (
 
 const AccountsList = ListHandler<AccountListInfo, number>({
   Component: Accounts,
-  key: "Account",
+  key: ["Account", "total-list"],
```
shelegdmitriy, line 26: @luixo is it OK to pass queryKey like this?

frontend/src/components/transactions/Transactions.tsx
```diff
@@ -36,14 +37,16 @@ const Transactions: React.FC<InnerProps> = React.memo(({ items }) => {
 
 export type Props = {
   fetch: StaticConfig<TransactionBaseInfo, TransactionPagination>["fetch"];
+  queryKey: ReactQuery.QueryKey | null;
```
luixo, line 40: I think we can make it optional, not nullable:
```ts
queryKey?: ReactQuery.QueryKey
```
In this case we can keep `transactions` page without a change

frontend/src/components/transactions/Transactions.tsx
```diff
@@ -36,14 +37,16 @@ const Transactions: React.FC<InnerProps> = React.memo(({ items }) => {
...
 };
 
-const TransactionsList: React.FC<Props> = ({ fetch }) => {
+const TransactionsList: React.FC<Props> = ({ fetch, queryKey }) => {
+  const queryParams = !queryKey ? "Transaction" : ["Transaction", queryKey];
```
luixo, line 44: We may use null/undefined as a part of the key, actually:
```ts
...
key: ["transaction", queryKey],
...
```

luixo (changes requested):

frontend/src/components/utils/ListHandler.tsx
```diff
@@ -38,7 +38,7 @@ const LoadButton = styled("button", {
 });
 
 export type StaticConfig<T, I> = {
-  key: string;
+  key: ReactQuery.QueryKey | ReactQuery.QueryKey[];
```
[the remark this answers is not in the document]
shelegdmitriy: You’re right!

frontend/src/pages/blocks/[hash].tsx
```diff
@@ -71,7 +71,7 @@ const BlockDetail: NextPage<Props> = React.memo((props) => {
             icon={<TransactionIcon />}
             title={<h2>{t("common.transactions.transactions")}</h2>}
           >
-            <Transactions fetch={fetch} />
+            <Transactions fetch={fetch} queryKey={props.hash} />
```
Comment (luixo) [inline frontend/src/pages/blocks/[hash].tsx:74]: Let's have `["block", props.hash]` here

luixo (approved):

Issue #1001, opened 2022-05-19
Small styling issue with horizontal overflow


mikedotexe:
I saw this today and thought I'd flag a minor UI / styling issue.

![explorer-issue](https://user-images.githubusercontent.com/1042667/169364313-27de90f7-e9b8-4893-8ebd-7a0f00000c0c.gif)

❤️

luixo:
Thanks for pointing out!
Fix is on the way: https://github.com/near/near-explorer/pull/1016

PR #1016, opened 2022-05-26
Fix: code wrapper marksup

luixo:
Fixes https://github.com/near/near-explorer/issues/1001

### DIFF
--- a/frontend/src/components/transactions/ActionRowBlock.tsx
+++ b/frontend/src/components/transactions/ActionRowBlock.tsx
@@ -162,7 +162,10 @@ const ActionRowBlock: React.FC<Props> = React.memo(
           </Col>
           <ActionRowDetails type={viewMode}>
             <ActionRowMessage type={viewMode} noGutters>
-              <Col md="8" xs="7">
+              <Col
+                md={detalizationMode === "detailed" ? 8 : 12}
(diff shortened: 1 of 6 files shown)


frontend/src/components/transactions/ActionRowBlock.tsx
```diff
@@ -162,7 +162,7 @@ const ActionRowBlock: React.FC<Props> = React.memo(
           </Col>
           <ActionRowDetails type={viewMode}>
             <ActionRowMessage type={viewMode} noGutters>
-              <Col md="8" xs="7">
+              <Col md={detalizationMode === "detailed" ? 8 : 12} xs="7">
```
shelegdmitriy, line 165: This is how it looks like on staging:
Снимок экрана 2022-05-27 в 13 04 26

Maybe it's better to make it full width on mobile?
```suggestion
              <Col md={detalizationMode === "detailed" ? 8 : 12} xs="12">
```

shelegdmitriy (changes requested):

frontend/src/components/transactions/ActionRowBlock.tsx
luixo: It should depend on `detalizationMode` as well.

PR #1019, opened 2022-05-27
Refactor: infinite queries

luixo:
I decoupled data fetch layer from representation layer in infinite queries, also introduced `cursor` that we'll need in TRPC infinite queries.


backend/src/pubsub.ts
```diff
@@ -59,7 +59,13 @@ export const initPubSub = (
     for (const [name, endpoint] of handlers) {
       const uri = wrapProcedure(config.networkName, name as ProcedureType);
       try {
-        await session.register(uri, (args: any) => endpoint(args, state));
+        await session.register(uri, (args, kwargs) => {
```
luixo, line 62: A weird part of autobahn - we can only use arrays in `args` and object in `kwargs`.
As usage of `cursor` introduced objects in input - I decided to split input depending or being an array both on incoming and outcoming code sections.

PR #1020, opened 2022-05-27
Feature: useSubscription object

luixo:
In TRPC subscription result can be an error, so we probably need a way to show that.
Also, we can benefit from a proper `loading` state.

### DIFF
--- a/frontend/src/hooks/subscriptions.ts
+++ b/frontend/src/hooks/subscriptions.ts
@@ -1,22 +1,66 @@
 import * as React from "react";
+import * as ReactQuery from "react-query";
 
 import { SubscriptionTopicType, SubscriptionTopicTypes } from "../types/common";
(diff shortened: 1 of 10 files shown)


frontend/src/hooks/subscriptions.ts
```diff
@@ -1,22 +1,66 @@
 import * as React from "react";
+import * as ReactQuery from "react-query";
 
 import { SubscriptionTopicType, SubscriptionTopicTypes } from "../types/common";
 import { subscribe } from "../libraries/transport";
 import { useNetworkContext } from "./use-network-context";
 
+type UseSubscriptionResult<Topic extends SubscriptionTopicType> = Omit<
```
luixo, line 8: We may change the resulting type eventually, but this one seems like a good start

PR #1022, opened 2022-05-31
Feature: decouple block-chain-stats

luixo:
This PR is an example of how migration to TRPC can improve Explorer.

Before, we had to keep stats together and refresh them every second so a user will have his data in at maximum 1 second after JS kicks in.
Now, we can pick intervals based on how often actual stats updates and don't send redundant data to the user (most users will get only one update of `latestGasPrice` in the whole session).


backend/src/config.ts
```diff
@@ -29,7 +29,9 @@ export const config = merge(
 
     intervals: {
       checkFinalityStatus: SECOND,
-      checkChainBlockStats: SECOND,
+      checkLatestBlockHeight: SECOND,
+      checkLatestGasPrice: 15 * SECOND,
+      checkBlockProductionSpeed: 5 * SECOND,
```
luixo, line 34: @frol 
Please make a sanity check of these three intervals.
I guess gas price hardly ever updates.
Can we afford 5second lag in block production speed or should we be more frequent?

frol: These look reasonable to me

PR #1031, opened 2022-06-03
Fix error with frontend Dockerfile

mieubrisse:
Current ENTRYPOINT doesn't work, but this fixes it


luixo (approved):
I usually use `docker-compose.yml`, entrypoint works from there

PR #1032, opened 2022-06-03
Fix backend Dockerfile

mieubrisse:
Backend Dockerfile ENTRYPOINT doesn't work; this fixes


mieubrisse:
Hey @frol @luixo mind if we get this merged quick, so we can have an official commit in this repo to use as the tag for the Docker image that we'll build?

PR #1033, opened 2022-06-04
Fix: localhostnet -> localnet

luixo:
I researched through other NEAR projects and found out nobody uses `localhostnet` except us but `localnet`.
This is the rename.


frol:
`localhostnet` was there for WAMP local development where you would not need to name the topics as mainnet/testnet, instead just spawn the backend and frontend using a common name `localhostnet`. It is too convoluted to run Explorer for localnet, which is why it has never been an option. You may just drop it now since with the migration away from WAMP, we don't need it.

PR #1035, opened 2022-06-04
Refactor: genesis config subscription

luixo:
It seems like we can fetch genesis details just once and propagate them later.

### DIFF
--- a/backend/src/providers/network.ts
+++ b/backend/src/providers/network.ts
@@ -1,9 +1,10 @@
+import { Context } from "../context";
 import {
   NetworkStats,
   ValidationProgress,
   ValidatorEpochData,
 } from "../router/types";
-import { RPC, CurrentEpochState } from "../types";
+import { RPC } from "../types";
 import * as nearApi from "../utils/near";
 
 export const queryFinalBlock = async (): Promise<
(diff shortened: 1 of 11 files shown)


backend/src/providers/network.ts
```diff
@@ -79,56 +80,53 @@ const mapValidators = (
...
+    seatPrice: context.state.genesis
+      ? nearApi.validators
+          .findSeatPrice(
+            epochStatus.current_validators,
+            maxNumberOfSeats,
+            context.state.genesis.minStakeRatio,
+            networkProtocolConfig.protocol_version
+          )
+          .toString()
+      : "0",
```
luixo, line 110: This is the only downside of this PR.
In a quite unlikely case of a slow genesis block fetch we will have zeroed seatPrice for some time.

frontend/src/components/stats/ProtocolConfigInfo.tsx
```diff
@@ -142,7 +133,7 @@ const ProtocolConfigInfo: React.FC = React.memo(() => {
           title={t("component.stats.ProtocolConfigInfo.accounts_in_genesis")}
           cellOptions={{ xs: "12", sm: "6", md: "6", xl: "4" }}
         >
-          {genesisAccountsAmount}
+          {genesisConfigSub.data?.accountCount ?? "loading.."}
```
luixo, line 136: Ah no, that's another one :)

backend/src/providers/network.ts
luixo: Just checked - it is zero on staging.
Apparently not that unlikely 🤔

luixo: That actually happened because I didn't try to recalculate seat price after genesis eventually resolved so I ended up with 0 seat price until next epoch.

frontend/src/components/stats/ProtocolConfigInfo.tsx
luixo: I put a spinner here, now it all looks unified

frol (approved):

PR #1036, opened 2022-06-04
Refactor: tokens supply sub

luixo:
We can significantly reduce amount of queries to the analytics database.
The only quirk here I see is timezones.
I'm not sure at what moment operate all the actors and we should update such statistics.

### DIFF
--- a/backend/src/cron/tasks.ts
+++ b/backend/src/cron/tasks.ts
@@ -10,6 +10,7 @@ import {
   queryRecentBlockProductionSpeed,
   queryOnlineNodesCount,
   queryGenesisAccountCount,
+  queryTokensSupply,
 } from "../database/queries";
 import * as nearApi from "../utils/near";
 import {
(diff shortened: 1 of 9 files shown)


backend/src/cron/tasks.ts
```diff
@@ -109,12 +110,17 @@ export const transactionCountHistoryCheck: RegularCheckFn = {
   },
 };
 
+export const tokensSupplyCheck: RegularCheckFn = {
+  description: "circulatingSupplyCheck",
+  fn: publishOnChange("tokensSupply", queryTokensSupply, () => {
+    const endOfDayTimestamp = endOfDay(new Date()).valueOf();
+    return endOfDayTimestamp - Date.now();
```
luixo, line 117: @frol I guess "end of the day" is not the right heuristics for the update interval.
Currently database lags for 2 days so we don't stumble upon showing incomplete data for say today, but could we in the future?

frol: Can we just update it hourly? Circulating supply is getting computed for the day before when it is a few minutes in after midnight UTC.I believe it is totally fine if we lag by 1 hour.

frol (approved):
Overall, this PR looks good to me (see my only comment response above)

backend/src/cron/tasks.ts
luixo: I guess we can

PR #1037, opened 2022-06-04
Refactor: transactions history

luixo:
We actually have two pretty much same fetches for the same data (one is a little bit wider than another).
I guess we can afford join them to one.


frontend/src/components/dashboard/DashboardTransaction.tsx
```diff
@@ -27,13 +27,12 @@ const TransactionCharts = styled(Row, {
 
 const DashboardTransaction: React.FC = React.memo(() => {
   const { t } = useTranslation();
-  const { data: transactionsCountHistoryForTwoWeeks } = trpc.useQuery([
-    "transaction-history",
-  ]);
+  const transactionHistorySub = useSubscription(["transactionsHistory"]);
```
frol, line 30: Do I read it right that you pull in the the whole history of transactions volume since genesis (500 records) to only use 14 of them?

I don't feel it is a good idea given that this is a landing page

luixo: Eventually, I managed to add input to the subscriptions - and we can now create output based on it, keeping the same underlying cache.
Now we get only the needed 14 records for the transaction history.

PR #1047, opened 2022-06-15
Feature: beta transaction header

shelegdmitriy:
As we have the final version of beta TransactionDetails page I decide to split this [PR](https://github.com/near/near-explorer/pull/977) into several PRs to make code review simpler and faster.

This PR includes data fetching and TransactionHeader


backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,127 @@ export const getTransactionInfo = async (
...
+    proof: receipt.proof,
+    outcome: receipt.outcome,
+    includedInBlock: {
+      hash: receipt.block_hash,
+      height: includedInBlockMap.get(receipt.block_hash).blockHeight,
+      timestamp: includedInBlockMap.get(receipt.block_hash).blockTimestamp,
+    },
+  }));
+
+  if (
```
luixo, line 167: Lot of this code is copy-pasted from current transaction fetch implementation.
Can we refactor our the duplicated part?

backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,127 @@ export const getTransactionInfo = async (
...
+        actions:
+          "Action" in receiptItem.receipt
+            ? receiptItem.receipt.Action.actions.map(mapRpcActionToAction)
+            : [],
+      });
+      return acc;
+    },
+    new Map()
+  );
+  const refundReceiptsMap: Map<string, any> = new Map();
```
luixo, line 214: `any` is never a good idea

backend/src/providers/transactions.ts
```diff
@@ -263,3 +360,187 @@ export const collectNestedReceiptWithOutcome = (
...
+          return collectReceiptsWithOutcome(
+            outcomeReceiptHash,
+            receiptOutcome.id,
+            refundReceiptsMap,
+            receiptsByIdMap,
+            receiptOutcomesByIdMap
+          );
+        }
+      })
+      .filter((i) => i !== null) as any,
```
luixo, line 414: Why `any`?

luixo (changes requested):

frontend/src/translations/en/common.json
```diff
@@ -130,6 +130,16 @@
...
         "lockedUp": "Locked up"
       }
+    },
+    "transaction": {
+      "header": {
+        "title": "Transaction",
+        "fee": "Total Fee",
+        "amount": "Amount",
+        "when": "When",
+        "processed": "Processed in "
```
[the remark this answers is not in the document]
shelegdmitriy: Yes, but honestly, we don’t use this key here so I'll move to to #1048

frontend/src/translations/en/common.json
```diff
@@ -130,6 +130,16 @@
...
+      },
+      "dateFormat": "DD/MM/YYYY h:mm A"
```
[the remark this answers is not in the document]
shelegdmitriy: Not really. Here is our common date formats https://github.com/near/near-explorer/blob/master/frontend/src/translations/en/common.json#L75-L79. How about to add this date format to our common date formats?

backend/src/router/procedures.ts
```diff
@@ -189,6 +189,20 @@ export const router = trpc
       )[0] as KeysOfUnion<RPC.FinalExecutionStatus>;
     },
   })
+  .query("transaction", {
+    input: z.strictObject({
+      transactionHash: validators.transactionHash,
+    }),
+    resolve: async ({ input: { transactionHash } }) => {
+      const isTransactionIndexed = await transactions.getIsTransactionIndexed(
```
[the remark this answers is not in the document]
shelegdmitriy: You’re right. We already do such check inside `getTransactionDetails`

backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,127 @@ export const getTransactionInfo = async (
...
+  );
+  const transactionAmount = getDeposit(txActions);
+
+  return {
+    hash: transactionHash,
+    created: {
+      timestamp: transactionBaseInfo.blockTimestamp,
+      blockHash: transactionBaseInfo.blockHash,
+    },
+    transaction: transactionInfo.transaction,
```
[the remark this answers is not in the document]
shelegdmitriy: Actually we can expand those props from `transactionInfo.transaction` and place at the top as it only belongs to transaction. Here is what inside `transactionInfo.transaction`:
```
  signer_id: AccountId;
  public_key: PublicKey;
  nonce: Nonce;
  receiver_id: AccountId;
  actions: Vec<ActionView>;
  signature: Signature;
  hash: CryptoHash;
```
and we need `signer_id`, `receiver_id`, `public_key`, `nonce` only.

backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,127 @@ export const getTransactionInfo = async (
...
+    transactionIndex: transactionBaseInfo.transactionIndex,
+    transactionOutcome: transactionInfo.transaction_outcome,
+    transactionFee,
+    transactionAmount,
+    status: Object.keys(transactionInfo.status)[0] as any, // need to resolve
```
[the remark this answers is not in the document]
shelegdmitriy: You mean something like this https://github.com/near/near-explorer/blob/14806e862456582607c2f3dbd75972e628a6005a/backend/src/providers/transactions.ts#L212 ?

backend/src/providers/transactions.ts
shelegdmitriy: I think it's better to do in #1048

frontend/src/translations/en/common.json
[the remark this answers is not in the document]
luixo: That's a good idea.

backend/src/providers/transactions.ts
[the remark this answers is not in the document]
luixo: Yep

frontend/src/components/beta/transactions/TransactionHeader.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+const Amount = styled("div", {
+  fontWeight: 500,
+  fontSize: 20,
+  color: "#fff",
+  marginTop: 16,
+  fontFamily: "SF Mono",
+});
+
+const TimeWrapper = styled(AuthorInfo);
```
luixo, line 103: Do we need a wrapper in this case?
Should we name it `CenteredContainer` in this case?

frontend/src/components/beta/transactions/TransactionHeader.tsx
```diff
@@ -0,0 +1,164 @@
...
+  height: "100%",
+  width: 1,
+  marginHorizontal: 20,
+});
+
+const AmountHeader = styled("div", {
+  fontSize: 12,
+  color: "#c9c9c9",
+  lineHeight: "14px",
+  fontFamily: "SF Pro Display",
```
luixo, line 92: Do we have all fonts loaded on the page?
We need a design system so desperately!

frontend/src/components/beta/transactions/TransactionHeader.tsx
```diff
@@ -0,0 +1,164 @@
...
+});
+
+const TransactionHash = styled("h1", {
+  fontFamily: "SF Pro Display",
+  fontSize: 18,
+  fontWeight: 400,
+  color: "#fff",
+  lineHeight: "21px",
+  marginRight: 14,
+  marginLeft: 34,
```
luixo, line 64: I believe this number is tied up to avatar size.
Should we do something like this?
```ts
const AVATAR_SIZE = 26; // any number

const Avatar = styled("div", {
  // ...
  size: AVATAR_SIZE
});

const TransactionHash = stlyed("h1", {
  paddingHorizontal: 14,
  marginLeft: AVATAR_SIZE
})
```

frontend/src/components/beta/common/UtcLabel.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from "react";
+import { styled } from "../../../libraries/styles";
+
+const UTC = styled("div", {
+  fontFamily: "SF Mono",
+  fontWeight: 600,
+  fontSize: 12,
+  lineHeight: "20px",
+  color: "#fff",
+  paddingHorizontal: 7,
```
luixo, line 10: I would discuss those odd numbers with @reefoh, I believe they should be dividable by 2 or even by 4.

backend/src/router/procedures.ts
```diff
@@ -189,6 +189,14 @@ export const router = trpc
       )[0] as KeysOfUnion<RPC.FinalExecutionStatus>;
     },
   })
+  .query("transaction", {
+    input: z.strictObject({
+      transactionHash: validators.transactionHash,
```
luixo, line 194: Could it be just `hash`?

backend/src/router/procedures.ts
```diff
@@ -189,6 +189,14 @@ export const router = trpc
...
+    }),
+    resolve: async ({ input: { transactionHash } }) => {
```
luixo, line 196: No need for `async / await` keywords, but it's up to you to keep them

backend/src/providers/transactions.ts
```diff
@@ -263,3 +342,138 @@ export const collectNestedReceiptWithOutcome = (
...
+  receiver_id: string;
+};
+
+export type TransactionDetails = {
+  hash: string;
+  created: {
+    timestamp: number;
+    blockHash: string;
+  };
+  signerId: RPC.FinalExecutionOutcomeView["transaction"]["signer_id"];
```
luixo, line 439: I believe `signerId` and `receiverId` are just strings

backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,109 @@ export const getTransactionInfo = async (
...
+    amount: transactionAmount,
+    status: convertRpcTxStatusToStatus(transactionInfo.status),
+    gasUsed,
+    gasAttached,
+  };
+};
+
+const convertRpcTxStatusToStatus = (
+  status: RPC.FinalExecutionStatus
+): TransactionStatus => {
```
luixo, line 220: Let's make `TransactionStatus` start with lowercase as it's keeping props `camelCase`

backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,109 @@ export const getTransactionInfo = async (
...
+  const txStatus = Object.keys(status)[0];
+  switch (txStatus) {
```
luixo, line 222: So what's the state machine of those statuses?
Not Started -> Started -> Fetching -> Either(Failure, Success)?

backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,109 @@ export const getTransactionInfo = async (
...
+    case "NotStarted":
+      return "NotStarted";
+    case "Started":
+      return "Started";
+    case "Failure":
+      return "Failure";
+    case "SuccessValue":
+      return "Succeeded";
```
luixo, line 230: I believe the counterpart of `Failure` would be `Success`

backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,109 @@ export const getTransactionInfo = async (
...
+  };
+};
+
+const convertRpcTxStatusToStatus = (
+  status: RPC.FinalExecutionStatus
+): TransactionStatus => {
+  const txStatus = Object.keys(status)[0];
+  switch (txStatus) {
+    case "NotStarted":
+      return "NotStarted";
```
luixo, line 224: What does `not started` mean in RPC response?

backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,127 @@ export const getTransactionInfo = async (
...
+  const txActions =
+    transactionInfo.transaction.actions.map(mapRpcActionToAction);
+  const gasAttached = getGasAttached(gasUsed, txActions);
+  const transactionFee = getTransactionFee(
+    transactionInfo.transaction_outcome,
+    receiptsOutcome
+  );
+  const transactionAmount = getDeposit(txActions);
+
+  return {
```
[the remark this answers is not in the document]
luixo: @shelegdmitriy 
We still have `gasUsed`, `gasAttached`, `index` not used.

backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,109 @@ export const getTransactionInfo = async (
...
+      height: includedInBlockMap.get(receipt.block_hash).blockHeight,
+      timestamp: includedInBlockMap.get(receipt.block_hash).blockTimestamp,
+    },
+  }));
+
+  if (
+    transactionInfo.receipts.length === 0 ||
+    transactionInfo.receipts[0].receipt_id !== receiptsOutcome[0].id
+  ) {
+    transactionInfo.receipts.unshift({
```
luixo, line 171: We don't use this code in this PR

backend/src/providers/transactions.ts
```diff
@@ -126,6 +130,109 @@ export const getTransactionInfo = async (
   return transaction[0] || null;
 };
 
+export const getTransactionDetails = async (
+  transactionHash: string
+): Promise<TransactionDetails | null> => {
+  const transactionBaseInfo = await getTransactionInfo(transactionHash);
+
+  if (!transactionBaseInfo) {
+    throw new Error(`No hash ${transactionHash} found`);
```
luixo, line 139: This probably should return null now

luixo (changes requested):

frontend/src/components/beta/transactions/TransactionHeader.tsx
shelegdmitriy: No. I'll left only one font until we receive the whole list of fonts

backend/src/providers/transactions.ts
shelegdmitriy: According to this [doc](https://docs.rs/near-primitives/0.12.0/near_primitives/views/enum.FinalExecutionStatus.html) 
> NotStarted
The execution has not yet started.
which means that transaction is created but it's execution has not started yet

@frol is it true?

backend/src/providers/transactions.ts
shelegdmitriy: Yes. Not Started -> Started -> Failure || Success

frol: You will never receive "Not Started" or "Started" from RPC, so it should be just "Failure" or "Success" for now. "Not Started" and "Started" statuses were added there due to my lack of understanding of this fact back in the days. We may add support for incomplete transactions in the future, but it is not that crucial given how fast NEAR network is.

backend/src/providers/transactions.ts
frol: See my reply above

backend/src/providers/transactions.ts
```diff
@@ -10,6 +10,8 @@ import {
 import { z } from "zod";
 import { validators } from "../router/validators";
 
+import { sendJsonRpc } from "../utils/near";
```
luixo, line 13: We use `* as nearApi` everywhere.
That's not a strict rule so far, but it would be nice to have things consistent

backend/src/providers/transactions.ts
luixo: So it should be something like this now?
```
switch (txStatus) {
  case "Failure":
    return "fail";
  case "SuccessValue":
    return "success";
  default:
    return "fetching";
}
```

backend/src/providers/transactions.ts
```diff
@@ -193,17 +257,6 @@ async function getTransactionsActionsList(
   return transactionsActionsByHash;
 }
 
-export type Action<A extends RPC.ActionView = RPC.ActionView> =
```
luixo, line 196: Can we keep types in place if we're not changing them?

backend/src/providers/transactions.ts
```diff
@@ -263,3 +299,86 @@ export const collectNestedReceiptWithOutcome = (
     },
   };
 };
+
+export const getDeposit = (actions: Action[]) =>
+  actions
+    .map((action: Action) => {
```
luixo, line 305: Excessive type

backend/src/providers/transactions.ts
```diff
@@ -263,3 +299,86 @@ export const collectNestedReceiptWithOutcome = (
...
+      }
+    })
+    .reduce((accumulator, deposit) => accumulator + deposit, 0n)
+    .toString();
+
+export const getTransactionFee = (
+  transactionOutcome: TransactionOutcome,
+  receiptsOutcome: ReceiptsOutcome[]
+) => {
+  const tokensBurntByTx = transactionOutcome
```
luixo, line 319: How come `transactionOutcome` can be falsy?

backend/src/providers/transactions.ts
```diff
@@ -263,3 +299,86 @@ export const collectNestedReceiptWithOutcome = (
...
+    ? BigInt(transactionOutcome.outcome.tokens_burnt)
+    : 0n;
+  const tokensBurntByReceipts = receiptsOutcome
```
luixo, line 322: The same question as above

backend/src/providers/transactions.ts
```diff
@@ -263,3 +299,86 @@ export const collectNestedReceiptWithOutcome = (
...
+  actions
+    .map((action: Action) => {
+      if ("deposit" in action.args) {
+        return BigInt(action.args.deposit);
+      } else {
+        return 0n;
+      }
+    })
+    .reduce((accumulator, deposit) => accumulator + deposit, 0n)
+    .toString();
```
luixo, line 313: I believe we should yield `BigInt`s here and convert to string the last mile - before sending values to the client.

backend/src/providers/transactions.ts
```diff
@@ -126,6 +128,68 @@ export const getTransactionInfo = async (
...
+
+  if (!transactionBaseInfo) {
+    return null;
+  }
+  const transactionInfo = await sendJsonRpc("EXPERIMENTAL_tx_status", [
+    transactionBaseInfo.hash,
+    transactionBaseInfo.signerId,
+  ]);
+
+  const receiptsOutcome = transactionInfo.receipts_outcome.map((receipt) => ({
```
luixo, line 144: We actually need only `tokensBurnt` from outcome here, I don't think we need handle around `proof` or `id` or `includedInBlock`.

luixo (approved):

Issue #1025, opened 2022-06-01
Implement a subscribe adding helper


luixo:
We currenly have to write a pretty lengthy construction to add a topic user can subscribe to:
```ts
const router = trpc.router
  .query("foo", {
    resolve: getQueryResolve("foo"),
  })
  .subscription("foo", {
    resolve: getSubscriptionResolve("foo"),
  })
```

I hope we can have a helper to rewrite it in a simpler way:
```ts
const router = ['foo'].reduce((acc, topic) => addSubscription(acc, topic), trpc.router());
```

It requires some thorough work with generics and TRPC types.
I leave this issue as a reminder to myself to implement this.

PR #1050, opened 2022-06-16
Feature: generic subscriptions helper

luixo:
Fixes: https://github.com/near/near-explorer/issues/1025


backend/src/router/subscriptions.ts
```diff
@@ -1,128 +1,87 @@
...
+              ctx.subscriptionsEventEmitter.off(topic, onData);
+            });
+          }
+          const cachedData = ctx.subscriptionsCache[topic] as Parameters<
+            SubscriptionEventMap[typeof topic]
+          >[0] &
+            Parameters<NonNullable<typeof filterFn>>[0];
+          return cachedData;
+        },
+      }) as any;
```
luixo, line 73: This is ugly, but I don't really see options which include proper typings.

PR #1051, opened 2022-06-17
Feature: Non-Fungible Tokens on account page

shelegdmitriy:
This is the very first iteration of Non-Fungible Tokens on account page and it's still in progress.

- https://frontend-pr-1051-u41u.onrender.com/beta/accounts/vlad.near/collectibles
- https://frontend-pr-1051-u41u.onrender.com/beta/accounts/dmitriy_sheleg.near/collectibles
- https://frontend-pr-1051-u41u.onrender.com/beta/accounts/x.paras.near/collectibles

image


frontend/next.config.ts
```diff
@@ -57,6 +57,19 @@ const nextConfig: ExplorerConfig & NextConfig = {
...
+      "cloudflare-ipfs.com",
+      "ipfs.fleek.co",
+      "nearnaut.mypinata.cloud",
+      "arweave.net",
+      "beerpunks.mypinata.cloud",
+      "bafybeidw4zstz4q3gezdbtslzvxkhkjko32octdgg2xbtsghjhsj7xkd7q.ipfs.dweb.link",
+      "nftstorage.link",
+      "ipfs.io",
+      "ewtd.mypinata.cloud",
+    ],
```
shelegdmitriy, line 71: This list of domains could be Infinite, so I think we need to drop  next/image `<Image />` component

backend/src/router/account/fungible-tokens.ts
```diff
@@ -20,7 +20,9 @@ type FungibleTokenMetadata = {
 };
 
 const base64ImageRegex = /^data:image/;
-const validateBase64Image = (base64Image: string | null): string | null => {
+export const validateBase64Image = (
+  base64Image: string | null
+): string | null => {
```
shelegdmitriy, line 25: I'd prefer to move this helper to utils, @luixo WDYT?

frontend/src/components/beta/common/NFTMedia.tsx
```diff
@@ -0,0 +1,62 @@
...
+      {isVideo ? (
+        <video muted={true} loop controls autoPlay={false} poster={poster}>
+          <source
+            src={mediaUrl}
+            type={`video/${mimeType}`}
+            onError={handlePoster}
+          />
+        </video>
+      ) : (
+        <Img
```
shelegdmitriy, line 50: I'd like to use nextjs/image `<Image />` component, but using that component requires a list of domains which could be infinite. And we can't set it as `dynamic` or something like that.

frontend/next.config.ts
frol: @shelegdmitriy Most of these domains are just gateways to IPFS, so it is expected that they take some time to load (we need to keep retrying on error and communicate that to the users - this might be done in the future PRs)

frontend/src/components/beta/common/NFTMedia.tsx
frol: There is no reason to make `` dynamic as there is no heavy rendering behind getting `` rendered to HTML output on the server side. Why do you want to use `<Image>` over ``? I feel `<Image>` might cache the image on build which is not the behavior we need here as NFT images are not known ahead of time and we don't want to pull them in to be part of the project (they are external resources).

shelegdmitriy: I wanted to use it because nextjs/image could compress images on the fly, that's why the images could load much faster

backend/src/router/account/fungible-tokens.ts
luixo: Sure thing

backend/src/router/account/non-fungible-tokens.ts
```diff
@@ -0,0 +1,170 @@
...
+
+import { Context } from "../../context";
+import { indexerDatabase } from "../../database/databases";
+import { div } from "../../database/utils";
+import * as nearApi from "../../utils/near";
+import { validators } from "../validators";
+import { validateBase64Image } from "./fungible-tokens";
+
+type Token = {
+  token_id: string;
```
luixo, line 12: Let's refer source for types we got from outside (if we got one)

backend/src/router/account/non-fungible-tokens.ts
```diff
@@ -0,0 +1,170 @@
...
+  updated_at: number | null; // When token was last updated, Unix epoch in milliseconds
+  extra: string | null; // anything extra the NFT wants to store on-chain. Can be stringified JSON.
+  reference: string | null; // URL to an off-chain JSON file with more info.
+  reference_hash: string | null; // Base64-encoded sha256 hash of JSON from reference field. Required if `reference` is included.
+};
+
+// copied from Wallet https://github.com/near/near-wallet/blob/master/packages/frontend/src/services/NonFungibleTokens.js#L95
+const buildMediaUrl = (
+  media: TokenMetadata["media"],
+  base_uri: NFTContractMetadata["base_uri"]
```
luixo, line 45: I believe we can use camelCase here

backend/src/router/account/non-fungible-tokens.ts
```diff
@@ -0,0 +1,170 @@
...
+    input: z.strictObject({
+      contractId: z.string(),
+      accountId: validators.accountId,
+      limit: validators.limit,
+      cursor: z.number().optional(),
+    }),
+    resolve: async ({
+      input: { contractId, accountId, limit, cursor = 0 },
+    }) => {
+      const nonFungibleTokenCount: string = await nearApi.callViewMethod(
```
luixo, line 83: Let's put `string` as generic to `callViewMethod` call

backend/src/router/account/non-fungible-tokens.ts
```diff
@@ -0,0 +1,170 @@
...
+        contractId,
+        "nft_supply_for_owner",
+        {
+          account_id: accountId,
+        }
+      );
+      const nonFungibleTokenContractMetadata: NFTContractMetadata =
+        await nearApi.callViewMethod(contractId, "nft_metadata", {});
+
+      const nonFungibleTokenMetadata: Token[] = await nearApi.callViewMethod(
```
luixo, line 93: Those three calls can be squashed into one `Promise.all` to make them parallel

backend/src/router/account/non-fungible-tokens.ts
```diff
@@ -0,0 +1,170 @@
...
+        contractId,
+        "nft_tokens_for_owner",
+        {
+          account_id: accountId,
+          from_index: cursor.toString(),
+          limit,
+        }
+      );
+
+      return nonFungibleTokenMetadata.map((token, index) => ({
```
luixo, line 103: We can now return `items` + `cursor` as an object (no need to extract `cursor` from last item), see transactions or beta account activity.

backend/src/router/account/non-fungible-tokens.ts
```diff
@@ -0,0 +1,170 @@
...
+          icon: validateBase64Image(nonFungibleTokenContractMetadata.icon),
+        },
+        index: cursor + index,
+        totalNftCount: parseInt(nonFungibleTokenCount ?? 0),
+      }));
+    },
+  })
+  .query("nonFungibleTokenHistory", {
+    input: z.strictObject({
+      tokenAuthorAccountId: z.string(),
```
luixo, line 129: `validators.accountId`?

frontend/src/components/beta/accounts/AccountNonFungibleTokens.tsx
```diff
@@ -0,0 +1,265 @@
...
+  modalOpen: boolean;
+};
+
+const AccountNonFungibleTokenView: React.FC<ItemProps> = React.memo(
+  ({ token, onClick, modalOpen }) => {
+    return (
+      <>
+        <Token
+          onClick={onClick}
+          css={modalOpen ? { borderColor: "#1e93ff" } : {}}
```
luixo, line 125: You can use variants here

frontend/src/components/beta/accounts/AccountNonFungibleTokens.tsx
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    return (
+      <Wrapper>
+        <ContractsWrapper>
+          {contracts.map((contract) => (
+            <ContractItem
+              key={contract}
+              onClick={setActiveContract(contract)}
+              css={
```
luixo, line 246: Variants here as well

frontend/src/components/beta/common/NFTMedia.tsx
```diff
@@ -0,0 +1,62 @@
...
+          alt="NFT"
+          src={mediaUrl}
+          loader={<PaginationSpinner />}
+          onError={handleMediaUrl}
+          style={{ width: "100%" }}
```
luixo, line 55: Looks like a job for `styled` :)

frontend/src/components/beta/common/NFTMedia.tsx
luixo: There is a reason to have a list of allowed domains - to prevent attacks from malicious users.
I guess you can use proxy servers for that: https://nextjs.org/docs/api-reference/next/image#loader-configuration

luixo (changes requested):

backend/src/router/account/non-fungible-tokens.ts
```diff
@@ -0,0 +1,198 @@
...
+    resolve: async ({ input: { accountId } }) => {
+      const selection = await indexerDatabase
+        .selectFrom("assets__non_fungible_token_events")
+        .select("emitted_by_contract_account_id as contractId")
+        .distinctOn("emitted_by_contract_account_id")
+        .where("token_new_owner_account_id", "=", accountId)
+        .execute();
+      const nftsPerContractCount = await Promise.all(
+        selection.map(async ({ contractId }) => {
+          const count = nearApi.callViewMethod<string>(
```
luixo, line 138: Redundant `async` keyword & `count` variable

luixo (approved):

PR #1058, opened 2022-06-29
Refactor: gas used subscription

luixo:
Carefully moving some stats to a more readable solution - aka subscriptions.

### DIFF
--- a/backend/src/cron/tasks.ts
+++ b/backend/src/cron/tasks.ts
@@ -12,6 +12,7 @@ import {
   queryOnlineNodesCount,
   queryGenesisAccountCount,
   queryTokensSupply,
+  queryGasUsedAggregatedByDate,
   healthCheck,
 } from "../database/queries";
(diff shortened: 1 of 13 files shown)


backend/src/cron/tasks.ts
```diff
@@ -120,12 +120,18 @@ export const tokensSupplyCheck: RegularCheckFn = {
...
 
+export const gasUsedHistoryCheck: RegularCheckFn = {
+  description: "gasUsedHistoryCheck",
+  fn: publishOnChange(
+    "gasUsedHistory",
+    queryGasUsedAggregatedByDate,
+    config.intervals.checkAggregatedStats
+  ),
+};
+
```
shelegdmitriy, line 131: Before this we have a `retriable` method wrapper for such queries because they could fail for some reason. How this case will be handled now?

shelegdmitriy (approved):

PR #1060, opened 2022-06-29
Refactor: account related subscriptions

luixo:
Moving account stats to subscriptions


backend/src/router/subscriptions.ts
```diff
@@ -71,6 +77,7 @@ const withTopics = <InitialRouter extends AnyRouter<Context>>(
             Parameters<SubscriptionEventMap[typeof topic]>[0]
           >((emit) => {
             const onData = (data: TopicDataType) => {
+              // @ts-ignore
```
luixo, line 80: This is pretty bad, I have to solve it.

luixo: I kinda solved it, but still have `any` there.

PR #1064, opened 2022-06-29
Refactor: block router

luixo:


backend/src/router/block.ts
```diff
@@ -3,46 +3,183 @@ import { z } from "zod";
...
+import { indexerDatabase } from "../database/databases";
+import { count, div, sum } from "../database/utils";
+import { millisecondsToNanoseconds } from "../utils/bigint";
 
 export const router = trpc
   .router<Context>()
-  .query("nearcore-final-block", {
-    resolve: () => {
-      return nearApi.sendJsonRpc("block", { finality: "final" });
+  .query("final", {
```
luixo, line 13: That looks for me like a more sensible way - as we use `nearcore-final-block` for just getting the block's height - to use it for both calls and resolve with just actual height.

backend/src/router/block.ts
```diff
@@ -3,46 +3,183 @@ import { z } from "zod";
...
+        hash: selection.hash!,
+        // TODO: Discover how to get rid of non-null type assertion
+        height: parseInt(selection.height!),
+        timestamp: parseInt(selection.timestamp),
+        // TODO: Discover how to get rid of non-null type assertion
+        prevHash: selection.prevHash!,
+        // TODO: Discover how to get rid of non-null type assertion
+        gasPrice: selection.gasPrice!,
+        // TODO: Discover how to get rid of non-null type assertion
+        // TODO: Remove totalSupply whenever `epochStartBlock` will be served
```
luixo, line 103: Thats a point I should not forget

frontend/src/components/utils/Search.tsx
```diff
@@ -186,19 +186,9 @@ const Search: React.FC<Props> = React.memo(({ dashboard }) => {
 
       const cleanedSearchValue = value.replace(/\s/g, "");
 
-      let blockPromise: Promise<string | null>;
-      const maybeBlockHeight = cleanedSearchValue.replace(/[,]/g, "");
```
luixo, line 190: No need for complicated logic on client - just check if it's a number server-side and return hash.

frontend/src/pages/api/metrics/indexer.ts
```diff
@@ -7,9 +7,11 @@ const handler: NextApiHandler = async (req, res) => {
   try {
     const networkName = getNearNetworkName(req.query, req.headers.host);
     const trpcClient = getTrpcClient(networkName);
-    const rpcFinalBlock = await trpcClient.query("nearcore-final-block");
-    const indexerFinalBlock = await trpcClient.query("blocks-list", {
-      limit: 1,
+    const rpcFinalBlock = await trpcClient.query("block.final", {
```
luixo, line 10: Isn't that more comprehendible?

shelegdmitriy: That's much better!

PR #1082, opened 2022-07-24
Fix search by block hash

### LINKED ISSUE #1081: Search by block hash does not find the block
Take any block hash, e.g. `4usxPdL68XbQXBeZPGj56wH1FAqrXhuFygHTz27xodMU`, and try to search it on Explorer. Even though I can perfectly open the page https://explorer.near.org/blocks/4usxPdL68XbQXBeZPGj56wH1FAqrXhuFygHTz27xodMU, the search still says that there is nothing found.

@shelegdmitriy Could you take a quick look at it?

shelegdmitriy:
Resolves #1081

### DIFF
--- a/backend/src/router/block/get-hash-by-id.ts
+++ b/backend/src/router/block/get-hash-by-id.ts
@@ -13,10 +13,12 @@ export const router = trpc.router<Context>().query("getHashById", {
     let selection = indexerDatabase
       .selectFrom("blocks")
       .select("block_hash as hash");
-    if (isNaN(parseInt(blockId))) {
+    if ([43, 44].indexOf(blockId.length) >= 0) {
       selection = selection.where("block_hash", "=", blockId);
-    } else {
+    } else if (!isNaN(Number(blockId))) {
       selection = selection.where("block_height", "=", blockId);
+    } else {
+      return null;
     }
     const block = await selection.limit(1).executeTakeFirst();
     if (block) {


backend/src/router/block/get-hash-by-id.ts
```diff
@@ -13,7 +13,7 @@ export const router = trpc.router<Context>().query("getHashById", {
     let selection = indexerDatabase
       .selectFrom("blocks")
       .select("block_hash as hash");
-    if (isNaN(parseInt(blockId))) {
```
shelegdmitriy, line 16: If we take blockHash from issue (`4usxPdL68XbQXBeZPGj56wH1FAqrXhuFygHTz27xodMU`) and `console.log("4usxPdL68XbQXBeZPGj56wH1FAqrXhuFygHTz27xodMU")` we'll see output `4` and `isNaN(parseInt("4usxPdL68XbQXBeZPGj56wH1FAqrXhuFygHTz27xodMU"))` returns `false`. If we use `Number` instead of `parseInt` - works as expected.

frol: This still looks incorrect to me. It is possible that block hash will only contain digits, so it will appear as a very big number. Let’s change this condition to just checking the length (if it is 43-44 characters long, search it as block hash, otherwise, check if it is a number and search as if it is block height, and otherwise skip search completely, as it does not make any sense to search lets say frol.near block)

backend/src/router/block/get-hash-by-id.ts
```diff
@@ -13,10 +13,12 @@ export const router = trpc.router<Context>().query("getHashById", {
...
+    if ([43, 44].indexOf(blockId.length) >= 0) {
       selection = selection.where("block_hash", "=", blockId);
-    } else {
+    } else if (!isNaN(parseInt(blockId))) {
```
luixo, line 18: This will search for block `1account.near`, and those accounts are valid.

shelegdmitriy: What if we use `Number`? So the condition will be `if (!isNaN(Number(blockId)))`

luixo: LGTM

luixo (approved):

PR #1083, opened 2022-07-25
Support a new TelemetryInfo field: 'blob'

nikurt:
Support a new TelemetryInfo field: 'blob', which will be stored in the DB as a string. This field will let us easily extend TelemetryInfo without modifying explorer backend or the DB schema.

### DIFF
--- a/backend/src/database/models/readOnlyTelemetry/nodes.ts
+++ b/backend/src/database/models/readOnlyTelemetry/nodes.ts
@@ -40,6 +40,8 @@ export default interface Nodes {
   longitude: string | null;
 
   city: string | null;
+
+  blob: string | null;
 }
 
 export interface NodesInitializer {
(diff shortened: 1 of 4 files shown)


luixo:
Do we plan to have any checks on what's in the field?
Can malicious actor fill up our DB with 1048575 chars of junk?

frol:
@nikurt I would like not to store garbage. Please, consider to make telemetry messages well-defined

nikurt:
@frol is your concern that a random user can set the new field to be 1M of arbitrary text, or is your concern that every field of Telemetry info needs to be well defined in advance?
If the latter, that is justified by making it much easier to make changes to telemetry info.
If the former, do you have suggestions how to counteract that? Is 1M length limit to high?

frol:
@nikurt 
> If the latter, that is justified by making it much easier to make changes to telemetry info.

Ease of making changes can turn into a problem that clients (e.g. Explorer) cannot rely on having that information available.

As per our private discussion, I think for the experimentation phase it is the best to go with a custom telemetry server given that config.json allows to have a list of telemetry endpoints.

PR #1087, opened 2022-08-15
Feature: logging

luixo:
I'm not that sure what kind of metrics should we track, so I enabled default ones and some request / response logging.


anshal-savla:
RE: prometheus metrics review request https://nearinc.atlassian.net/browse/ISR-41

I am not very familiar with type script but I think you have got it right.

The normal steps are:

1. Define prometheus metrics that are relevant: 
- resources consumption (CPU, memory, IO, …)
- network traffic 
- business metrics (endpoint response time, requests errors, …)
2. Define a prometheus registry -> Collect defined metrics in the registry,
3. Expose the metrics.
4. Configure prometheus to scrape you targets.

One thing I see missing is custom labels. At the least, you would want to  to add job/service name (devconsole), environment (dev,prod..). 
Prometheus uses these labels to differentiate between different time-series if multiple services, environments are exposing metrics with the same name.

for example: if dev and prod environments in  dev-console are reporting `httpRequestCount` metrics then it can overwrite the metrics if you don't specify a label to differentiate between those metrics.

You can add default labels to each environment

```
const register = new promClient.Registry();
register.setDefaultLabels({
  job: 'devconsole', environment: 'prod'
});
```

luixo:
> One thing I see missing is custom labels. At the least, you would want to to add job/service name (devconsole), environment (dev,prod..).

I added them, though didn't commit at the moment.
You can check it out now as I commited it.

PR #1090, opened 2022-08-24
The index page now supports an optional 'q' query string param for searching

ryancwalsh:
I love setting up Custom Search Engines in Chrome and Brave browsers, but it seemed impossible to do it for NEAR Explorer until this PR.

chrome://settings/searchEngines

Now, as you can see in this gif, I can set up `ex` (or any key word that I choose) to be a key word that allows me to conveniently search for NEAR accounts or transactions or blocks straight from my Chrome address bar.

custom search engine

PR #1091, opened 2022-08-30
The index page now supports an optional 'q' query string param for searching

luixo:
As per original PR: https://github.com/near/near-explorer/pull/1090


ryancwalsh:
Do you know when this will be available on the public site? Thanks.

ryancwalsh:
Oh it already is!

The param is `query`; got it.  Thanks.

https://explorer.near.org/?query=6f44f491235c9a3cd452178aa0c677f201eb76245fb5e8e5a4a1039963a876d9

PR #1093, opened 2022-09-06
Store all fields from TelemetryInfo in the telemetry DB

nikurt:
`block_production_tracking_delay` and similar fields are already populated by nodes in testnet, but not yet in mainnet.

The PR is not tested, please advise how to test it.

### DIFF
--- a/backend/src/database/models/readOnlyTelemetry/nodes.ts
+++ b/backend/src/database/models/readOnlyTelemetry/nodes.ts
@@ -40,6 +40,24 @@ export default interface Nodes {
   longitude: string | null;
 
   city: string | null;
+
+  bandwidth_download: number | null;
+
+  bandwidth_upload: number | null;
+
+  cpu_usage: number | null;
+
+  memory_usage: number | null;
+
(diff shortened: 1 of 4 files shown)


luixo:
> The PR is not tested, please advise how to test it.

We don't have e2e testing on Explorer yet.

> but not yet in mainnet

Will you manually add the fields? We effectively don't use migrations (as it should be for DBs), but one-time tables addition for those who run their local DBs. Therefore, we will not be able to migrate real DBs by this code.

luixo:
> The PR is not tested, please advise how to test it.

For local manual testing I imagine running explorer locally and sending a request to `localhost:3000/api/nodes` with appropriate data.

luixo:
> We probably can just drop the telemetry tables before the deployment and let it recreate them on start

Do you mean we should just drop the table on every backend restart?

frol:
@luixo  nope, just do it once manually for these rare occasions. I know it is suboptimal, but this service should find its own home outside of Explorer one day anyway.

luixo (approved):

PR #1096, opened 2022-09-23
Replace mixpanel with segment

charleslavon:
Removes Mixpanel in favor of Segment events 

Tested manually by confirming events are added into Segment. 

e2e test: The trpc 429s and HTTP timeout errors have been occurring over at least past two merged PRs.


package.json
```diff
@@ -36,6 +36,7 @@
     ]
   },
   "dependencies": {
+    "@types/analytics-node": "^3.1.9",
```
frol, line 39: Should it be actually in the frontend/package.json?

charleslavon: yep

luixo: Types probably should go into `devDependecies` section.

frontend/src/hooks/analytics/analytics.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface Dict {
+  [key: string]: any;
+}
+
+let segment: Analytics;
+let anonymousId: string;
+
+function init() {
+  if (segment)
```
luixo, line 13: Do you mind adding `{ }` around to keep code unified through the projects (I didn't have time to add linting yet)

shelegdmitriy (approved):

frontend/src/hooks/analytics/analytics.ts
```diff
@@ -0,0 +1,44 @@
...
+    return console.log("Segment Analytics has already been initialized");
+  anonymousId = uniqueId(Date.now() + "_");
```
luixo, line 15: Why do we need `uniqueId` for that?

frontend/src/hooks/analytics/analytics.ts
```diff
@@ -0,0 +1,44 @@
...
+  //flushAt=1 is useful for testing new events
+  const {
+    publicRuntimeConfig: { segmentWriteKey },
+  } = getConfig();
+  const options = process.env.NODE_ENV === "development" ? { flushAt: 1 } : {};
+  segment = new Analytics(segmentWriteKey, options);
+}
+
+function track(eventLabel: string, properties?: Dict) {
```
luixo, line 24: Do you mind changing `function`s to constants (unification)?
```ts
const track = (eventLabel: string, properties: Dict) => {
  // ..
}
```

frontend/src/hooks/analytics/analytics.ts
```diff
@@ -0,0 +1,44 @@
...
+  segment.track({
+    anonymousId,
+    event: eventLabel,
+    properties,
+  });
+}
+
+function identify() {
```
luixo, line 32: Same here

frontend/src/hooks/analytics/analytics.ts
```diff
@@ -0,0 +1,44 @@
...
+  segment.identify({
+    userId: anonymousId,
+  });
+}
+
+const fns = {
```
luixo, line 38: Why not just export `identify`, `init` and `track`?

frontend/src/hooks/analytics/use-analytics-init.ts
```diff
@@ -1,17 +1,13 @@
 import * as React from "react";
-import mixpanel from "mixpanel-browser";
-
-const BROWSER_MIXPANEL_TOKEN = "df164f13212cbb0dfdae991da60e87f2";
+import analytics from "./analytics";
 
 if (typeof window !== "undefined") {
-  mixpanel.init(BROWSER_MIXPANEL_TOKEN);
-  mixpanel.register({ timestamp: new Date().toString() });
+  analytics.init();
```
luixo, line 5: I would suggest this piece actually be located in `analytics.ts`, but don't mind keeping it here as it would be status-quo

frontend.env
```diff
@@ -1,3 +1,4 @@
+NEAR_EXPLORER_CONFIG__SEGMENT_WRITE_KEY=7s4Na9mAfC7092R6pxrwpfBIAEek9Dne
```
luixo, line 1: Is this a development key?

luixo:
> e2e test: The trpc 429s and HTTP timeout errors have been occurring over at least past two merged PRs.

The 429 is on the backend checks (the functions run on timer to update network info) side, the RPC is responding with this code for some reason, it is not connected to TRPC.

I get the same error running explorer locally, but on production everything is fine. I believe IP addresses from which we run CI / I run server locally might get into filter list on RPC side or something like that.

frontend/src/hooks/analytics/analytics.ts
charleslavon: A value to use as an anonymous ID required by segment to track a user's actions

frontend/src/hooks/analytics/analytics.ts
charleslavon: I was following the original pattern of this file from the console. Perhaps they required name exports?

frontend.env
charleslavon: correct

frontend/src/hooks/analytics/analytics.ts
luixo: No, I mean why not just `Date.now()`?
Why do we need this extra wrapper?
Why not `Date.now() + '_' + Math.random()`?

frontend/src/hooks/analytics/analytics.ts
luixo: Yep, we're good with named exports in Explorer :)

frontend/src/hooks/analytics/analytics.ts
```diff
@@ -0,0 +1,43 @@
...
+
+let segment: Analytics;
+let anonymousId: string;
+
+const init = () => {
+  if (segment) {
+    return console.log("Segment Analytics has already been initialized");
+  }
+  anonymousId = uniqueId(Date.now() + "_");
+  //flushAt=1 is useful for testing new events
```
luixo, line 17: I believe this should go 3 lines down

luixo (approved):

frontend/src/hooks/analytics/analytics.ts
charleslavon: `Math.random()` is a solid option

PR #1105, opened 2022-10-17
DEC-566: Promote /beta/transactions Explorer page to non-beta

luixo:


luixo:
There is a problem on render, it doesn't deploy frontend. I have a discussion with tech support there.

shelegdmitriy (approved):

frol:
Does beta version of the page render well on mobile?

Also, there are quite some places where the transaction action misses quite some details, e.g. Stake action is rendering nothing useful:

image

https://explorer.near.org/beta/transactions/DpqDQri9u2monLxGu5sHztqV7vqDZvxsNuwZ1ZXtt3z2

Function call action could feature attached gas and attached deposit (well, it is missing from the current design as well)

And we have not solved the small NEAR amounts display, which might easily confuse people.

luixo:
> Does beta version of the page render well on mobile?

Nope, it does not.
We need this PR to preview what /beta pages look like (idk why not just use /beta prefix, but apparently PR is easier) to discuss how we can adapt it to production.

> Also, there are quite some places where the transaction action misses quite some details, e.g. Stake action is rendering nothing useful

Thanks for pointing out!

frol:
@luixo We never had time to sit down and actually think through each of the action details we display, and that is important to do.

luixo:
> We never had time to sit down and actually think through each of the action details we display, and that is important to do.

I'll point that out to those who make decisions.

charleslavon:
@luixo  please include in this PR a fix to the mobile layout. Notice how the header extends beyond the view's border.
IMG_4891



For the suggested improvements targeting enhancements which are also missing from the current explorer pages, let's save those for subsequent issues/PRs.

Issue #1118, opened 2022-11-21
Date format in Chinese is incorrect


think-in-universe:
### Description

The Chinese date format `第 1010日, 2022, 下午6:44:45` is wrong. 

The correct one should be `10月10日, 2022, 下午6:44:45`

This is probably caused by change made [here](https://github.com/near/near-explorer/pull/1039/files#diff-67a51ac485d4380ebb2428a29fc08e3cbe0659ec7dae9fce48081e6e2946146d)

### Example

https://explorer.testnet.near.org/accounts/v4-lfg.community-genesis2.bhc8521.testnet

<img width="326" alt="image" src="https://user-images.githubusercontent.com/46699230/203107976-0df7785d-c78a-453e-b424-3447fa4b4913.png">

frol:
@think-in-universe Would you be able to submit a PR to the code here: https://github.com/near/near-explorer/tree/master/frontend/src/libraries/locales? 🙏

think-in-universe:
@frol  yeah. I'm working on this today.

think-in-universe:
@frol Is there any way to start the frontend on testnet without launching the backend server? 

I have trouble to start the backend due to issues

```
✨ backend % npm run dev:mainnet

> near-explorer-backend@0.0.1 dev:mainnet
> dotenv -c -e ../mainnet.env -- npm run dev


> near-explorer-backend@0.0.1 dev
> nodemon src/index.ts

[nodemon] 2.0.20
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): *.*
[nodemon] watching extensions: ts,json
[nodemon] starting `ts-node src/index.ts`
/Users/robort/workspace/blockchain/projects/near/projects/toolbox/near-explorer/backend/node_modules/ts-node/src/index.ts:859
    return new TSError(diagnosticText, diagnosticCodes, diagnostics);
           ^
TSError: ⨯ Unable to compile TypeScript:
src/router/index.ts:17:20 - error TS2345: Argument of type 'Router<Context, Context, {}, Record<"protocolVersion", Procedure<Context, Context, {}, undefined, undefined, number, unknown, number>> & Record<...>, {}, {}, DefaultErrorShape>' is not assignable to parameter of type 'Router<unknown, any, {}, any, any, any, any>'.
  The types of '_def.middlewares' are incompatible between these types.
[...]

PR #1120, opened 2022-12-03
fix: Chinese date time format

think-in-universe:
Fixes: #1118

With this fix, the Chinese data format will be `2022年12月3日, 下午6:27:45`

image

### DIFF
--- a/frontend/src/translations/zh-hans/common.json
+++ b/frontend/src/translations/zh-hans/common.json
@@ -71,7 +71,7 @@
       "no": "否"
     },
     "date_time": {
-      "date_time_format": "Modo, yyyy, aaah:mm:ss",
+      "date_time_format": "yyyy年M月d日, ah:mm:ss",
       "date_format": "Modo, yyyy",
       "time_format": "aaa h:mm:ss",
       "transaction_date_format": "dd/MM/yyyy h:mm"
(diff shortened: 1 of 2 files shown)


frontend/src/translations/zh-hans/common.json
```diff
@@ -71,7 +71,7 @@
       "no": "否"
     },
     "date_time": {
-      "date_time_format": "Modo, yyyy, aaah:mm:ss",
+      "date_time_format": "yyyy年M月d日, ah:mm:ss",
```
luixo, line 74: Is it ok it was change to `ah` here, but kept as `aaah` in `zh-hant`?

frontend/src/translations/zh-hant/common.json
```diff
@@ -71,7 +71,7 @@
       "no": "否"
     },
     "date_time": {
-      "date_time_format": "Modo, yyyy, aaah:mm:ss",
+      "date_time_format": "yyyy年M月d日, aaah:mm:ss",
```
luixo, line 74: Is that ok that we have fixed values here instead of templates?
The documentation for formatting is here: https://date-fns.org/v2.29.3/docs/format

think-in-universe: Thanks for reminding. I tried formatting without fixed characters, but didn't find a way that works as expected last time.

I just tried again and find some examples here and it's working now: https://github.com/date-fns/date-fns/blob/main/src/locale/zh-CN/snapshot.md 😄

PR #1122, opened 2022-12-06
Migrate old account page to the new design

luixo:
We decided to reject activity list for now as we have to work through the design decisions there and first transfer Explorer to the design with existing data.


frol:
I would like to suggest rewriting this action message:

image
(Taken from: https://frontend-pr-1122-gv4j.onrender.com/beta/accounts/frol.near)

The new version that is accurate and hopefully easier to understand:

> <receiver_account_id> added an access key (ed25519:8d..Kaqr) that can be used only to call functions on social.near contract

luixo:
> The new version that is accurate and hopefully easier to understand

Added that, please check it out.
Why didn't you include a method list in your string?

PR #1125, opened 2022-12-12
Refactor: faster transactions list on account

luixo:
@telezhnaya would be great if you can add details why this query is better than the old one.

### DIFF
--- a/backend/src/router/transaction/list.ts
+++ b/backend/src/router/transaction/list.ts
@@ -4,7 +4,7 @@ import { z } from "zod";
 import { Context } from "../../context";
 import { validators } from "../validators";
 import { Indexer, indexerDatabase } from "../../database/databases";
-import { SelectQueryBuilder } from "kysely";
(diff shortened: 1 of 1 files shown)


shelegdmitriy (approved):

telezhnaya:
@luixo sure.
Short explanation: Postgres can heavily optimize joins, and that's what we see here. At the same time, it's much harder to optimize subqueries.

Some more details and reasoning:
We have the inner subquery without time ranges
```sql
select 
      "originated_from_transaction_hash" 
    from 
      "receipts" 
    where 
      "predecessor_account_id" = $1 
      or "receiver_account_id" = $2
```
It can produce huge temp table, and we will never use most of it.

The recommendation here: always prefer join over subquery if there's a chance to do so

As the homework, lol, you can try to add time ranges into the subquery, it should start working faster, but my guess it would still a little bit slower than the proposed option.

Issue #997, opened 2022-05-17
Add the info for developers


telezhnaya:
Previously block_height was just a number (e.g. `65775464`) on the main page, now it's comma-separated (`65,775,464`). It becomes impossible to just copy-paste it and use it while you debug anything, it's required to remove all commas at first.

Also, we have the `created` field which is parsed to a human-readable datetime. My dream is to have a raw timestamp somewhere.

Adding this stuff to the main interface sounds like overkill, but we have a special additional info icon. I think it would be great to add this raw info there.

frol:
I think we can copy the raw value to the clipboard when user clicks on those fields. @telezhnaya @khorolets Would that work for you?

telezhnaya:
Yes, it's even better!

luixo:
@telezhnaya 
How do you like that [with a separate button](https://github.com/near/near-explorer/pull/1017)?

telezhnaya:
Perfect, if the designers like it

telezhnaya:
Works for block height, thank you!
Could you please do the same for timestamps fields?

luixo:
Yep, can you please list which timestamp fields do you need me to put it?

telezhnaya:
`Created` field on block, account, and transaction pages
https://explorer.near.org/blocks/4VGLrzPQYogyEdKDZkkKSPK5K6M19xCgHJpA6AVBPwej
https://explorer.near.org/transactions/8ku14bnoFhsrwNBaCXifm323qXu6XwEBGTS3AREEb7L4
https://explorer.near.org/accounts/app.nearcrowd.near
Account also has `Last updated` sometimes

PR #1131, opened 2022-12-19
Feature: copy to clipboard timestamps

luixo:
Closes https://github.com/near/near-explorer/issues/997

@telezhnaya Please let me know if I forgot to add a button somewhere.


telezhnaya:
From the user perspective, it looks great to me. Thank you!

PR #1134, opened 2022-12-27
Fix: counting uptime for chunk-only producers

ruziev-dev:
Uptime counting has used only produced/expected blocks ratio. It's a bug for chunk-only producers uptime indicating.

| Current problem  | Fixed problem  |
| ------------------| -------------- |
|    |        |


luixo:
@ruziev-dev Thanks for the fix!

PR #1140, opened 2023-01-12
Refactor: search handler

luixo:
It doesn't make much sense to make a bunch of requests on every search request (they are batched, but anyway), we can have just one handler controlled from backend side.
Also, two of the requests got simplified to just checking existence instead of full scaled fetch of all the data.


backend/src/router/utils/search.ts
```diff
@@ -0,0 +1,114 @@
...
+      return;
+    } else if ("blockHash" in result) {
+      return "/blocks/" + result.blockHash;
+    } else if ("receiptId" in result) {
+      return `/transactions/${result.transactionHash}#${result.receiptId}`;
+    } else if ("transactionHash" in result) {
+      return "/transactions/" + result.transactionHash;
+    } else if ("accountId" in result) {
+      return "/accounts/" + result.accountId;
+    }
```
frol, line 112: nit: I think it is not a concern of the backend to decide on the redirection URL for the frontend. I would consider returning an object like `{"kind": "block", "id": result.blockHash}`.

luixo: That's true, I'll change that.
Does that create attack surface? I was thinking about that but didn't find one.

frol: @luixo I don't see an attack surface, but it creates unnecessary coupling between the backend and frontend. It is not critical for Explorer, yet it makes sense

PR #1145, opened 2023-01-25
Update render plan for testnet/backend to 'standard'

ecp88:
- The service is currently running out of memory.

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -82,6 +82,7 @@ services:
   - type: web
     name: backend/testnet
     env: node
+    plan: standard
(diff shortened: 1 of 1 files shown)


frol (approved):

luixo:
This was triggered by a crash of testnet server that happened first time in one-and-a-half month.
We currently have more severe issues on mainnet and reason behind both mainnet and testnet OOMs are yet to be discovered (probably after migrating from render to GCP).
If crashes will rise again - we should reopen this (or open a new) PR.

PR #1154, opened 2023-02-23
Fix: search text input caret behaviour

ruziev-dev:
### Current caret behaviour

https://user-images.githubusercontent.com/74172766/220989104-0f406e20-d7d8-43d7-8b95-ab9c517b63c8.mov

### Caret behaviour with this pull request

https://user-images.githubusercontent.com/74172766/220989148-80c09494-2812-4529-b761-73fa61e963aa.mov

### DIFF
--- a/frontend/src/components/utils/Search.tsx
+++ b/frontend/src/components/utils/Search.tsx
@@ -182,7 +182,8 @@ const Search: React.FC<Props> = React.memo(({ dashboard }) => {
   const { t } = useTranslation();
   const track = useAnalyticsTrack();
 
-  const [value, setValue] = useQueryParam("query");
+  const [inputValue, setInputValue] = React.useState<string>("");
+  const [queryValue, setQueryValue] = useQueryParam("query");
   const [searchValue, setSearchValue] = React.useState<string | undefined>(
     undefined
   );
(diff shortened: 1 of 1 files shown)


luixo (changes requested):

frontend/src/components/utils/Search.tsx
```diff
@@ -182,7 +182,10 @@ const Search: React.FC<Props> = React.memo(({ dashboard }) => {
   const { t } = useTranslation();
   const track = useAnalyticsTrack();
 
-  const [value, setValue] = useQueryParam("query");
+  const [queryValue, setQueryValue] = useQueryParam("query");
+  const [inputValue, setInputValue] = React.useState<string | undefined>(
```
luixo, line 186: Sorry I didn't make a thorough review in one go, but I think we can do 2 more things here:
1. Make `inputValue` always be a string - (`React.useState(queryValue || "")`), otherwise react will complain that element is changing state between being controllable and uncontrollable.
2. Use only one setter for `onChange` handler - `setInputValue`, and update query via `useEffect` as it is actually an effect - query string should update as an effect of a input value change.

I'll merge right after those are fulfilled.

@ruziev-dev Appreciate your effort a lot!

PR #1156, opened 2023-02-27
Feature: delegate actions

luixo:
Example: https://pr-1156---explorer-frontend-vcqilefdcq-uc.a.run.app/transactions/EAuEaQAZsgySoQ3gWEKzqLHdjd7Myzic1kZ9GpFUu5L5?network=testnet


frontend/src/translations/ru/common.json
```diff
@@ -347,6 +349,9 @@
         },
         "DeleteKey": {
           "key_deleted": "Ключ удалён: ${public_key}..."
+        },
+        "delegate": {
```
shelegdmitriy, line 353: Let's use PascalCase here as we used before: 
```suggestion
        "Delegate": {
```

backend/src/utils/actions.ts
```diff
@@ -1,68 +1,86 @@
...
+  | FunctionCallAction
+  | TransferAction
+  | StakeAction
+  | AddKeyAction
+  | DeleteKeyAction
+  | DeleteAccountAction;
+
+type DelegateAction = {
+  kind: "delegate";
+  args: {
```
khorolets, line 78: From what I can gather the list of arguments should be bigger, including:
- sender_id
- receiver_id
- actions +
- nonce
- max_block_height
- public_key

Sorry if you just skipped it during the draft and you was going to implement it later.

luixo: As we don't know what we are going to render on frontend - I didn't pass any of the props to the mapped action.

luixo: @khorolets Do you feel like we need to add anything more than we have at the moment?

PR #1188, opened 2023-04-05
Refactor: remove svgr

luixo:
Nextjs 13 [doesn't work](https://github.com/vercel/next.js/issues/47974) properly with custom webpack config in jest, so I removed it for now.


luixo:
Seems like we can workaround the problem without removing svgr

PR #1195, opened 2023-05-01
Refactor: migrate from cypress to playwright

luixo:
A migration to Playwright (hopefully it will get a testing DX in a better state it is now)


backend/src/server.ts
```diff
@@ -35,6 +36,14 @@ export const createApp = (options: RouterOptions) => {
     .use(cors())
     .use("/trpc", trpcExpress.createExpressMiddleware(options))
     .use("/ping", (_req, res) => res.send("OK"))
+    .use("/global-state", async (req, res) => {
```
luixo, line 39: This handle is needed to wait in playwright for global state to get populated so we always get success state of subscriptions on pages.

PR #1236, opened 2023-09-14
chore: change interval to 30 seconds rather than every second

roshaans:
According to the Database Query Insights, [this query](https://github.com/near/near-explorer/blob/97dd25c2d3b3fe3be492c558aa3811b727c06341/backend/src/cron/tasks.ts#L118-L141) was run in the backend over 20 thousand times in just the past hour alone (Almost 4 million requests in the past 7 days) and each request takes on average 1.8 seconds to make and the result is just always one row.

This PR hopes to lessen this bottleneck query by extending the interval for this query from every second to every 30 seconds.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -107,7 +107,7 @@ jobs:
       - uses: ./.github/actions/setup-node
       - uses: ./.github/actions/install-packages
       - name: Run unit tests
-        run: npm run -w frontend test
+        run: npm run -w frontend test -- --updateSnapshot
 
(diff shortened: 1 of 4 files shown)


shelegdmitriy (approved):

frol (approved):

roshaans:
Playwright tests seem to be failing because of a dependency on the database.

Screenshot 2023-09-18 at 5 45 21 PM

frol:
@roshaans Unit-tests fail for a completely different reason (snapshot mismatch), I don't understand how they used to pass, though 🤷‍♂️

PR #1246, opened 2024-02-19
feat: Enable StatelessNet network

khorolets:
*I am aware that NEAR Explorer has been sunset and considered as deprecated, but we decided to use it one last time for the Stake Wars IV since we don't expect the data on that network to overload all the underlying technologies behind the NEAR Explorer project*

I believe this change should do the trick to make this instance https://legacy.explorer.statelessnet.near.org/ of NEAR Explorer work for a new network launched for the Stake Wars IV.

P.S. I have no idea what I am doing with NEAR Explorer (following the method of the blind kitten)

### DIFF
--- a/common/src/types/common.ts
+++ b/common/src/types/common.ts
@@ -3,6 +3,7 @@ export type NetworkName =
   | "testnet"
   | "shardnet"
   | "guildnet"
+  | "statelessnet"
   | "localnet";
 
 // Workaround of Omit breaking discriminated unions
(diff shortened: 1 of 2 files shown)


shelegdmitriy (approved):

shelegdmitriy:
Hey @khorolets! The PR looks good but may I ask you to update this [type](https://github.com/near/near-explorer/blob/master/common/src/types/common.ts#L1) with your proposed network to fix **CI / Frontend typecheck**? Also you'll need to update [this](https://github.com/near/near-explorer/blob/master/frontend/src/components/utils/HeaderNetworkDropdown.tsx#L30) lines and I believe this will fix all the issues

khorolets:
Sorry for the false alarm. This change is redundant, and the instance wasn't working because of the `networks` configuration (I've set `offline: true` without figuring out what that meant).

I am closing this PR. It doesn't add any value to the project.

